PR #2, opened 2023-05-30
chore: removing static file server

ricardomaraschini:
as we are now installing helm charts after the cluster is online we don't need to have a static file server anymore. this commit removes the code related to the static file server.


ricardomaraschini:
Closing this one, it became too old now and we need to rewrite it.

PR #3, opened 2023-05-30
chore: stop asking for root permissions during helm install

ricardomaraschini:
with USE_SUDO=false the Makefile stops asking for the root password during build.


ricardomaraschini:
You have already implemented this, closing it.

PR #11, opened 2023-06-22
doc: improving readme file

ricardomaraschini:
Includes a general overview of the product as well as basic procedures for running it (stopping, starting), visualizing Kubernetes objects, and compiling it from source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,85 +1,88 @@
-# helmbin
+# Helmbin
 
-Embeds Kubernetes and Helm charts as a single installable binary.
+Helmbin is a command-line tool that enables users to render and run Helm Charts within a Virtual Machine.
 
-```bash
-$ ./bin/helmbin 
-An embeddable Kubernetes distribution
+## Introduction
 
-Usage:
-  helmbin [command]
+Helmbin embeds Kubernetes and Helm charts in a single installable binary. It is designed to simplify the process of rendering and deploying Helm Charts by providing a lightweight and isolated environment that can then be installed in a Virtual Machine. With Helmbin, you can quickly deploy Helm-based applications without the need for complex setup or dependencies.
 
-Available Commands:
-  completion  Generate the autocompletion script for the specified shell
-  help        Help about any command
-  install     Installs and starts a controller+worker as a systemd service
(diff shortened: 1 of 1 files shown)


camtr0n:
changed the label to `type::docs` (with an `s`) -- the `type::doc` label on this is causing some data tests to fail

README.md
```diff
@@ -1,85 +1,88 @@
...
+
+```bash
+$ sudo helmbin stop
+$ sudo helmbin start
+```
+
+You can enable Helmbin startup during boot by running:
+
+```bash
+$ sudo systemctl enable k0scontroller.service
```
marccampbell, line 59: This should be "branded" with the name of the helm chart that's packaged.
So, if we are defaulting to `kotsadm`, should should be `sudo systemctl enable kotsadm.service`

emosbaugh: 1. can we build this into the install command?
2. in the future i think we should write our own install command so we can brand things rather than passthrough to the k0s command.

README.md
```diff
@@ -1,85 +1,99 @@
-# helmbin
+# Helmbin
```
emosbaugh, line 1: should we work with product to name this project?

README.md
```diff
@@ -1,85 +1,99 @@
...
+
+## Introduction
+
+Helmbin embeds Kubernetes and Helm charts in a single installable binary. It is designed to simplify the process of rendering and deploying Helm Charts by providing a lightweight and isolated environment that can then be installed in a Virtual Machine. With Helmbin, you can quickly deploy Helm-based applications without the need for complex setup or dependencies.
+
+## Installation
+
+To install Helmbin, follow these steps:
+
+1. Download the latest release from the [Helmbin releases page](https://github.com/replicatedhq/helmbin/releases).
```
emosbaugh, line 13: for the future. i would like to add an install.sh similar to [this](https://github.com/k3s-io/k3s/blob/master/install.sh)

emosbaugh (approved):
only product feedback from me related to the state of the world and not the readme. pr lgtm

emosbaugh:
[sc-80437]

ricardomaraschini:
This is deprecated.

PR #12, opened 2023-06-22
chore: cover supervisor package with unit tests

ricardomaraschini:
This PR covers 84.1% of the `supervisor` package with unit tests. It also hugely simplifies its logic, everything that is not being used got removed.


pkg/supervisor/supervisor.go
```diff
@@ -21,277 +17,258 @@ import (
...
-	if err := s.maybeKillPidFile(nil, nil); err != nil {
-		return err
+// supervise starts the process and waits for it to exit.
+func (s *Supervisor) supervise(ctx context.Context) error {
+	defer func() {
+		close(s.done)
+	}()
+	s.log.Infof("Starting to supervise %q", s.name)
+	s.cmd = exec.Command(s.binPath, s.args...)
+	s.cmd.Stdout = logrus.WithField("stream", "stdout").Writer()
```
emosbaugh, line 146: i do see this comment in logWriter

```

// logWriter implements [io.Writer] by forwarding whole lines to log. In case
// the lines get too long, it logs them in multiple chunks.
//
// This is in contrast to logrus's implementation of io.Writer, which simply
// errors out if the log line gets longer than 64k.
```

emosbaugh (approved):

pkg/supervisor/supervisor.go
ricardomaraschini: I have tested this and it works just fine with lines longer than 64k, that is why I removed it. Maybe that was a problem but got fixed in newer logrus versions? I don't know. Anyways, it logs just fine.

ricardomaraschini: For sake of completeness: I tested with log lines > 1MB.

PR #15, opened 2023-06-27
chore: increase lint timeout

ricardomaraschini:
Increases the `lint` timeout to 10 minutes.


ricardomaraschini:
The verbose output showed that we take 5m29s to run the lint.

emosbaugh (approved):

PR #19, opened 2023-07-21
bug: filter helm releases by name

ricardomaraschini:
We were not filtering the returning helm releases by their name so we were able to install only one helm chart.


ricardomaraschini:
This is not needed anymore.

PR #23, opened 2023-08-09
feat: enable token based node join process

ricardomaraschini:
Adds a new way of deploying clusters. now we also support deploying one controller node and then from there generate a token to manually add other nodes. This makes HelmVM to behave closer to how kURL works.

What follows next are the text that this PR adds to the README.md file:

## Deploying Individual Nodes

HelmVM also facilitates deploying individual nodes through the use of tokens, deviating from the centralized approach.
To follow this path, you need to exclude yourself from the centralized management facilitated via SSH.

### Installing a Multi-Node Setup using Token-Based Deployment

All operations should be executed directly on the Linux servers and require root privileges.
Begin by deploying the first node:

```
server-0# helmvm install
```

After the cluster is online, you can generate a token to enable the addition of other nodes:

```
server-0# helmvm node token create --role controller
INFO[0000] Creating node join token for role controller
WARN[0000] You are opting out of the centralized cluster management.
WARN[0000] Through the centralized management you can manage all your
WARN[0000] cluster nodes from a single location. If you decide to move
WARN[0000] on the centralized management won't be available anymore
? Do you want to use continue ? Yes
INFO[0002] Token created successfully.
INFO[0002] This token is valid for 24h0m0s hours.
INFO[0002] You can now run the following command in a remote node to add it
INFO[0002] to the cluster as a "controller" node:
helmvm node join --role "controller" "<token redacted>"
server-0#
```

Upon generating the token, you will be prompted to continue; press Enter to proceed (you will be opting out of the centralized management).
The role in the command above can be either "controller" or "worker", with the generated token tailored to the selected role.
Copy the command provided and run it on the server you wish to join to the cluster:

```
server-1# helmvm node join --role "controller" "<token redacted>"
```

For this to function, you must ensure that the HelmVM binary is present on all nodes within the cluster.

### Upgrading clusters

If your installation employs centralized management, simply download the newer version of HelmVM and execute:

```
$ helmvm apply
```

For installations without centralized management, download HelmVM, upload it to each server in your cluster, and execute the following command as **root** on each server:

```
# helmvm node upgrade
```


cmd/helmvm/token.go
```diff
@@ -0,0 +1,98 @@
...
+		cmd.Stderr = io.Discard
+		if err := cmd.Run(); err != nil {
+			return fmt.Errorf("failed to create token: %w", err)
+		}
+		if !defaults.DecentralizedInstall() {
+			if err := defaults.SetInstallAsDecentralized(); err != nil {
+				return fmt.Errorf("failed to set decentralized install: %w", err)
+			}
+		}
+		logrus.Infof("Token created successfully.")
```
emosbaugh, line 91: it doesnt seem appropriate to use logging for output in this file. how about just print to stdout?

cmd/helmvm/install.go
```diff
@@ -310,6 +313,12 @@ var installCommand = &cli.Command{
 		},
 	},
 	Action: func(c *cli.Context) error {
+		if defaults.DecentralizedInstall() {
+			logrus.Warnf("Decentralized install was detected. To manage the cluster")
+			logrus.Warnf("you have to use the '%s node' commands instead.", defaults.BinaryName())
+			logrus.Warnf("Run '%s node --help' for more information.", defaults.BinaryName())
+			return nil
```
emosbaugh, line 320: seems like you should return an error here

cmd/helmvm/token.go
```diff
@@ -0,0 +1,98 @@
...
+		logrus.Infof("This token is valid for %s hours.", dur)
+		logrus.Infof("You can now run the following command in a remote node to add it")
+		logrus.Infof("to the cluster as a %q node:", role)
+		fmt.Printf("%s node join --role %q %s", defaults.BinaryName(), role, buf.String())
```
emosbaugh, line 95: would be cool to add a `--quiet, -q` flag to just print the token

emosbaugh (approved):

cmd/helmvm/token.go
kyleprice: Can we use a log level instead?

chris-sanders:
I'm impatient and a poor reader, you're going to have to put more guardrails to keep me from screwing it up :)

You're allowed to run `helmvm install` without sudo, but you can't run `helmvm node token create --role controller` without sudo. This results in there being no configuration when you do run sudo with the node command.

After resetting, and doing the install with sudo it was clean and worked as documented. The longest part of the whole process was waiting to upload the file 3x to the machines.

I'm curious why is sudo necessary to generate a token but not for an install? The downside to this is I also had to sudo to run `helmvm shell` which means I lost all my personal shell settings which is dissapointing.

Can we do the install and token w/o sudo?
If it's not something that can be done I suppose a local install should insist on sudo? I can see where a remote install (like doing `--multi-node` from a Mac wouldn't need it).

cmd/helmvm/token.go
ricardomaraschini: I have been using this across all the files, unfortunately. I can go ahead and remove it from all of them. Let me know what your thoughts are on how should we signalise things like `Warnf()` and `Errorf()` before I act on this.

cmd/helmvm/install.go
ricardomaraschini: Done.

cmd/helmvm/token.go
ricardomaraschini: I think that what @emosbaugh is talking about here is to print only the token and not the full join command. Is that it @emosbaugh ?

ricardomaraschini:
> You're allowed to run `helmvm install` without sudo, but you can't run `helmvm node token create --role controller` without sudo. This results in there being no configuration when you do run sudo with the node command.

I am curious, have you seen [this](https://github.com/replicatedhq/helmbin/pull/23/files#diff-97f9c68ca901e782d545b2bcbbf9c99f4b6acb0bddf00d9953322d87937b4ba9R54-R58) message asking you to run it with `sudo -E` ?

> I'm curious why is sudo necessary to generate a token but not for an install? The downside to this is I also had to sudo to run `helmvm shell` which means I lost all my personal shell settings which is dissapointing.

We need to run `k0s` binary to generate a token and we can only execute `k0s` binary if we are `root` or part of the `root` group:

```
ubuntu@ip-172-16-10-68:~$ ls -la /usr/local/bin/k0s
-rwxr-x--- 1 root root 230559893 Aug 10 06:20 /usr/local/bin/k0s
ubuntu@ip-172-16-10-68:~$
```

> Can we do the install and token w/o sudo? If it's not something that can be done I suppose a local install should insist on sudo? I can see where a remote install (like doing `--multi-node` from a Mac wouldn't need it).

Yeah, that makes sense.

chris-sanders:
> I am curious, have you seen [this](https://github.com/replicatedhq/helmbin/pull/23/files#diff-97f9c68ca901e782d545b2bcbbf9c99f4b6acb0bddf00d9953322d87937b4ba9R54-R58) message asking you to run it with sudo -E ?

I didn't see that but I'll have to try again to see if I was just not reading closely. It might have told me that and I missed it.

cmd/helmvm/token.go
emosbaugh: don't act on it yet but i think we should give it some more thought and maybe explore how other clis message the user. in my opinion, output that is informational for the user without verbosity increased should probably not be in log format with level and timestamp. this feedback is not for the prototype phase.

cmd/helmvm/token.go
ricardomaraschini: With all the logging happening here I am not sure this would be useful. At least not until we figure out what to do with the logging. :-(

ricardomaraschini:
> The only thing might be to lower some of the default logging as there was enough printed to screen that I wasn't reading the actually important stuff. That's a pretty simple tweak and more input than just my initial use of it should be seen to decide on that.

I have improved the messaging.

PR #31, opened 2023-08-16
chore: adding end to end tests

ricardomaraschini:
Start adding end to end tests to be run as part of GitHub actions.  The following E2E tests are added by this PR:

- E2E - TestBuildBundle
  - Tests if we can build the air-gap bundle. 
- E2E - TestEmbedAndInstall
  - Tests if we can embed a Helm Chart into HelmVM and install it during cluster deployment.
- E2E - TestInstallSingleNodeAndUpgradeToEmbed 
  - Tests we can install HelmVM and then "upgrade" it to a version that includes an Embedded Helm Chart.
- E2E - TestMultiNodeInstallation
  - Tests we can create multi-node clusters through SSH.
- E2E - TestSingleNodeInstallation
  - Tests we can deploy single node clusters.
- E2E - TestTokenBasedMultiNodeInstallation
  - Tests we can deploy multi-node clusters using token based method.

This is the `e2e/README.md` for the test framework introduced by this PR:

## E2E tests

Integration tests depends on LXD. The following procedure shows how to get everything installed on Ubuntu 22.04.

### Installing LXD

You can install LXD using snap:

```
$ snap install lxd
$ lxd init --auto
```
Once LXD is configured you need to set up ovn so we can create isolated networks:

```
$ sudo apt install ovn-host ovn-central -y
$ sudo ovs-vsctl set open_vswitch . \
   external_ids:ovn-remote=unix:/var/run/ovn/ovnsb_db.sock \
   external_ids:ovn-encap-type=geneve \
   external_ids:ovn-encap-ip=127.0.0.1
```

If you have problems setting up ovn you can check their docs: https://documentation.ubuntu.com/lxd/en/latest/howto/network_ovn_setup/

### Test workflow

Scripts inside the `scripts` directory are copied to all nodes. If you have a new test you want to add then start by creating a
shell script to execute it and save it under the `scripts` dir. You can then call the script from your Go code.

### Running all the tests

You can run the tests from within this directory:

```
$ make e2e-tests
```

### Running individual tests

You an run a single test with:

```
$ make e2e-test TEST_NAME=TestSomething
```

TestSomething is the name of the test function you want to run.

### Adding more tests

To add more tests you just need to create one inside this directory and then run:

```
$ make create-e2e-workflows
```


.github/workflows/e2e_testbuildbundle.yaml
```diff
@@ -0,0 +1,34 @@
+# this file has been generated by e2e/hack/create-e2e-workflows.sh.
+# edit there if you want to change something.
+name: E2E - TestBuildBundle
+on:
+- pull_request
+jobs:
+  e2e:
+    name: E2E - TestBuildBundle
+    runs-on: ubuntu-latest
```
emosbaugh, line 9: is the only difference in these tests the TEST_NAME? if so i wonder if we can use a matrix?

https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

```suggestion
  e2e:
    name: E2E
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      matrix:
        test:
          - TestBuildBundle
          - TestEmbedAndInstall
          - TestInstallSingleNodeAndUpgradeToEmbed
          - TestMultiNodeInstallation
          - TestSingleNodeInstallation
          - TestTokenBasedMultiNodeInstallation
```

.github/workflows/e2e_testbuildbundle.yaml
```diff
@@ -0,0 +1,34 @@
...
+      run: |
+        sudo apt install ovn-host ovn-central -y
+        sudo ovs-vsctl set open_vswitch .           external_ids:ovn-remote=unix:/var/run/ovn/ovnsb_db.sock           external_ids:ovn-encap-type=geneve           external_ids:ovn-encap-ip=127.0.0.1
+    - name: Setup Go
+      uses: actions/setup-go@v4
+      with:
+        go-version: "1.21.0"
+    - name: E2E tests (TestBuildBundle)
+      run: |
+        make e2e-test TEST_NAME=TestBuildBundle
```
emosbaugh, line 34:
```suggestion
    - name: E2E tests (${{ matrix.test }})
      run: |
        make e2e-test TEST_NAME=${{ matrix.test }}
```

.github/workflows/e2e_testbuildbundle.yaml
```diff
@@ -0,0 +1,34 @@
+# this file has been generated by e2e/hack/create-e2e-workflows.sh.
+# edit there if you want to change something.
+name: E2E - TestBuildBundle
```
emosbaugh, line 3:
```suggestion
name: E2E
```

.github/workflows/e2e_testbuildbundle.yaml
```diff
@@ -0,0 +1,34 @@
...
+on:
+- pull_request
+jobs:
+  e2e:
+    name: E2E - TestBuildBundle
```
emosbaugh, line 8:
```suggestion
    name: E2E
```

Makefile
```diff
@@ -125,15 +125,33 @@ builder: static static-linux-amd64
 
 .PHONY: unit-tests
 unit-tests:
-	go test -v ./...
+	go test -v $(shell go list ./... | grep -v /e2e)
```
emosbaugh, line 128: would be cool to use build tags https://www.digitalocean.com/community/tutorials/customizing-go-binaries-with-build-tags

e2e/scripts/embed-and-install.sh
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env bash
+# This script is used to test the embed capability of helmvm. It will pull the memcached helm chart and
+# embed it into the binary, issuing then a single node install. This script waits for the memcached pod
+# to be ready.
+
```
emosbaugh, line 5: i find it typically safer to set the following when writing scripts although pipefail can be difficult if you dont build with it in mind. take it or leave the suggestion

```suggestion

set -euo pipefail

```

e2e/scripts/multi-node-install.sh
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env bash
+
```
emosbaugh, line 2: ditto re `set -euo pipefail`

e2e/scripts/single-node-install.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
```
emosbaugh, line 2: ditto re `set -euo pipefail`

e2e/scripts/wait-for-ready-nodes.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+# This script waits for X nodes to be ready. X is the first argument.
+
```
emosbaugh, line 3: ditto re `set -euo pipefail`

.github/workflows/e2e_testbuildbundle.yaml
ricardomaraschini: Yes, this is much better. Done.

Makefile
ricardomaraschini: Yeah, that would be nice but it just breaks my gopls. I will let this out for now, until I figure out how to get gopls properly working here.

e2e/scripts/wait-for-ready-nodes.sh
ricardomaraschini: As this has been mentioned multiple times I am closing this to keep the discussion [here](https://github.com/replicatedhq/helmbin/pull/31#discussion_r1298473417).

e2e/scripts/embed-and-install.sh
ricardomaraschini: I was using it but then it was making things sorta flake because [this](https://github.com/replicatedhq/helmbin/pull/31/files#diff-e1c5d3a0f14947426c8657d3a54466390a0c27fef9f8d1e0474d63d175fae518R7) was failing from time to time. I mean, it is OK if it fails, it should keep trying.

In other words: how can I make the following command not to make the whole script fail with `set -euo pipefail` ?

```
ready=$(kubectl get nodes | grep -v NotReady | grep -c Ready)
```

ricardomaraschini: For extra context: when deploying with the token based authentication sometimes the `kubectl get nodes` was returning an error like "leader changed" or something. With the `-euo pipefail` the whole test was failing instead of just set "ready" to 0.

emosbaugh: perhaps something like

```
ready=$(kubectl get nodes | grep -v NotReady | { grep  -c Ready || true; }
```

emosbaugh: perhaps at the very least you can add `set -eu`

ricardomaraschini: Seems like the following did the trick:

```
ready=$(kubectl get nodes | grep -v NotReady | grep -c Ready || true)
```

ricardomaraschini: Resolving.

e2e/scripts/embed-and-install.sh
```diff
@@ -62,8 +63,8 @@ wait_for_memcached_pods() {
         sleep 5
         counter=$((counter+1))
         echo "Waiting for memcached pods to be ready"
-        ready=$(kubectl get pods -n helmvm | grep  Running | grep -c memcached)
-        kubectl get pods -n helmvm 2>&1
+        ready=$(kubectl get pods -n helmvm | grep  Running | grep -c memcached || true)
+        kubectl get pods -n helmvm 2>&1 || true
```
emosbaugh, line 67: why does this fail?

emosbaugh (approved):

e2e/scripts/embed-and-install.sh
ricardomaraschini: That should not. When it started to flake in other part I added `|| true` in every place where to work or not is optional.

PR #37, opened 2023-08-21
feat: introducing plain prompts

ricardomaraschini:
plain prompts do not leverage survey package. survey package does not behave well with stdin redirection so it is quite hard to test an interactive installation. with plain prompts this should be straight forward to integrate with tools like `expect`.

this commit also adds an end-to-end test that uses plain prompts.


pkg/prompts/prompts.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"os"
+
+	"github.com/replicatedhq/helmvm/pkg/prompts/decorative"
+	"github.com/replicatedhq/helmvm/pkg/prompts/plain"
+)
+
+// Confirm asks a user for a "Yes" or "No" response. The default value
+// is used if the user just presses enter without typing anything.
+func Confirm(msg string, defvalue bool) bool {
```
emosbaugh, line 16: why not make this an interface?

cmd/helmvm/token.go
```diff
@@ -69,17 +69,8 @@ var tokenCreateCommand = &cli.Command{
...
-					Message: "Do you want to use continue ?",
-					Default: true,
-				}
-				var moveOn bool
-				if err := survey.AskOne(question, &moveOn); err != nil {
-					return fmt.Errorf("unable to ask for confirmation: %w", err)
-				} else if !moveOn {
-					return nil
-				}
+			if useprompt && !prompts.Confirm("Do you want to use continue ?", true) {
```
emosbaugh, line 72:
```suggestion
			if useprompt && !prompts.Confirm("Do you want to continue ?", true) {
```

pkg/config/config.go
```diff
@@ -79,67 +82,35 @@ func listUserSSHKeys() ([]string, error) {
...
-	}, host); err != nil {
-		return fmt.Errorf("unable to ask for host config: %w", err)
+	host.Address = prompts.Input("Node address:", host.Address, true)
+	host.User = prompts.Input("SSH user:", host.User, true)
+	var err error
+	var port int
+	for port == 0 {
+		str := prompts.Input("SSH port:", strconv.Itoa(host.Port), true)
+		if port, err = strconv.Atoi(str); err != nil {
+			logrus.Warnf("Invalid port number")
```
emosbaugh, line 109: just curious why this is a warning and doesnt return an error?

emosbaugh (changes requested):

pkg/config/config.go
ricardomaraschini: We allow users to type the value again if they by mistake entered a string there.

pkg/prompts/prompts.go
ricardomaraschini: Makes total sense, done.

emosbaugh (approved):

PR #38, opened 2023-08-22
bug: version command should work without cluster access

ricardomaraschini:
`helmvm version` should work even without having access to a cluster. prior to this commit version command required a kubernetes client to work properly, this is not a requirement.

### DIFF
--- a/e2e/version_test.go
+++ b/e2e/version_test.go
@@ -1,0 +1,41 @@
+package e2e
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/replicatedhq/helmvm/e2e/cluster"
+)
+
+func TestVersion(t *testing.T) {
+	t.Parallel()
+	tc := cluster.NewTestCluster(&cluster.Input{
+		T:             t,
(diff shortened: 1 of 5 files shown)


e2e/version_test.go
```diff
@@ -0,0 +1,41 @@
...
+	line := []string{"helmvm", "version"}
+	stdout, stderr, err := RunCommandOnNode(t, tc, 0, line)
+	if err != nil {
+		t.Fatalf("fail to install ssh on node %s: %v", tc.Nodes[0], err)
+	}
+	var failed bool
+	output := fmt.Sprintf("%s\n%s", stdout, stderr)
+	expected := []string{"Installer", "Kubernetes", "OpenEBS", "AdminConsole"}
+	for _, component := range expected {
+		if strings.Contains(output, "Installer") {
```
laverya, line 32: should this not be 'component'?
```suggestion
		if strings.Contains(output, component) {
```

Right now you check for `Installer` 4x

ricardomaraschini: whoops^4

PR #55, opened 2023-09-05
chore: implementing some unit tests

ricardomaraschini:
This is an ongoing effort. This PR implements unit tests for both `defaults` and `ssh` packages. As I move forward I will be opening new PRs for different packages.

### DIFF
--- a/pkg/defaults/provider.go
+++ b/pkg/defaults/provider.go
@@ -1,0 +1,209 @@
+package defaults
+
+import (
+	"fmt"
+	"net"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/gosimple/slug"
+)
+
+const (
+	k0sBinsSubDirDarwin = "Library/Caches/k0sctl/k0s/linux/amd64"
+	k0sBinsSubDirLinux  = ".cache/k0sctl/k0s/linux/amd64"
+)
+
+// NewProvider returns a new DefaultsProvider using the provided base dir.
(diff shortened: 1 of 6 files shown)


pkg/defaults/provider.go
```diff
@@ -0,0 +1,198 @@
...
+// DefaultsProvider is an entity that provides default values used during
+// HelmVM installation.
+type DefaultsProvider struct {
+	Base string
+}
+
+// Init makes sure all the necessary directory exists on the system.
+func (d *DefaultsProvider) Init() {
+	if err := os.MkdirAll(d.K0sctlBinsSubDir(), 0755); err != nil {
+		panic(fmt.Errorf("unable to create basedir: %w", err))
```
danj-replicated, line 37: I'm not a massive fan of using `panic()` like this when an end user is going to be seeing the messages. it's nicer to use something like `logrus.Fatalf()`

danj-replicated (changes requested):
aside from a few issues with the UX of using `panic` for fatal errors, looks good 👍

pkg/defaults/provider.go
ricardomaraschini: Yeah, that makes sense. Let me refactor that to replace the calls by `logrus.Fatalf`.

ricardomaraschini: @danj-replicated solved, can you please take a look ?

danj-replicated (approved):

PR #56, opened 2023-09-05
feat: tidy up install output

danj-replicated:
This PR:
- creates a new `logging` package to control the default logrus writer
- sets up hooks to:
   - allow selective writing of logs to stdout
   - write all logs to a file regardless of level
- uses the above to tidy up messages presented to the user at install / upgrade time. 
- adds a -d / --debug flag to enable printing _all_ loglevels to stdout

the result of this is: 
```
k0s:~# ./helmvm install
Processing cluster configuration
✓  [ssh] 192.168.0.8:22: executing `[ "$(id -u)" = 0 ]` 
Applying cluster configuration
✓  Running k0sctl apply 
✓  Kubernetes API server is ready. 
? Enter a new Admin Console password: ************
✓  Post apply process finished 
Cluster configuration has been applied
Kubeconfig file has been placed at at /root/.helmvm/etc/kubeconfig
Cluster configuration file has been placed at /root/.helmvm/etc/k0sctl.yaml
You can now access your cluster with kubectl by running:
  ./helmvm shell
k0s:~# ls
helmvm	helmvm-20230905114958.log
```


cmd/helmvm/main.go
```diff
@@ -5,14 +5,30 @@ import (
...
+	logging.SetupLogging()
+
 	name := path.Base(os.Args[0])
 	var app = &cli.App{
-		Name:  name,
+		Name: name,
+		Flags: []cli.Flag{
+			&cli.BoolFlag{
+				Name:    "debug",
+				Usage:   "output all setup messages to stdout",
```
ricardomaraschini, line 23: > 
```suggestion
				Usage:   "Output all setup messages to stdout",
```

pkg/logging/logging.go
```diff
@@ -0,0 +1,88 @@
+package logging
```
ricardomaraschini, line 1: At least one file in each package should contain a documentation indicating what the package does. You can see examples in other packages inside the repository.

pkg/logging/logging.go
```diff
@@ -0,0 +1,88 @@
...
+	Formatter *logrus.TextFormatter
+}
+
+func (hook *StandardHook) Fire(entry *logrus.Entry) error {
+	text, err := hook.Formatter.Format(entry)
+	if err != nil {
+		return err
+	}
+
+	if !Debug && entry.Level == logrus.InfoLevel || entry.Level == logrus.DebugLevel {
```
ricardomaraschini, line 61: I don't fully understand this `if` here. Can you elaborate a little bit longer ? I mean, how these are evaluated ?

pkg/logging/logging.go
```diff
@@ -0,0 +1,88 @@
...
+
+	logrus.AddHook(&StandardHook{
+		Writer:    os.Stderr,
+		LogLevels: logrus.AllLevels,
+		Formatter: &logrus.TextFormatter{ForceColors: true},
+	})
+
+	now := time.Now().Format("20060102150405")
+
+	fileHook, err := NewLogrusFileHook("helmvm-"+now+".log", os.O_CREATE|os.O_APPEND|os.O_RDWR, 0666)
```
ricardomaraschini, line 84: I think that this should be placed under `~/.helmvm/logs/`. What do you think ? Keep everything inside the same place is something we have to aim. I was implementing something like this [here](https://github.com/replicatedhq/helmvm/pull/52/files#diff-873ba6ea91d97fbdbd851e2af37c2dc38cdd079dedef667a3a9dca4f457ae7e9). Later on we can make a Support Bundle to collect everything inside this directory.

ricardomaraschini (changes requested):
Excellent job Dan, this is lovely. I have left a few comments.

pkg/logging/logging.go
danj-replicated: as I understand it, it should work like ( if debug is true ) and ( line loglevel == info or debug )

ricardomaraschini (approved):

PR #86, opened 2023-09-26
feat: deploying admin console with the embedded license

ricardomaraschini:
if a license exists inside the binary we configure admin console to use it. if no license was found deploys admin console as we used to (without license).

### DIFF
--- a/pkg/addons/adminconsole/adminconsole.go
+++ b/pkg/addons/adminconsole/adminconsole.go
@@ -13,6 +13,7 @@ import (
 	"helm.sh/helm/v3/pkg/chart/loader"
 	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/release"
+	"sigs.k8s.io/yaml"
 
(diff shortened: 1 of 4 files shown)


pkg/addons/adminconsole/adminconsole.go
```diff
@@ -63,6 +64,28 @@ func (a *AdminConsole) HostPreflights() (*v1beta2.HostPreflightSpec, error) {
...
+	raw, err := yaml.Marshal(license)
+	if err != nil {
+		return fmt.Errorf("unable to marshal license: %w", err)
+	}
+	helmValues["automation"] = map[string]interface{}{
+		"license": map[string]interface{}{
+			"slug": license.Spec.AppSlug,
+			"data": string(raw),
+		},
+	}
```
laverya, line 85: I would _prefer_ if this was a safeadd (ie it doesn't wipe automation.somethingelse) but this works for now

laverya (approved):

PR #87, opened 2023-09-27
chore: sets a version on development build

ricardomaraschini:
use `git describe --tags` to set a proper version on development builds (builds from `main` branch).


chris-sanders (approved):
Looks good!

PR #88, opened 2023-09-29
chore: switch our own addons to use the native helm installer

danj-replicated:
This PR swiches the helm chart installation method to use the k0s native helm reconciler.

addon packages now generate snippets of k0s config that are assembled along with the rest of the cluster config.

It also changes the password prompt for kotsadm to ask for password confirmation


pkg/addons/custom/custom.go
```diff
@@ -57,43 +54,72 @@ func (c *Custom) HostPreflights() (*v1beta2.HostPreflightSpec, error) {
...
+// GenerateHelmConfig generates the helm config for all the embedded charts.
+// and writes the charts to the disk.
+func (c *Custom) GenerateHelmConfig(ctx *cli.Context, isUpgrade bool) ([]dig.Mapping, error) {
+
+	chartConfigs := []dig.Mapping{}
+
 	exe, err := os.Executable()
 	if err != nil {
-		return fmt.Errorf("unable to get executable path: %w", err)
+		return chartConfigs, fmt.Errorf("unable to get executable path: %w", err)
```
ricardomaraschini, line 65: this should return nil and the error.

pkg/addons/custom/custom.go
```diff
@@ -57,43 +54,72 @@ func (c *Custom) HostPreflights() (*v1beta2.HostPreflightSpec, error) {
...
 	}
 	opts, err := hembed.ReadEmbedOptionsFromBinary(exe)
 	if err != nil {
-		return fmt.Errorf("unable to read embed options: %w", err)
+		return chartConfigs, fmt.Errorf("unable to read embed options: %w", err)
```
ricardomaraschini, line 69: this should return nil and the error.

pkg/addons/custom/custom.go
```diff
@@ -57,43 +54,72 @@ func (c *Custom) HostPreflights() (*v1beta2.HostPreflightSpec, error) {
...
+			"name":      chartName,
+			"namespace": c.namespace,
+			"version":   chartData.Metadata.Version,
+		}
+
+		chartConfig["chartName"] = dstpath
+
+		chartConfig["values"] = chart.Values
+
+		dst, err := os.Create(dstpath)
```
ricardomaraschini, line 101: isn't this going to crash if the file already exists?

pkg/addons/custom/custom.go
```diff
@@ -57,43 +54,72 @@ func (c *Custom) HostPreflights() (*v1beta2.HostPreflightSpec, error) {
...
+
+		defer func() {
+			if err := dst.Close(); err != nil {
+				logrus.Errorf("error closing file %s: %s", dstpath, err)
+			}
+		}()
+
+		if err != nil {
```
ricardomaraschini, line 109: this is a bit off, if err is not nil then dst is and we are calling close on it in defer.

pkg/addons/custom/custom.go
```diff
@@ -57,43 +54,72 @@ func (c *Custom) HostPreflights() (*v1beta2.HostPreflightSpec, error) {
...
+			logrus.Fatalf("could not write helm chart archive: %s", err)
+		}
+
+		_, err = io.Copy(dst, chart.ChartReader())
+		if err != nil {
+			logrus.Fatalf("could not write helm chart archive: %s", err)
```
ricardomaraschini, line 115: Can we return the error here?

pkg/addons/openebs/openebs.go
```diff
@@ -3,23 +3,26 @@
...
+	"gopkg.in/yaml.v2"
 
 	"github.com/replicatedhq/helmvm/pkg/addons/openebs/charts"
-	pb "github.com/replicatedhq/helmvm/pkg/progressbar"
+	"github.com/replicatedhq/helmvm/pkg/defaults"
 )
 
 const (
 	releaseName = "openebs"
+	appVersion  = "3.7.0"
```
ricardomaraschini, line 25: This shouldn't be hardcoded here (the same is valid for the admin console version). We have this version set also in the Makefile and now we need to remember updating in two places.

pkg/addons/openebs/openebs.go
```diff
@@ -56,63 +57,65 @@ func (o *OpenEBS) HostPreflights() (*v1beta2.HostPreflightSpec, error) {
...
 	}
-	fname := fmt.Sprintf("openebs-%s.tgz", strings.TrimPrefix(version, "v"))
-	hfp, err := charts.FS.Open(fname)
+
+	chartConfig["chartName"] = filepath.Join(defaults.HelmChartSubDir(), o.GetChartFileName())
+
+	valuesStringData, err := yaml.Marshal(helmValues)
 	if err != nil {
-		return fmt.Errorf("unable to find version %s: %w", version, err)
+		return chartConfigs, err
```
ricardomaraschini, line 78: this should return nil and the error. i suspect you have done this by design so i won't comment on it anymore.

pkg/addons/openebs/openebs.go
```diff
@@ -56,63 +57,65 @@ func (o *OpenEBS) HostPreflights() (*v1beta2.HostPreflightSpec, error) {
...
-	if release == nil {
-		o.logger("OpenEBS hasn't been installed yet, installing it.")
-		act := action.NewInstall(o.config)
-		act.Namespace = o.namespace
-		act.ReleaseName = releaseName
-		act.CreateNamespace = true
-		if _, err := act.RunWithContext(ctx, hchart, helmValues); err != nil {
-			return fmt.Errorf("unable to install chart: %w", err)
+	dstpath := filepath.Join(defaults.HelmChartSubDir(), chartfile)
+	dst, err := os.Create(dstpath)
```
ricardomaraschini, line 101: isn't this going to crash if the file already exists? here the problem with the close on defer also happens (the if err != nil is a bit off thing).

pkg/config/config.go
```diff
@@ -216,12 +217,12 @@ func askForLoadBalancer() (string, error) {
 
 // renderMultiNodeConfig renders a configuration to allow k0sctl to install in a multi-node
 // configuration.
-func renderMultiNodeConfig(ctx context.Context, nodes []infra.Node) (*v1beta1.Cluster, error) {
+func renderMultiNodeConfig(ctx *cli.Context, nodes []infra.Node, helmConfigs dig.Mapping) (*v1beta1.Cluster, error) {
```
ricardomaraschini, line 220: can we use another variable name for the cli.Context ? In some other places here it is called only `c`.  BTW, why do we need to pass in the urfave type this deep into the code path ? Consider we might want to migrate to cobra later on.

pkg/defaults/provider.go
```diff
@@ -150,6 +159,11 @@ func (d *DefaultsProvider) PathToHelmVMBinary(name string) string {
 	return filepath.Join(d.HelmVMBinsSubDir(), name)
 }
 
+// PathToHelmChart returns the directory helm chart archives are meterialized in.
+func (d *DefaultsProvider) PathToHelmChart(name string, version string) string {
```
ricardomaraschini, line 163: OH, you already have the helper function but is not using it across the PR.

ricardomaraschini (changes requested):

cmd/helmvm/install.go
```diff
@@ -145,10 +145,11 @@ func createK0sctlConfigBackup(ctx context.Context) error {
 // updates the files that need to be uploaded to the nodes). This function also
 // makes sure that the k0s version used in the configuration matches the version
 // we are planning to install.
-func updateConfigBundle(ctx context.Context, bundledir string) error {
-	if err := createK0sctlConfigBackup(ctx); err != nil {
+func updateConfigBundle(c *cli.Context, bundledir string) error {
```
ricardomaraschini, line 148: Now that we are passing on the `cli.Context` we don't need to pass in `bundledir`, what do you think ? I believe that we also need to rephrase the comment for this function, it does not totally reflect what the function does. Maybe this function should be called `updateConfig` or something.

cmd/helmvm/install.go
```diff
@@ -412,50 +443,33 @@ var installCommand = &cli.Command{
 			metrics.ReportApplyFinished(c, err)
```
ricardomaraschini, line 443: We need to update the comment at line `388` to reflect the new way we install the add-ons.

cmd/helmvm/install.go
```diff
@@ -412,50 +443,33 @@ var installCommand = &cli.Command{
...
 			return err
 		}
-		if !c.Bool("addons-only") {
```
ricardomaraschini, line 415: Let's remove the `addons-only` flag altogether ? It seems like we don't need it anymore. What do you think ? Remember to check if it is not referenced in the README.

cmd/helmvm/upgrade.go
```diff
@@ -148,21 +148,9 @@ var upgradeCommand = &cli.Command{
...
-		}
-		for _, addon := range c.StringSlice("disable-addon") {
-			opts = append(opts, addons.WithoutAddon(addon))
-		}
-		if err := addons.NewApplier(opts...).Apply(c.Context); err != nil {
-			err := fmt.Errorf("unable to apply addons: %w", err)
-			metrics.ReportNodeUpgradeFailed(c.Context, err)
-			return err
-		}
+
```
ricardomaraschini, line 151: We were applying the new add-ons but now we are not. Shouldn't we be updating the k0s configuration with the new add-ons here ?

go.mod
```diff
@@ -30,6 +30,11 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require (
+	github.com/Masterminds/sprig v2.22.0+incompatible // indirect
+	github.com/denisbrodbeck/machineid v1.0.1 // indirect
+)
+
```
ricardomaraschini, line 37: Why are these in a separated `require` section ?

pkg/addons/adminconsole/adminconsole.go
```diff
@@ -36,26 +36,44 @@ var helmValues = map[string]interface{}{
...
-	return prompts.New().Password("Enter a new Admin Console password:"), nil
+
+	maxTries := 3
+	for i := 0; i < maxTries; i++ {
+		promptA := prompts.New().Password("Enter a new Admin Console password:")
+		promptB := prompts.New().Password("Confirm password:")
+
+		if promptA == promptB {
+			return promptA, nil
+		} else {
```
ricardomaraschini, line 60: this else is not necessary as we are returning in the previous line.

pkg/addons/adminconsole/adminconsole.go
```diff
@@ -36,26 +36,44 @@ var helmValues = map[string]interface{}{
...
+			fmt.Println("Passwords don't match, please try again.")
+		}
+	}
+
+	logrus.Fatalf("Unable to set Admin Console password after %d tries", maxTries)
```
ricardomaraschini, line 65: We should return the error instead of Fatalf.

pkg/addons/adminconsole/adminconsole.go
```diff
@@ -86,81 +104,156 @@ func (a *AdminConsole) addLicenseToHelmValues() error {
...
 }
 
-func (a *AdminConsole) Apply(ctx context.Context) error {
-	version, err := a.Latest()
-	if err != nil {
-		return fmt.Errorf("unable to get latest Admin Console version: %w", err)
+// GetPasswordFromConfig returns the adminconsole password from the cluster config.
+func (a *AdminConsole) GetPasswordFromConfig() (string, error) {
+
+	chartList := a.config.Spec.Extensions.Helm.Charts
```
ricardomaraschini, line 110: Can't Spec be nil an then this will segfault ? There may be other nullable properties inside Spec.

pkg/addons/adminconsole/adminconsole.go
```diff
@@ -86,81 +104,156 @@ func (a *AdminConsole) addLicenseToHelmValues() error {
...
+			pass = values["password"].(string)
+			break
+		}
 	}
-	if !semver.IsValid(version) {
-		return fmt.Errorf("unable to parse version %s", version)
+
+	if pass == "" {
+		return "", fmt.Errorf("unable to find adminconsole chart in cluster config")
+	} else {
```
ricardomaraschini, line 128: This else is not necessary. It just creates a warning on my linter:

```
if block ends with a return statement, so drop this else and outdent its block
```

I think it would be better and cleaner to just return empty if a password could not be find (judging by how you are using the values returned by this function).

pkg/addons/adminconsole/adminconsole.go
```diff
@@ -86,81 +104,156 @@ func (a *AdminConsole) addLicenseToHelmValues() error {
...
-	fname := fmt.Sprintf("adminconsole-%s.tgz", strings.TrimPrefix(version, "v"))
-	hfp, err := charts.FS.Open(fname)
-	if err != nil {
-		return fmt.Errorf("unable to find version %s: %w", version, err)
+}
+
+func (a *AdminConsole) CheckForExistingConfig() bool {
+
+	spec := a.config.Spec
+	if spec.Extensions == nil {
```
ricardomaraschini, line 137: This will segfault if `a.config.Spec` is nil.

pkg/addons/adminconsole/adminconsole.go
```diff
@@ -86,81 +104,156 @@ func (a *AdminConsole) addLicenseToHelmValues() error {
...
+		return false
 	}
-	defer hfp.Close()
 
-	if err := a.addLicenseToHelmValues(); err != nil {
-		return fmt.Errorf("unable to add license to helm values: %w", err)
+	extensions := spec.Extensions
+	if extensions.Helm == nil {
```
ricardomaraschini, line 142: This will segfault if spec.Extensions is nil.

pkg/addons/adminconsole/adminconsole.go
```diff
@@ -3,20 +3,20 @@
...
+	"github.com/k0sproject/k0s/pkg/apis/v1beta1"
 	"github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/mod/semver"
-	"helm.sh/helm/v3/pkg/action"
-	"helm.sh/helm/v3/pkg/chart/loader"
-	"helm.sh/helm/v3/pkg/cli"
-	"helm.sh/helm/v3/pkg/release"
-	"sigs.k8s.io/yaml"
+	"gopkg.in/yaml.v3"
```
ricardomaraschini, line 16: This import change most likely will break the embedded license marshalling. License is a kubernetes CRD and gopkg yaml file does not couple well with those types.

pkg/addons/adminconsole/adminconsole.go
```diff
@@ -86,81 +104,156 @@ func (a *AdminConsole) addLicenseToHelmValues() error {
...
-		a.logger("Admin Console hasn't been installed yet, installing it.")
+	chartConfig := v1beta1.Chart{
+		Name:      releaseName,
+		ChartName: "",
+		Version:   latest,
+		Values:    "",
+		TargetNS:  a.namespace,
+	}
+
+	chartConfig.ChartName = defaults.PathToHelmChart("adminconsole", latest)
```
ricardomaraschini, line 172: I will feel better if this was moved to line 166.

pkg/addons/adminconsole/adminconsole.go
```diff
@@ -86,81 +104,156 @@ func (a *AdminConsole) addLicenseToHelmValues() error {
...
+
+	if err := a.addLicenseToHelmValues(); err != nil {
+		return nil, fmt.Errorf("unable to add license to helm values: %w", err)
+	}
+
+	if a.CheckForExistingConfig() {
+		currentPassword, err := a.GetPasswordFromConfig()
+		if err != nil {
```
ricardomaraschini, line 180: I would argue that if an error is returned then we should bubble it up.

pkg/addons/adminconsole/adminconsole.go
```diff
@@ -86,81 +104,156 @@ func (a *AdminConsole) addLicenseToHelmValues() error {
...
-	loading.Infof("Applying AdminConsole addon")
-	defer loading.Close()
-	a.logger("Updating Admin Console from %s to %s", installedVersion, version)
-	act := action.NewUpgrade(a.config)
-	act.Namespace = a.namespace
-	if _, err := act.RunWithContext(ctx, releaseName, hchart, helmValues); err != nil {
-		return fmt.Errorf("unable to upgrade chart: %w", err)
+	err = a.WriteChartFile(latest)
+	if err != nil {
+		logrus.Fatalf("Unable to write chart file to disk: %s", err)
```
ricardomaraschini, line 211: No Fatalf. Bubble up the error instead.

pkg/addons/applier.go
```diff
@@ -18,8 +19,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
-	"github.com/replicatedhq/helmvm/pkg/addons/adminconsole"
 	"github.com/replicatedhq/helmvm/pkg/addons/custom"
+
+	//"github.com/replicatedhq/helmvm/pkg/addons/custom"
```
ricardomaraschini, line 24: Remove this commented line and glue similar imports together.

pkg/addons/applier.go
```diff
@@ -34,34 +37,40 @@ func getLogger(addon string, verbose bool) action.DebugLog {
...
-		return fmt.Errorf("unable to load addons: %w", err)
+		return nil, fmt.Errorf("unable to load addons: %w", err)
 	}
 	for _, addon := range addons {
-		if err := addon.Apply(ctx); err != nil {
-			return err
+
+		addonChartConfig, err := addon.GenerateHelmConfig()
+		if err != nil {
+			return nil, fmt.Errorf("Could not add chart: %w", err)
```
ricardomaraschini, line 67: This error message is not right. Should be `unable to ...`.

pkg/addons/custom/custom.go
```diff
@@ -57,43 +53,65 @@ func (c *Custom) HostPreflights() (*v1beta2.HostPreflightSpec, error) {
...
+		return nil, fmt.Errorf("unable to get executable path: %w", err)
 	}
 	opts, err := hembed.ReadEmbedOptionsFromBinary(exe)
 	if err != nil {
-		return fmt.Errorf("unable to read embed options: %w", err)
+		return nil, fmt.Errorf("unable to read embed options: %w", err)
 	} else if opts == nil {
-		c.logger("No embed charts found, skipping custom addons.")
-		return nil
+		logrus.Warn("No embed charts found, skipping custom addons.")
```
ricardomaraschini, line 70: Maybe we should omit this. It is OK not to have charts embedded.

pkg/addons/custom/custom.go
```diff
@@ -57,43 +53,65 @@ func (c *Custom) HostPreflights() (*v1beta2.HostPreflightSpec, error) {
...
-		if err := yaml.Unmarshal([]byte(ochart.Values), &values); err != nil {
-			return fmt.Errorf("unable to unmarshal values: %w", err)
+		if c.chartHasBeenDisabled(chartData) {
+			logrus.Infof("skipping disabled chart %s", chartData.Name())
+			continue
+		}
+
+		chartName := strings.ToLower(chartData.Name())
+		chartFile := fmt.Sprintf("%s-%s.tgz", chartName, chartData.Metadata.Version)
+		dstpath := filepath.Join(defaults.HelmChartSubDir(), chartFile)
```
ricardomaraschini, line 88: Shouldn't we use the function in the `defaults` instead of filepath.Join ?

pkg/addons/custom/custom.go
```diff
@@ -57,43 +53,65 @@ func (c *Custom) HostPreflights() (*v1beta2.HostPreflightSpec, error) {
...
+
+		chartConfig := v1beta1.Chart{
+			Name:     chartName,
+			Version:  chartData.Metadata.Version,
+			TargetNS: c.namespace,
+		}
+
+		chartConfig.ChartName = dstpath
+		chartConfig.Values = chart.Values
```
ricardomaraschini, line 97: These could be moved to line 91.

pkg/addons/openebs/openebs.go
```diff
@@ -56,63 +54,61 @@ func (o *OpenEBS) HostPreflights() (*v1beta2.HostPreflightSpec, error) {
...
+
+	chartConfig := v1beta1.Chart{
+		Name:     releaseName,
+		TargetNS: o.namespace,
+		Version:  latest,
 	}
-	fname := fmt.Sprintf("openebs-%s.tgz", strings.TrimPrefix(version, "v"))
-	hfp, err := charts.FS.Open(fname)
+
+	chartConfig.ChartName = defaults.PathToHelmChart(releaseName, latest)
```
ricardomaraschini, line 70: Can we move this to line 68 ?

pkg/addons/options.go
```diff
@@ -26,3 +30,10 @@ func Quiet() Option {
 		a.verbose = false
 	}
 }
+
+// Config sets the helm config for the addons.
+func Config(config v1beta1.ClusterConfig) Option {
```
ricardomaraschini, line 35: Would you mind naming this `WithConfig` instead ?

pkg/config/config.go
```diff
@@ -301,7 +307,45 @@ func renderSingleNodeConfig(ctx context.Context) (*v1beta1.Cluster, error) {
 		return nil, fmt.Errorf("unable to connect to %s: %w", ipaddr, err)
 	}
 	rhost := host.render()
-	return generateConfigForHosts("", rhost)
+	return generateConfigForHosts(ctx, "", rhost)
+}
+
+func UpdateHelmConfigs(cfg *v1beta1.Cluster, opts ...addons.Option) error {
```
ricardomaraschini, line 313: We need a comment here otherwise a lint warning is generated.

pkg/config/config.go
```diff
@@ -301,7 +307,45 @@ func renderSingleNodeConfig(ctx context.Context) (*v1beta1.Cluster, error) {
...
+
+	currentSpec := cfg.Spec.K0s.Config
```
ricardomaraschini, line 315: Can't spec be nil here ? What about the other properties ?

pkg/config/config.go
```diff
@@ -301,7 +307,45 @@ func renderSingleNodeConfig(ctx context.Context) (*v1beta1.Cluster, error) {
...
+
+	newClusterExtensions := &k0sconfig.ClusterExtensions{
+		Helm: newHelmExtension,
+	}
+
+	if spec, ok := cfg.Spec.K0s.Config["spec"].(map[string]interface{}); ok {
+		spec["extensions"] = newClusterExtensions
+		cfg.Spec.K0s.Config["spec"] = spec
+	} else {
+		spec := cfg.Spec.K0s.Config["spec"].(dig.Mapping)
```
ricardomaraschini, line 343: This should verify if the type matches as you did on line 339.

ricardomaraschini (changes requested):

go.mod
danj-replicated: honestly I have no idea how this happened?

ricardomaraschini (approved):

PR #91, opened 2023-10-03
chore: fix end to end tests

ricardomaraschini:
we were not running `apt update` before attempting to install packages and we need to change the way we pull the memcached helmchart (https://github.com/bitnami/charts/issues/19599).

### DIFF
--- a/e2e/embed_test.go
+++ b/e2e/embed_test.go
@@ -18,12 +18,15 @@ func TestEmbedAndInstall(t *testing.T) {
 	})
 	defer tc.Destroy()
 	t.Log("installing ssh in node 0")
-	line := []string{"apt", "install", "openssh-server", "-y"}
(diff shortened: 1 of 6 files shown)


e2e/embed_test.go
```diff
@@ -18,12 +18,15 @@ func TestEmbedAndInstall(t *testing.T) {
 	})
 	defer tc.Destroy()
 	t.Log("installing ssh in node 0")
-	line := []string{"apt", "install", "openssh-server", "-y"}
-	if _, _, err := RunCommandOnNode(t, tc, 0, line); err != nil {
+	commands := [][]string{
+		{"apt", "update", "-y"},
```
danj-replicated, line 22: minor nitpick but you might want to switch these to apt-get instead.

danj-replicated (approved):

PR #92, opened 2023-10-03
feat: reporting back installation and upgrade metrics

ricardomaraschini:
Reporting back Installation, Join, and Upgrades data. This work is described in details [here](https://docs.google.com/document/d/1LhOC5oHV5vGWjjFaN38Gk5hvi8nyIgVRSrcJzSa1ews/edit?usp=sharing).

## HelmVM adoption metrics

The reporting is based on events, some events will be generated by the installer (that this PR implements) and some by an operator running on the cluster (still to be implemented). The operator is responsible for keeping track of the nodes that compose the cluster while the installer is responsible for reporting installation, upgrades, and joins successes and failures.

These are the events reported through this PR:
- InstallationStarted
- InstallationSucceeded
- InstallationFailed
- UpgradeStarted
- UpgradeSucceeded
- UpgradeFailed
- NodeUpgradeStarted
- NodeUpgradeSucceeded
- NodeUpgradeFailed
- JoinStarted
- JoinSucceeded
- JoinFailed

This PR depends on https://github.com/replicatedhq/vandoor/pull/4333.

### DIFF
--- a/e2e/install_test.go
+++ b/e2e/install_test.go
@@ -283,7 +283,7 @@ func TestHostPreflight(t *testing.T) {
 	defer tc.Destroy()
 	t.Log("installing ssh and binutils on node 0")
 	commands := [][]string{
-		{"dnf", "install", "-y", "openssh-server", "binutils", "tar"},
+		{"dnf", "--setopt=metadata_expire=120", "install", "-y", "openssh-server", "binutils", "tar"},
 		{"systemctl", "enable", "sshd"},
 		{"systemctl", "start", "sshd"},
 	}
--- a/e2e/utils.go
+++ b/e2e/utils.go
@@ -18,17 +18,22 @@ func (b *buffer) Close() error {
 	return nil
 }
 
+// RunCommandsOnNode runs a series of commands on a node.
 func RunCommandsOnNode(t *testing.T, cl *cluster.Output, node int, cmds [][]string) error {
 	for _, cmd := range cmds {
 		cmdstr := strings.Join(cmd, " ")
 		t.Logf("running `%s` node %d", cmdstr, node)
-		if _, _, err := RunCommandOnNode(t, cl, node, cmd); err != nil {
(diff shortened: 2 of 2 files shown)


cmd/helmvm/token.go
```diff
@@ -83,14 +95,22 @@ var tokenCreateCommand = &cli.Command{
...
 			if err := defaults.SetInstallAsDecentralized(); err != nil {
-				return fmt.Errorf("failed to set decentralized install: %w", err)
+				return fmt.Errorf("unable to set decentralized install: %w", err)
 			}
 		}
+		token := JoinToken{ClusterID: metrics.ClusterID(), Token: buf.String(), Role: role}
+		encodedToken := bytes.NewBuffer(nil)
+		encoder := base64.NewEncoder(base64.StdEncoding, encodedToken)
+		marshaler := json.NewEncoder(encoder)
+		if err := marshaler.Encode(token); err != nil {
```
emosbaugh, line 105: How about an Encode() function on the JoinToken struct?

pkg/metrics/sender.go
```diff
@@ -0,0 +1,70 @@
...
+type Sender struct {
+	baseURL string
+}
+
+// Send sends an event to the metrics endpoint.
+func (s *Sender) Send(ctx context.Context, ev Event) {
+	url := fmt.Sprintf("%s/helmbin_metrics/%s", s.baseURL, ev.Title())
+	payload, err := s.payload(ev)
+	if err != nil {
+		logrus.Infof("unable to get payload for event %s: %s", ev.Title(), err)
```
emosbaugh, line 46: Why is this info and the rest level debug?

cmd/helmvm/token.go
ricardomaraschini: that makes sense. done.

pkg/metrics/sender.go
ricardomaraschini: my bad, it should have been a debugf as well. fixed.

emosbaugh (approved):

PR #105, opened 2023-10-13
Add supportbundle to e2d testing

chris-sanders:
Run support bundle and upload results for better e2e failure debugging.


chris-sanders:
I see how this is working now, this can't be done in the action like this.

PR #106, opened 2023-10-16
feat: create a cluster configuration config map

ricardomaraschini:
this config map will hold the cluster configuration. for now it is empty.

### DIFF
--- a/pkg/addons/adminconsole/customize.go
+++ b/pkg/addons/adminconsole/customize.go
@@ -128,6 +128,14 @@ func (a *AdminConsoleCustomization) apply(ctx context.Context) error {
 		logrus.Infof("Skipping admin console customization on %s", runtime.GOOS)
 		return nil
 	}
+	if err := a.kotsConfig(ctx); err != nil {
+		return err
+	}
+	return a.clusterConfig(ctx)
+}
+
+// kotsConfig reads the embedded kots config and creates it in the cluster.
+func (a *AdminConsoleCustomization) kotsConfig(ctx context.Context) error {
 	cust, err := a.extractCustomization()
 	if err != nil {
 		return fmt.Errorf("unable to extract customization from binary: %w", err)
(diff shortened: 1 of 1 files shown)


pkg/addons/adminconsole/customize.go
```diff
@@ -170,6 +178,28 @@ func (a *AdminConsoleCustomization) apply(ctx context.Context) error {
...
 }
 
+// clusterConfig makes sure a config map named "embedded-cluster-config" exists in the cluster.
+// XXX this will eventually be customizable by the user through the app release.
+func (a *AdminConsoleCustomization) clusterConfig(ctx context.Context) error {
+	kubeclient, err := a.kubeClient()
+	if err != nil {
+		return fmt.Errorf("unable to create kubernetes client: %w", err)
+	}
+	nsn := client.ObjectKey{Namespace: "kube-system", Name: "embedded-cluster-config"}
```
marccampbell, line 188: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L35

should we use that const?

pkg/addons/adminconsole/customize.go
```diff
@@ -170,6 +178,28 @@ func (a *AdminConsoleCustomization) apply(ctx context.Context) error {
...
+	var cm corev1.ConfigMap
+	if err := kubeclient.Get(ctx, nsn, &cm); err == nil {
+		return nil
+	}
+	cm = corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{Namespace: nsn.Namespace, Name: nsn.Name},
+		Data:       map[string]string{},
```
laverya, line 195: could we toss the helmvm version in this to begin with?

laverya (approved):

pkg/addons/adminconsole/customize.go
laverya: importing kubernetes/kubernetes is not to be done lightly IIRC

pkg/addons/adminconsole/customize.go
ricardomaraschini: done.

PR #107, opened 2023-10-16
chore: unify two ci jobs doing the same thing

ricardomaraschini:
both jobs were cleaning runner disk space.


chris-sanders (approved):
lgtm

PR #108, opened 2023-10-16
feat: include installer version to the cluster configuration configmap

ricardomaraschini:
include installer version to the cluster configuration configmap.


chris-sanders (approved):
lgtm

PR #115, opened 2023-10-17
feat: allow for ssh-agent key management

ricardomaraschini:
allows users to bypass providing a ssh key if it has already been loaded / is managed by the ssh agent.


chris-sanders (approved):
Let's try it!

PR #120, opened 2023-10-17
feat: allow users to provide an embedded cluster config path

ricardomaraschini:
allow users to provide through the --overrides flag a path to a file containing an embedded cluster configuration object. the file is then read and its unsupported overrides are used to overrides the generated config.


chris-sanders (approved):
looks good!

PR #122, opened 2023-10-17
chore: removed code to support disconnected environments

ricardomaraschini:
this code is long dead and discontinued. it is not useful to have it laying around when we don't intend to support disconnected installs from the start.

to remove this code allows us to have a cleaner code, this helps at this stage where we are constantly changing things (less friction).


chris-sanders (approved):
Great lgtm

PR #172, opened 2023-11-08
chore: use a license during e2e tests

ricardomaraschini:
in order to capture problems with the embedded license we need to use a version of the embedded cluster binary with an embedded kots license.


.gitignore
```diff
@@ -3,4 +3,5 @@ bundle
 pkg/goods/bins
 pkg/goods/images
 *tgz
-license.yaml
```
laverya, line 6: why get rid of the `license.yaml` gitignore in the root?

laverya: ...because the make command no longer relies on it, got it

e2e/scripts/interactive-multi-node-install.exp
```diff
@@ -21,4 +21,8 @@ expect -re "Add another node?.*:" { send "n\r" }
 set timeout 600
 expect "Enter a new Admin Console password:" { send "password\r" }
 expect "Confirm password:" { send "password\r" }
-expect "Admin Console is ready!"
+expect "Confirm password:" { send "password\r" }
```
laverya, line 24: should we really have to type the password three times?

ricardomaraschini: My mistake. Removed.

laverya (approved):

PR #184, opened 2023-11-17
use custom 'controller' role name if available

laverya:


pkg/addons/embeddedclusteroperator/embeddedclusteroperator.go
```diff
@@ -76,9 +76,9 @@ func (e *EmbeddedClusterOperator) GenerateHelmConfig() ([]v1beta1.Chart, []v1bet
 	return []v1beta1.Chart{chartConfig}, nil, nil
 }
 
-// readEmbeddedClusterConfig reads and unmarshal the Config object from the embedded cluster
+// ReadEmbeddedClusterConfig reads and unmarshal the Config object from the embedded cluster
 // that has been embedded into this binary through a release.
-func (e *EmbeddedClusterOperator) readEmbeddedClusterConfig() (*embeddedclusterv1beta1.Config, error) {
+func (e *EmbeddedClusterOperator) ReadEmbeddedClusterConfig() (*embeddedclusterv1beta1.Config, error) {
```
ricardomaraschini, line 81: Couldn't you call `customization.AdminConsole{}.EmbeddedClusterConfig()` directly ?

laverya: yeah but this does the parsing and I'd rather just have that in one place

investigating that is what led to my "we expect unsupported overrides to be two different formats" comment in slack

laverya: I've updated `customization.AdminConsole{}.EmbeddedClusterConfig()` to return the parsed version, and removed this function entirely

pkg/config/config.go
```diff
@@ -320,21 +325,10 @@ type UnsupportedConfigOverrides struct {
 }
 
 // ApplyEmbeddedUnsupportedOverrides applies the custom configuration to the cluster config.
-func ApplyEmbeddedUnsupportedOverrides(config *v1beta1.Cluster, embconfig []byte) error {
-	if embconfig == nil {
+func ApplyEmbeddedUnsupportedOverrides(config *v1beta1.Cluster, embconfig string) error {
+	if embconfig == "" {
 		return nil
 	}
-	var overrides UnsupportedConfigOverrides
```
ricardomaraschini, line 327: I think we should drop this struct, useless stuff.

ricardomaraschini (dismissed):
If you can remove that struct that would be perfect. Not a blocker.

pkg/config/config.go
laverya: done

ricardomaraschini (approved):

PR #186, opened 2023-11-20
update to embedded-cluster-operator v0.4.1

laverya:


ricardomaraschini:
I wonder how the e2e is working without bumping the Operator Helm Chart version in the Makefile.

laverya:
@ricardomaraschini because if the helm chart isn't valid (and it won't be here AFAICT, whoops) it doesn't block install, just blocks that helm chart being applied

ricardomaraschini:
OH, I see. I thought we have added a test to check the existence and consistence of the Installation/Config object.

ricardomaraschini (approved):

laverya:
Nope, just a test that makes sure that the node has the label we would expect from the controller role

PR #194, opened 2023-11-24
Make config generation and helm reconciliation work with new k0s library

danj-replicated:
In this PR we:

- bump k0s library and binary to latest
- change extension generation logic to remove reliance on yamlv2 package
- introduce ordering to helm charts to ensure storage provider comes up before kots, effectively working around the new non-concurrent reconciler.


danj-replicated:
While this fixes a number of issues and allows us to use the latest k0s library, it somewhat breaks k0sctl ssh based multi-node installs. this is because the k0s helm reconciler is configured in `atomic` mode. `atomic` mode will abort and uninstall a chart if etcd leader election happens at the same time as the chart install.

ricardomaraschini:
As agreed we will be dropping the SSH installations for the time being. @danj-replicated would you mind disabling the failing test so we can get this in ?

ricardomaraschini (approved):

PR #198, opened 2023-11-30
feat: publishing versions as part of the release

ricardomaraschini:
start to publish a file called versions.json with a content similar to the following:

```json
{
        "AdminConsole": "v1.104.3",
        "EmbeddedClusterOperator": "v0.4.1",
        "Installer": "v1.28.4+ec.0-dirty",
        "Kubernetes": "v1.28.4+k0s.0",
        "OpenEBS": "v3.9.0"
}
```

this file will be useful later on when we need to check what versions are included in a given embedded cluster release.

### DIFF
--- a/cmd/embedded-cluster/version.go
+++ b/cmd/embedded-cluster/version.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


cmd/embedded-cluster/version.go
```diff
@@ -14,16 +15,27 @@ import (
...
+				return fmt.Errorf("unable to marshal versions: %w", err)
+			}
+			fmt.Println(string(data))
+			return nil
+		}
 		writer := table.NewWriter()
 		writer.AppendHeader(table.Row{"component", "version"})
-		writer.AppendRow(table.Row{"Installer", defaults.Version})
-		writer.AppendRow(table.Row{"Kubernetes", defaults.K0sVersion})
 		for name, version := range versions {
```
laverya, line 39: is this deterministic? I have to assume it's not

ricardomaraschini:
I will be reopening this as we have other needs that are not accomplished by it.

cmd/embedded-cluster/version.go
laverya: yes

PR #202, opened 2023-12-05
do not require downloading k0s binaries from the internet

laverya:
this would save install time when installing on a different host than the embedded-cluster binary, but that is not currently supported


danj-replicated:
While we're making this work again. something I noticed when I was building with this locally is that k0sctl will try and use a copy of the binary saved locally without the `v` in the name... not sure why, couldn't get it to not so in my custom build I just copied the binary to what it was trying to use. 

we should see about fixing this otherwise it will still download the binary when it can't find it on disk .

ricardomaraschini (approved):

PR #216, opened 2023-12-11
feat: add chart config to version metadata

danj-replicated:


pkg/addons/options.go
```diff
@@ -37,3 +37,9 @@ func WithConfig(config v1beta1.ClusterConfig) Option {
 		a.config = config
 	}
 }
+
+func OnlyDefaults() Option {
```
ricardomaraschini, line 41: I think this needs a comment indicating what "OnlyDefaults" means.

ricardomaraschini (approved):

PR #224, opened 2023-12-18
feat: use binary-name instead of .binary-name

ricardomaraschini:
inside the .config we should be using binary-name and not .binary-name.


danj-replicated (approved):
Fantastic

PR #238, opened 2024-01-02
feat: disregard -ec.X suffix when bumping k0s

ricardomaraschini:
if we are shipping our own version of k0s we should only bump k0s dependency when the prefix (whatever comes before -ec.X) has been updated. we also need to clear the download url used.

### DIFF
--- a/.github/workflows/dependencies.yaml
+++ b/.github/workflows/dependencies.yaml
@@ -33,8 +33,13 @@ jobs:
         sed -i "/^K0SCTL_VERSION/c\K0SCTL_VERSION = $VERSION" Makefile
     - name: K0s
       run: |
+        # Remove the '-ec.X' suffix and only update if the prefix (upstream k0s release) has changed.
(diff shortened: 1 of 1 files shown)


.github/workflows/dependencies.yaml
```diff
@@ -33,8 +33,13 @@ jobs:
         sed -i "/^K0SCTL_VERSION/c\K0SCTL_VERSION = $VERSION" Makefile
     - name: K0s
       run: |
+        # Remove the '-ec.X' suffix and only update if the prefix (upstream k0s release) has changed.
+        export CURVERSION=$(grep ^K0S_VERSION Makefile | cut -d= -f2 | tr -d " " | cut -d - -f1)
```
danj-replicated, line 37: This could be simplified to: `awk '/^K0S_VERSION/{split($3,a,"-"); print a[1]}' Makefile`

danj-replicated (changes requested):

.github/workflows/dependencies.yaml
ricardomaraschini: Added.

danj-replicated (approved):

PR #243, opened 2024-01-05
feat: apply config overrides during node join

ricardomaraschini:
starts to apply patches to both k0s' config `api` and `storage` properties. patches are received from the kots api and applied in the following order: embedded overrides -> user provided.

this commit also adds unit tests for the new functionality and to the old pathing functions.


pkg/config/testdata/k0s-config/overide-helm-charts.yaml
```diff
@@ -0,0 +1,48 @@
...
+          - chartname: another/chart
+            version: 1.0.0
+            values: |-
+              test: 123
+              another-property: 123
+          - chartname: yet-another/chart
+            version: 2.0.0
+            values: |-
+              test: 321
+              another-property: 321
```
laverya, line 48: Should the desired behavior be to keep the openebs chart too? Honestly not sure

laverya (approved):

pkg/config/testdata/k0s-config/overide-helm-charts.yaml
ricardomaraschini: This is a known drawback of how the overrides work today. At this stage arrays are completely overwritten.

PR #263, opened 2024-01-12
feat: install kotsadm in default namespace, not embedded-cluster

laverya:
This will probably break upgrades, but we really shouldn't be placing kotsadm and the operator in the same namespace. As such, the sooner we make the change the better, and I should have done this weeks ago.

### DIFF
--- a/e2e/scripts/unsupported-overrides.sh
+++ b/e2e/scripts/unsupported-overrides.sh
@@ -41,7 +41,7 @@ spec:
                 version: 0.13.0
               - chartname: oci://registry.replicated.com/library/admin-console
                 name: admin-console
-                namespace: embedded-cluster
+                namespace: default
                 order: 3
                 version: 1.105.1
                 values: |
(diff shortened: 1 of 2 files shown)


e2e/scripts/unsupported-overrides.sh
```diff
@@ -41,7 +41,7 @@ spec:
                 version: 0.13.0
               - chartname: oci://registry.replicated.com/library/admin-console
                 name: admin-console
-                namespace: embedded-cluster
+                namespace: default
```
chris-sanders, line 44: I'm not clear on the change, but we should not use default.

pkg/addons/applier.go
```diff
@@ -135,7 +135,7 @@ func (a *Applier) load() (map[string]AddOn, error) {
 		addons["embeddedclusteroperator"] = embedoperator
 	}
 	if _, disabledAddons := a.disabledAddons["adminconsole"]; !disabledAddons {
-		aconsole, err := adminconsole.New("embedded-cluster", a.prompt, a.config)
+		aconsole, err := adminconsole.New("default", a.prompt, a.config)
```
chris-sanders, line 138: same

chris-sanders (changes requested):

chris-sanders (approved):
I like it

PR #276, opened 2024-01-17
feat: reset command

danj-replicated:
TODO: 
- [x] quorum safey for controllers

removing binaries and extra files can be looked at later along with the other efforts towards that task.

### DIFF
--- a/cmd/embedded-cluster/reset.go
+++ b/cmd/embedded-cluster/reset.go
@@ -120,7 +120,7 @@ func (h *hostInfo) checkQuorumSafety(c *cli.Context) (bool, string, error) {
 		return true, "", nil
 	}
 	if len(etcd.Members) == 3 {
-		return false, "cluster has 3 nodes, removing this node will cause etcd to lose quorum", nil
+		return false, "cluster has 3 control-plane nodes, removing this node will cause etcd to lose quorum", nil
 	}
 	if len(etcd.Members)%2 != 0 {
 		return false, "cluster would have even number of control-plane nodes after resetting this node, this could cause etcd to become unstable", nil
--- a/cmd/embedded-cluster/reset.go
+++ b/cmd/embedded-cluster/reset.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/replicatedhq/embedded-cluster/pkg/defaults"
+	"github.com/replicatedhq/embedded-cluster/pkg/prompts"
 )
 
 type etcdMembers struct {
@@ -157,6 +158,18 @@ func newHostInfo(ctx context.Context) (hostInfo, error) {
 	return currentHost, nil
 }
 
+func checkErrPrompt(err error) bool {
+	if err == nil {
+		return true
+	}
+	fmt.Println("-----")
+	fmt.Println(err)
+	fmt.Println("-----")
+	fmt.Println("An error has occured while trying to reset this node.")
+	fmt.Println("Continuing may leave the cluster in an unexpected state")
+	return prompts.New().Confirm("Do you want to continue anyway?", false)
+}
+
 var resetCommand = &cli.Command{
 	Name:  "reset",
 	Usage: "Reset the node this command is run from",
@@ -173,16 +186,14 @@ var resetCommand = &cli.Command{
 		// drain node
 		fmt.Println("draining node...")
 		err = currentHost.drainNode()
-		if err != nil {
-			fmt.Println(err)
+		if !checkErrPrompt(err) {
 			return nil
 		}
 
 		// remove node from cluster
 		fmt.Println("removing node from cluster...")
 		err = currentHost.Kclient.Delete(c.Context, &currentHost.Node)
-		if err != nil {
-			fmt.Println(err)
+		if !checkErrPrompt(err) {
 			return nil
 		}
 
@@ -192,16 +203,14 @@ var resetCommand = &cli.Command{
 			// delete controlNode object from cluster
 			fmt.Println("deleting controlNode...")
 			err := currentHost.Kclient.Delete(c.Context, &currentHost.ControlNode)
-			if err != nil {
-				fmt.Println(err)
+			if !checkErrPrompt(err) {
 				return nil
 			}
 
 			// try and leave etcd cluster
 			fmt.Println("leaving etcd cluster...")
 			err = currentHost.leaveEtcdcluster()
-			if err != nil {
-				fmt.Println(err)
+			if !checkErrPrompt(err) {
 				return nil
 			}
 
@@ -213,16 +222,14 @@ var resetCommand = &cli.Command{
 		// stop k0s
 		fmt.Printf("stopping %s...\n", binName)
 		err = currentHost.stopK0s()
-		if err != nil {
-			fmt.Println(err)
+		if !checkErrPrompt(err) {
 			return nil
 		}
 
 		// reset local node
 		fmt.Printf("resetting %s...\n", binName)
 		err = currentHost.resetK0s()
-		if err != nil {
-			fmt.Println(err)
+		if !checkErrPrompt(err) {
 			return nil
 		}
 
--- a/cmd/embedded-cluster/reset.go
+++ b/cmd/embedded-cluster/reset.go
@@ -143,6 +143,12 @@ var resetCommand = &cli.Command{
 				return nil
 			}
 
+      err := kcli.Delete(c.Context,&controlNode)
+      if err !=nil {
+        fmt.Println(err)
+        return nil
+      }
+
 			etcdPeerAddress, err := getEtcdMemberAddress(hostname)
 			if err != nil {
 				fmt.Println(err)
--- a/cmd/embedded-cluster/reset.go
+++ b/cmd/embedded-cluster/reset.go
@@ -32,8 +32,8 @@ type hostInfo struct {
 }
 
 var (
-  k0s = defaults.K0sBinaryPath()
-  binName = defaults.BinaryName()
+	k0s     = defaults.K0sBinaryPath()
+	binName = defaults.BinaryName()
 )
 
 // getEtcdMemberAddress uses k0s to obtain the etcd member address for the given hostname
--- a/cmd/embedded-cluster/reset.go
+++ b/cmd/embedded-cluster/reset.go
@@ -121,7 +121,7 @@ func (h *hostInfo) stopAndResetK0s() error {
 	if err != nil {
 		return fmt.Errorf("could not reset k0s: %w, %s", err, string(out))
 	}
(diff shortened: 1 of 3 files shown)


cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,187 @@
...
+			return nil
+		}
+
+		tempKubeConfig, err := os.CreateTemp("", "*")
+		if err != nil {
+			fmt.Println(err)
+			return nil
+		}
+		defer tempKubeConfig.Close()
+		defer os.Remove(tempKubeConfig.Name())
```
laverya, line 74: I believe this is backwards
deferred functions are Last-In-First-Out so you're removing the file before closing it

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,187 @@
...
+			fmt.Println(err)
+			return nil
+		}
+
+		isControlPlane := false
+		labels := node.GetLabels()
+		_, ok := labels["node-role.kubernetes.io/control-plane"]
+		if ok && labels["node-role.kubernetes.io/control-plane"] == "true" {
+			isControlPlane = true
+		}
```
laverya, line 109: this is equivalent - the default string value if it's not populated is `""`
```suggestion
		if labels["node-role.kubernetes.io/control-plane"] == "true" {
			isControlPlane = true
		}
```

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,187 @@
...
+
+		// drain controller node
+		if isControlPlane {
```
laverya, line 112: is there a reason not to drain non-controller nodes?

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,299 @@
...
+	etcd := etcdMembers{}
+	err = json.Unmarshal(out, &etcd)
+	if err != nil {
+		return false, "", fmt.Errorf("unable to unmarshal etcd member list, %w, %s", err, out)
+	}
+	if len(etcd.Members) < 3 {
+		return true, "", nil
+	}
+	if len(etcd.Members) == 3 {
+		return false, "cluster has 3 nodes, removing this node will cause etcd to lose quorum", nil
```
laverya, line 123: From standup
```suggestion
		return false, "cluster has 3 primary nodes, removing this node will cause etcd to lose quorum", nil
```

laverya (approved):

danj-replicated:
merging this as is, will iterate later on:
- better safety for waiting on etcd rebuilds when downscaling below 3 nodes
- tests

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,299 @@
...
+
+func checkErrPrompt(err error) bool {
+	if err == nil {
+		return true
+	}
+	fmt.Println("-----")
+	fmt.Println(err)
+	fmt.Println("-----")
+	fmt.Println("An error has occured while trying to reset this node.")
+	fmt.Println("Continuing may leave the cluster in an unexpected state")
```
ajp-io, line 191: Let's be consistent with periods and use them at the end of sentences. There are a few other cases without them too.

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,299 @@
...
+	return prompts.New().Confirm("Do you want to continue anyway?", false)
+}
+
+var resetCommand = &cli.Command{
+	Name: "reset",
+	Flags: []cli.Flag{
+		&cli.BoolFlag{
+			Name:   "yes-really-reset",
```
ajp-io, line 199: I know we're inspired by Rook here, but I think we should make this `--confirm` instead. If this was only a Replicated CLI, that would be one thing. But every vendor ships this CLI and has to use these flags, and I think sticking to convention will have broader appeal.

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,299 @@
...
+			Hidden: true,
+			Value:  false,
+		},
+	},
+	Usage: "Reset the node this command is run from",
```
ajp-io, line 204: Maybe "Reset the current node". I'm not sure of the right wording, but "the node this command is run from" sounds off to me.

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,299 @@
...
+	Action: func(c *cli.Context) error {
+
+		fmt.Println("This command will completely reset this node, removing it from the cluster")
```
ajp-io, line 207: This will remove this node from the cluster and completely reset it.

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,299 @@
...
+		if !prompts.New().Confirm("Do you want to continue?", false) {
+			fmt.Println("aborting.")
+			return nil
+		}
+
+		fmt.Println("gathering facts...")
```
ajp-io, line 213: I don't know that this line adds much other than cluttering the output. As I recall in the demo, we reached the next step pretty quickly.

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,299 @@
...
+func (h *hostInfo) stopAndResetK0s() error {
+	out, err := exec.Command(k0s, "stop").CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("could not stop k0s service: %w, %s", err, string(out))
+	}
+	out, err = exec.Command(k0s, "reset").CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("could not reset k0s: %w, %s", err, string(out))
+	}
+	fmt.Println("Node has been reset, please reboot to ensure transient configuration is also reset")
```
ajp-io, line 150: Node has been reset. Please reboot to ensure transient configuration is also reset.

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,299 @@
...
+		// basic check to see if it's safe to remove this node from the cluster
+		if currentHost.isControlPlane() {
+			safeToRemove, reason, err := currentHost.checkQuorumSafety(c)
+			if err != nil {
+				fmt.Println(err)
+				return nil
+			}
+			if !safeToRemove {
+				fmt.Println(reason)
+				fmt.Println("run reset command again with --yes-really-reset to ignore this")
```
ajp-io, line 230: Rerun reset command command with `--confirm` to ignore this.

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,299 @@
...
+		}
+
+		// determine if this is the only node in the cluster
+		// if this is a single node we can skip a lot of steps
+		nodeList := corev1.NodeList{}
+		currentHost.Kclient.List(c.Context, &nodeList)
+		if len(nodeList.Items) == 1 {
+			nodeName := nodeList.Items[0].Name
+			if nodeName != currentHost.Hostname {
+				fmt.Println("detected a single node cluster, but the node's name doesn't match our hostname")
```
ajp-io, line 242: Detected a single-node cluster, but the node's name doesn't match our hostname.

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,299 @@
...
+				return nil
+			}
+			// stop k0s
+			fmt.Printf("resetting %s...\n", binName)
```
ajp-io, line 246: Capital R

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,299 @@
...
+			err = currentHost.stopAndResetK0s()
+			if !checkErrPrompt(err) {
+				return nil
+			}
+			return nil
+		}
+
+		// drain node
+		fmt.Println("draining node...")
```
ajp-io, line 255: Capital D

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,299 @@
...
+		err = currentHost.drainNode()
+		if !checkErrPrompt(err) {
+			return nil
+		}
+
+		// remove node from cluster
+		fmt.Println("removing node from cluster...")
```
ajp-io, line 262: Capital R

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,299 @@
...
+
+			// delete controlNode object from cluster
+			fmt.Println("deleting controlNode...")
+			err := currentHost.Kclient.Delete(c.Context, &currentHost.ControlNode)
+			if !checkErrPrompt(err) {
+				return nil
+			}
+
+			// try and leave etcd cluster
+			fmt.Println("leaving etcd cluster...")
```
ajp-io, line 279: Capital L

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,299 @@
...
+				return nil
+			}
+
+		} else if err != nil {
+			fmt.Println(err)
+			return nil
+		}
+
+		// reset
+		fmt.Printf("resetting %s...\n", binName)
```
ajp-io, line 291: Capital R

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+func checkErrPrompt(err error) bool {
+	if err == nil {
+		return true
+	}
+	fmt.Println("-----")
+	fmt.Println(err)
+	fmt.Println("-----")
+	fmt.Println("An error has occured while trying to reset this node.")
```
ajp-io, line 190: Remove "has"

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,299 @@
...
+			Hidden: true,
+			Value:  false,
+		},
+	},
+	Usage: "Reset the node this command is run from",
+	Action: func(c *cli.Context) error {
+
+		fmt.Println("This command will completely reset this node, removing it from the cluster")
+		if !prompts.New().Confirm("Do you want to continue?", false) {
+			fmt.Println("aborting.")
```
ajp-io, line 209: Capital A

cmd/embedded-cluster/reset.go
```diff
@@ -0,0 +1,299 @@
...
+		err = currentHost.Kclient.Delete(c.Context, &currentHost.Node)
+		if !checkErrPrompt(err) {
+			return nil
+		}
+
+		// controller pre-reset
+		if currentHost.isControlPlane() {
+
+			// delete controlNode object from cluster
+			fmt.Println("deleting controlNode...")
```
ajp-io, line 272: Capital D.

I assume this is the same controlnode as [here](https://docs.k0sproject.io/v1.28.4+k0s.0/autopilot/?h=controlnode#q-how-do-i-apply-the-plan-and-controlnode-crds). If so, let's capitalize ControlNode like k0s does.

PR #279, opened 2024-01-18
feat: create library for embedding kots release

ricardomaraschini:
created a library so we can use it on airgap project. this library
allows for embedding kots releases into an embedded cluster binary.


pkg/embed/append.go
```diff
@@ -0,0 +1,175 @@
...
+		binfp.Close()
+		return nil, 0, fmt.Errorf("unable to open release file: %w", err)
+	}
+	relstat, err := relfp.Stat()
+	if err != nil {
+		binfp.Close()
+		relfp.Close()
+		return nil, 0, fmt.Errorf("unable to stat release file: %w", err)
+	}
+	// blocks releases larger than 100MB.
```
laverya, line 161: is this the largest size that can be parsed by the 'calculate rewind' function, or is it here for some other reason?

pkg/embed/append.go
```diff
@@ -0,0 +1,175 @@
...
+	if relstat.Size() > 104857600 {
```
laverya, line 162: the endmark format is `%s%010d` - IMO this implies that the size should be all `9`s and not this number?

pkg/embed/append.go
laverya: looks like 10^11-1 bytes is intended to be the maximum?

pkg/embed/append.go
```diff
@@ -0,0 +1,175 @@
...
+			}
+		}
+	}
+	return fs, nil
+}
+
+// Build builds the binary with the embedded release. Returns an io.ReadCloser
+// from where the binary can be read. Returns the ReadCloser, the total size of
+// the binary and an error if any.
+func Build(bin string, release string) (io.ReadCloser, int64, error) {
```
laverya, line 141: I would _prefer_ to have the inputs be readclosers here too, that way this can be entirely streamed and not require writing to disk (we would then be able to skip the airgap builder service)

laverya: We would have to get the sizes with a wrapper
```
type wrapper struct {
    io.Reader
    n int
}

func (w *wrapper) Read(p []byte) (int, error) {
    n, err := w.Reader.Read(p)
    w.n += n
    return n, err
}
```

pkg/embed/append.go
ricardomaraschini: Yes, that 100MB was completely random as I never imagined someone using more than that size of YAML files. It is fixed. Please take a look.

ricardomaraschini: I am resolving this because these files were removed from the PR. Now we have only way of building these binaries.

laverya (approved):

laverya (approved):

laverya (approved):

PR #281, opened 2024-01-19
feat: improve node rests 

danj-replicated:
- workers can now drain and remove themselves from the cluster
- warn about removing the last controller node in a cluster with dependant worker nodes


ricardomaraschini:
We gonna need an e2e for this.

cmd/embedded-cluster/reset.go
```diff
@@ -116,14 +116,33 @@ func (h *hostInfo) checkQuorumSafety(c *cli.Context) (bool, string, error) {
...
+	}
+	if len(workers) > 0 && len(controllers) == 1 {
+		return false, "Cluster has a worker node joined and you are attempting to remove the only controller node.", nil
+	}
 	if len(etcd.Members) < 3 {
 		return true, "", nil
 	}
 	if len(etcd.Members) == 3 {
-		return false, "cluster has 3 control-plane nodes, removing this node will cause etcd to lose quorum", nil
+		return false, "Cluster has 3 control-plane nodes, removing this node will cause etcd to lose quorum.", nil
```
ajp-io, line 142: Instead of control plane, can we show the controller node role name instead?
```suggestion
		return false, "Cluster has 3 control plane nodes. Removing this node will cause etcd to lose quorum.", nil
```

cmd/embedded-cluster/reset.go
```diff
@@ -116,14 +116,33 @@ func (h *hostInfo) checkQuorumSafety(c *cli.Context) (bool, string, error) {
...
+	for _, node := range nodeList.Items {
+		labels := node.GetLabels()
+		if labels["node-role.kubernetes.io/control-plane"] == "true" {
+			controllers = append(controllers, node.Name)
+		} else {
+			workers = append(workers, node.Name)
+		}
+	}
+	if len(workers) > 0 && len(controllers) == 1 {
+		return false, "Cluster has a worker node joined and you are attempting to remove the only controller node.", nil
```
ajp-io, line 136: A few thoughts.
* From an end-user perspective, controller is a node role, but the vendor might have [customized its name](https://docs.google.com/document/d/1NKXIj7fPhYc6BadscPznXMcbXVj74tBfY-xbNJqbVGk/edit#heading=h.bqi1aw166qh7). Can we show the actual node role name here instead of saying controller?
* Do you think we should tell them they can reset the workers first if they really want to reset this node, or should we just tell them the issue?
```suggestion
		return false, "Cannot reset the last controller node when there are other nodes in the cluster.", nil
```

cmd/embedded-cluster/reset.go
```diff
@@ -116,14 +116,33 @@ func (h *hostInfo) checkQuorumSafety(c *cli.Context) (bool, string, error) {
...
 	if len(etcd.Members) < 3 {
 		return true, "", nil
 	}
 	if len(etcd.Members) == 3 {
-		return false, "cluster has 3 control-plane nodes, removing this node will cause etcd to lose quorum", nil
+		return false, "Cluster has 3 control-plane nodes, removing this node will cause etcd to lose quorum.", nil
 	}
 	if len(etcd.Members)%2 != 0 {
-		return false, "cluster would have even number of control-plane nodes after resetting this node, this could cause etcd to become unstable", nil
+		return false, "Cluster would have even number of control-plane nodes after resetting this node, this could cause etcd to become unstable.", nil
```
ajp-io, line 145: Same comment about controller node role name if possible.

What do you think about saying it would leave the cluster unstable instead of etcd, since we're abstracting away the Kubernetes stuff, and the user likely doesn't know about etcd?
```suggestion
		return false, "This will leave the cluster with an even number of control plane nodes, which could make it unstable.", nil
```

cmd/embedded-cluster/reset.go
```diff
@@ -170,7 +198,7 @@ func newHostInfo(ctx context.Context) (hostInfo, error) {
 		return currentHost, err
 	}
 	// control plane only stff
-	if currentHost.isControlPlane() {
+	if currentHost.Status.Role == "controller" {
```
ajp-io, line 201: Does this factor in the case where the vendor changed the controller node role name?

cmd/embedded-cluster/reset.go
```diff
@@ -187,47 +215,59 @@ func checkErrPrompt(err error) bool {
 	fmt.Println("-----")
 	fmt.Println(err)
 	fmt.Println("-----")
-	fmt.Println("An error has occured while trying to reset this node.")
-	fmt.Println("Continuing may leave the cluster in an unexpected state")
+	fmt.Println("An error occured while trying to reset this node")
+	fmt.Println("continuing may leave the cluster in an unexpected state.")
```
ajp-io, line 219:
```suggestion
	fmt.Println("An error occurred while trying to reset this node.")
	fmt.Println("Continuing may leave the cluster in an unexpected state.")
```

cmd/embedded-cluster/reset.go
```diff
@@ -187,47 +215,59 @@ func checkErrPrompt(err error) bool {
...
+		if currentHost.Status.Role == "controller" {
 			safeToRemove, reason, err := currentHost.checkQuorumSafety(c)
 			if err != nil {
 				fmt.Println(err)
 				return nil
 			}
 			if !safeToRemove {
 				fmt.Println(reason)
-				fmt.Println("run reset command again with --yes-really-reset to ignore this")
+				fmt.Println("run reset command again with --confirm to ignore this")
```
ajp-io, line 270:
```suggestion
				fmt.Println("Run reset command command with --confirm to ignore this.")
```

ajp-io (commented):
A few other wording suggestions

cmd/embedded-cluster/reset.go
danj-replicated: @ajp-io this specific role is set by k0s, there is only "controller" and "worker"

cmd/embedded-cluster/reset.go
```diff
@@ -169,66 +201,85 @@ func newHostInfo(ctx context.Context) (hostInfo, error) {
...
+	}
 	return prompts.New().Confirm("Do you want to continue anyway?", false)
 }
 
 var resetCommand = &cli.Command{
 	Name: "reset",
 	Flags: []cli.Flag{
 		&cli.BoolFlag{
-			Name:   "yes-really-reset",
+			Name:   "confirm",
```
ajp-io, line 241: How come this one and force are hidden?

cmd/embedded-cluster/reset.go
```diff
@@ -169,66 +201,85 @@ func newHostInfo(ctx context.Context) (hostInfo, error) {
...
+			Hidden: true,
+			Value:  false,
+		},
+		&cli.BoolFlag{
+			Name:  "no-prompt",
+			Usage: "Do not prompt user when it is not necessary",
+			Value: false,
+		},
+		&cli.BoolFlag{
+			Name:   "force",
```
ajp-io, line 251: Remind me what exactly force does. Looks like it skips all the checks that it's safe to remove the node. Is that right? And in what situation might someone use that as opposed to `--confirm`?

ajp-io (commented):
A couple questions

cmd/embedded-cluster/reset.go
danj-replicated: they're hidden because they're potentially dangerous and exist as break-glass flags. 

by making them visible in the help text a user might attempt to run with them by default and skip any pre-reset checks

cmd/embedded-cluster/reset.go
danj-replicated: correct, `--force` skips all checks and does not attempt to contact the cluster for clean removal, this flag provides the ability to reset nodes in the following scenarios:

- the control-plane no longer exists
- the node has no internet connectivity
- the cluster is in such a state where a clean exit is not possible

cmd/embedded-cluster/reset.go
danj-replicated: thinking about it, maybe `--force` should be visible but explain the difference in a prompt

danj-replicated: https://github.com/replicatedhq/embedded-cluster/pull/281/commits/2dae3c18280705dddc4de229441b6ed05dbe27f0

cmd/embedded-cluster/reset.go
```diff
@@ -147,151 +216,134 @@ func (h *hostInfo) stopAndResetK0s() error {
...
 	}
-	fmt.Println("-----")
 	fmt.Println(err)
-	fmt.Println("-----")
-	fmt.Println("An error has occured while trying to reset this node.")
-	fmt.Println("Continuing may leave the cluster in an unexpected state")
+	if c.Bool("force") || c.Bool("no-prompt") {
+		return true
+	}
+	fmt.Println("An error occurred while trying to reset this node.")
```
ajp-io, line 266: Sorry if I'm totally missing it, but I thought `no-prompt` was going to bail if an error occurred, rather than just continuing through it. It would just skip confirmation type prompts. (I could have misunderstood too.)

cmd/embedded-cluster/reset.go
```diff
@@ -147,151 +216,134 @@ func (h *hostInfo) stopAndResetK0s() error {
...
-			Name:   "yes-really-reset",
-			Hidden: true,
-			Value:  false,
+			Name:  "no-prompt",
+			Usage: "Do not prompt user when it is not necessary",
+			Value: false,
+		},
+		&cli.BoolFlag{
+			Name:  "force",
+			Usage: "Ignore errors encountered when resetting the node, implies --no-prompt",
```
ajp-io, line 281:
```suggestion
			Usage: "Ignore errors encountered when resetting the node. Implies --no-prompt.",
```

cmd/embedded-cluster/reset.go
```diff
@@ -147,151 +216,134 @@ func (h *hostInfo) stopAndResetK0s() error {
...
 
 var resetCommand = &cli.Command{
 	Name: "reset",
 	Flags: []cli.Flag{
 		&cli.BoolFlag{
-			Name:   "yes-really-reset",
-			Hidden: true,
-			Value:  false,
+			Name:  "no-prompt",
+			Usage: "Do not prompt user when it is not necessary",
```
ajp-io, line 276:
```suggestion
			Usage: "Disable interactive prompts",
```

ajp-io (commented):
lgtm other than these comments

cmd/embedded-cluster/reset.go
danj-replicated: No you're correct, that's the intended result. 

looking into this

danj-replicated: Should be fixed

cmd/embedded-cluster/reset.go
```diff
@@ -23,20 +22,64 @@ type etcdMembers struct {
...
 
 var (
 	binName = defaults.BinaryName()
 	k0s     = defaults.K0sBinaryPath()
 )
 
+// deleteNode removes the node from the cluster
+func (h *hostInfo) deleteNode(ctx context.Context) error {
+	if h.KclientError != nil {
+		return fmt.Errorf("unable to delete Node: %w", h.KclientError)
```
laverya, line 53: a personal preference is that each error message should be unique in some way

laverya (approved):

PR #285, opened 2024-01-19
feat: e2e tests for multi node clusters

ricardomaraschini:
- Create scripts to generate worker and controller join token through Kots web interface.
- Create e2e test that deploys a multi-node cluster (3 controllers + 1 worker).

### DIFF
--- a/e2e/scripts/generate-controller-join-token.js
+++ b/e2e/scripts/generate-controller-join-token.js
@@ -22,14 +22,14 @@ const puppeteer = require('puppeteer'); // v20.7.4 or later
     page.setDefaultTimeout(timeout);
     const args = process.argv.slice(2);
     if (args.length !== 1) {
-        throw new Error('usage: check-app-and-cluster-status.js <kotsadm-ip>');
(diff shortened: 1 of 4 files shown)


e2e/scripts/generate-controller-join-token.js
```diff
@@ -0,0 +1,272 @@
...
+        const targetPage = page;
+        await targetPage.waitForSelector('.react-prism.language-bash');
+        let elementContent = await targetPage.evaluate(() => {
+            const element = document.querySelector('.react-prism.language-bash');
+            return element ? element.textContent : null;
+        });
+        if (!elementContent) {
+            throw new Error("Could not find the node join command");
+        }
+        console.log(elementContent);
```
laverya, line 264: Could we log a json object instead to simplify error logging in the future?

laverya (dismissed):

e2e/scripts/generate-controller-join-token.js
ricardomaraschini: Done. While doing it I also added a script to generate a worker token as well.

laverya (approved):

PR #303, opened 2024-01-26
set the version to 'dev-metadata' in integration tests

laverya:
also: use a new staging test app, and run integration tests on merge


.github/workflows/pull-request.yaml
```diff
@@ -71,3 +62,38 @@ jobs:
       run: |
         export EC_VERSION="dev-metadata"
         ./scripts/cache-files.sh
+
+    - name: Cache files for integration test
```
ricardomaraschini, line 66: Why is this step duplicated ?

e2e/install_test.go
```diff
@@ -70,6 +70,15 @@ func TestSingleNodeInstallation(t *testing.T) {
...
 	}
+	t.Log("checking installation state")
+	line = []string{"check-installation-state.sh"}
+	stdout, stderr, err = RunCommandOnNode(t, tc, 0, line)
+	if err != nil {
+		t.Log("stdout:", stdout)
+		t.Log("stderr:", stderr)
+		t.Fatalf("fail to check installation state: %v", err)
+	}
+	t.Log("installation state:", stdout)
```
ricardomaraschini, line 81: Is this just so we can see the output after the test run ?

.github/workflows/pull-request.yaml
laverya: Because it's cashing to two different places

e2e/install_test.go
laverya: Yes - I would prefer to have a check for the actual installation state, but it isn't failing yet 😆

ricardomaraschini (approved):

PR #311, opened 2024-01-31
Revert "feat: require uid 0 for install, shell and node commands"

ricardomaraschini:
Reverts replicatedhq/embedded-cluster#306


ricardomaraschini:
We have decided not to revert this but try to fix the issue moving forward.

PR #328, opened 2024-02-06
test upgrading the app to add + change vendor addons

laverya:

PR #331, opened 2024-02-07
support expanding nodeport range

laverya:


laverya:
this was included in https://github.com/replicatedhq/embedded-cluster/pull/328

PR #356, opened 2024-02-14
feat: polishing cli arguments and help messages

ricardomaraschini:
polishing cli arguments and help messages

### DIFF
--- a/cmd/embedded-cluster/install.go
+++ b/cmd/embedded-cluster/install.go
@@ -265,7 +265,7 @@ func runOutro(c *cli.Context) error {
 // file is created. Resulting kubeconfig is stored in the configuration dir.
 var installCommand = &cli.Command{
 	Name:  "install",
-	Usage: "Installs and starts a new cluster",
+	Usage: "Install a cluster",
 	Before: func(c *cli.Context) error {
 		if os.Getuid() != 0 {
 			return fmt.Errorf("install command must be run as root")
@@ -280,7 +280,7 @@ var installCommand = &cli.Command{
 		},
 		&cli.StringSliceFlag{
 			Name:  "disable-addon",
-			Usage: "Disable addon during install/upgrade",
+			Usage: "Disable addon during install",
 		},
 		&cli.StringFlag{
 			Name:   "overrides",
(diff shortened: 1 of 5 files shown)


laverya (dismissed):

cmd/embedded-cluster/install.go
```diff
@@ -265,7 +265,7 @@ func runOutro(c *cli.Context) error {
 // file is created. Resulting kubeconfig is stored in the configuration dir.
 var installCommand = &cli.Command{
 	Name:  "install",
-	Usage: "Installs and starts a new cluster",
+	Usage: "Install a cluster",
```
ajp-io, line 268: This is more than a cluster to the end user. In fact, they think of it as the app more so than the cluster. WDYT about "Install \<binary name>"? Maybe "Install \<binary name> cluster", but in a single-node situation, the word cluster feels confusing.

This lines up with the `Install and manage %s", name` wording you used below too.

cmd/embedded-cluster/install.go
```diff
@@ -280,7 +280,7 @@ var installCommand = &cli.Command{
 		},
 		&cli.StringSliceFlag{
 			Name:  "disable-addon",
-			Usage: "Disable addon during install/upgrade",
+			Usage: "Disable addon during install",
```
ajp-io, line 283: But I'm all for just getting rid of this one
```suggestionx
			Usage: "Disable add-on during install",
```

cmd/embedded-cluster/reset.go
```diff
@@ -299,7 +299,7 @@ var resetCommand = &cli.Command{
 		},
 		&cli.BoolFlag{
 			Name:  "force",
-			Usage: "Ignore errors encountered when resetting the node. Implies --no-prompt.",
+			Usage: "Ignore errors encountered when resetting the node. Implies --no-prompt",
```
ajp-io, line 302: If we're going to remove periods, let's do something like this
```suggestion
			Usage: "Ignore errors encountered when resetting the node (implies --no-prompt)",
```

ajp-io (commented):
I wonder if we need the `help` command if we have `--help, -h` already. Nbd, just thinking out loud.

cmd/embedded-cluster/install.go
ricardomaraschini: This will be done in a separate PR.

cmd/embedded-cluster/install.go
ricardomaraschini: OK, `Install <binary name>` it is.

ricardomaraschini:
```suggestion
	Usage: fmt.Sprintf("Install %s", defaults.BinaryName()),
```

laverya (approved):

PR #365, opened 2024-02-14
update chart values so that charts are not immediately reapplied

laverya:


laverya:
The disadvantage of this is that it keeps the password preserved in the chart values forever
I'm going to look at creating that secret from embedded-cluster instead

laverya:
Looks like all that was actually needed was just not restarting k0s

PR #367, opened 2024-02-15
remove k0s restart in CI

laverya:


laverya:
This restart seems to be the cause of the issues we've been seeing where charts will fail to create due to the name already being in use etc

e2e/kots-release-install/cluster-config.yaml
```diff
@@ -38,7 +38,7 @@ spec:
           chartname: ingress-nginx/ingress-nginx
           namespace: ingress-nginx
           version: "4.8.3"
-          order: 2
+          order: 4
```
laverya, line 41: This chart being added with order 4 almost always triggered the race condition before

ricardomaraschini (approved):

PR #368, opened 2024-02-15
wait for all charts before completing

laverya:


laverya:
oh of course that object isn't actually persistent
I'm going to table this in favor of just getting rid of the restart

PR #373, opened 2024-02-15
download the build from staging.replicated.app and install upon merge

laverya:


e2e/scripts/vandoor-prepare.sh
```diff
@@ -0,0 +1,18 @@
...
+
+
+main() {
+    local sha=
+    sha="$1"
+
+    apt-get update
+    apt-get install curl ca-certificates -y
+
+    curl "https://staging.replicated.app/embedded/embedded-cluster-smoke-test-staging-app/ci/${sha}" -H 'Authorization: 2cHePFvlGDksGOKfTicouqJKGtZ' -o ec-release.tgz
```
sgalsaleh, line 12: Should we invalidate this license, create a new one and store it as a Github secret?

laverya: this license file is stored in the repo
so, maybe?
but I don't think we care that this one is exposed

sgalsaleh: ah i see. ok then.

sgalsaleh (approved):

PR #374, opened 2024-02-15
Set embedded cluster version Helm value for the adminconsole addon

sgalsaleh:


pkg/addons/adminconsole/adminconsole.go
```diff
@@ -39,7 +39,7 @@ var (
 )
 
 // protectedFields are helm values that are not overwritten when upgrading the addon.
-var protectedFields = []string{"automation", "embeddedClusterID"}
+var protectedFields = []string{"automation", "embeddedClusterID", "embeddedClusterVersion"}
```
laverya, line 42: this should not be protected, doing so will prevent it from changing when we update embedded-cluster

pkg/addons/adminconsole/adminconsole.go
```diff
@@ -53,7 +53,8 @@ var helmValues = map[string]interface{}{
 		"enabled":  true,
 		"nodePort": DEFAULT_ADMIN_CONSOLE_NODE_PORT,
 	},
-	"embeddedClusterID": metrics.ClusterID().String(),
+	"embeddedClusterID":      metrics.ClusterID().String(),
+	"embeddedClusterVersion": defaults.Version,
```
laverya, line 57: IIRC this isn't actually used for anything besides populating a configmap?

laverya: Is that configmap actually used for something?

sgalsaleh: it's passed as an environment variable to the kotsadm deployment for reporting purposes

pkg/addons/adminconsole/adminconsole.go
sgalsaleh: whoops. good catch.

pkg/addons/adminconsole/adminconsole.go
sgalsaleh: https://github.com/replicatedhq/kots-helm/pull/35

sgalsaleh: alternatively, we can store the embedded cluster in the cluster somewhere and have KOTS detect it from there. not sure which approach is better.

laverya (approved):

pkg/addons/adminconsole/adminconsole.go
laverya: oh gotcha
yeah feel free to fix this then

PR #377, opened 2024-02-16
feat: use k0s kubeconfig

laverya:


ricardomaraschini:
Autocomplete seems to have stopped to work with this.

chris-sanders (changes requested):
Actually I don't want to use any files in home at all.

See discussion here: https://app.shortcut.com/replicated/story/94401/remove-all-unnecessary-files-created-on-worker-node-hosts-during-installation#activity-97013

Summary: kubeconfig should not be needed at all. Use `k0s kubeconfig` with path settings which handles kubeconfig from the `/var` directories already.

chris-sanders (approved):
Much better, looks good!

PR #385, opened 2024-02-17
feat: only print "running preflights" if preflights exist

ricardomaraschini:
- only print "running preflights" if preflights exist.
- rename progressbar to spinner (it does not indicate progress neither is a bar).


ricardomaraschini:
replaced by https://github.com/replicatedhq/embedded-cluster/pull/386

PR #393, opened 2024-02-19
fix missing admin console branding

cbodonnell:
This PR adds `kotsApplication` to the list of protected values for the admin console add-on to prevent it from being overwritten during reconciliation. This fixes an issue where the admin console was missing application branding after the initial cluster install.

Added validation for branding to `check-installation-state.sh`. Example failure: https://github.com/replicatedhq/embedded-cluster/actions/runs/7965716798/job/21745820483?pr=393#step:4:9597

### DIFF
--- a/e2e/scripts/check-installation-state.sh
+++ b/e2e/scripts/check-installation-state.sh
@@ -34,6 +34,9 @@ main() {
     echo "ensure that installation is installed"
     wait_for_installation
     kubectl get installations --no-headers | grep -q "Installed"
+
+    echo "ensure that the admin console branding is available"
+    kubectl get cm -n kotsadm kotsadm-application-metadata
 }
 
 export EMBEDDED_CLUSTER_METRICS_BASEURL="https://staging.replicated.app"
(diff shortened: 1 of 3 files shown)


e2e/scripts/check-installation-state.sh
```diff
@@ -34,6 +34,9 @@ main() {
     echo "ensure that installation is installed"
     wait_for_installation
     kubectl get installations --no-headers | grep -q "Installed"
+
+    echo "ensure that the admin console branding is available"
+    kubectl get cm -n kotsadm kotsadm-application-metadata
```
cbodonnell, line 39: Added to `check-installation-state.sh` so we can ensure we have branding during the initial install flow.

cbodonnell, line 39: Added to `check-installation-state.sh` so we can ensure we have branding during the initial install flow. If we don't want to add stuff here, I can move it to the post upgrade checks, but this seems like a reasonable one to start with.

laverya: Part of the reason I suggested it in post upgrade checks is that we know that there will be a change of the installation object at that point - it might be good to have it in both places

laverya (dismissed):

laverya (approved):

PR #401, opened 2024-02-21
feat: only wait 60s draining a node during reset

laverya:


cmd/embedded-cluster/reset.go
```diff
@@ -91,6 +91,7 @@ func (h *hostInfo) drainNode() error {
 		"drain",
 		"--ignore-daemonsets",
 		"--delete-emptydir-data",
+		"--timeout", "60s",
 		h.Hostname,
```
chris-sanders, line 95: Shouldn't we also pass --delete-local-data

chris-sanders:
Only other comment I have, from the story have you: "All other commands have been reviewed and confirmed to already have a timeout or one is added".

IE. Do you see any other commands that don't have a way to timeout and proceed in the reset command other than this one?

laverya:
I'll take a look

chris-sanders (approved):
As long as tests pass this looks good, thanks!

PR #405, opened 2024-02-22
feat: materialize host support bundle

ricardomaraschini:
we want to have the host suppport bundle definition stored locally in the node in case of the cluster being offline.


e2e/scripts/collect-support-bundle.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+main() {
+    if ! kubectl support-bundle --output host.tar.gz --interactive=false /var/lib/embedded-cluster/support/host-support-bundle.yaml; then
+        echo "Failed to collect local host support bundle"
+        return 1
+    fi
+
+    if ! kubectl support-bundle --output cluster.tar.gz --interactive=false --load-cluster-specs; then
```
laverya, line 10: should we check the size of these files at all?

PR #406, opened 2024-02-22
feat: move join and reset commands to top level

laverya:


cmd/embedded-cluster/reset.go
```diff
@@ -293,10 +293,11 @@ func checkErrPrompt(c *cli.Context, err error) bool {
 }
 
 var resetCommand = &cli.Command{
-	Name: "reset",
+	Name:    "reset",
+	Aliases: []string{"uninstall"},
```
ajp-io, line 297: Having an alias means that people might start using either name, and then it's harder for us to ever remove one or the other.

Let's call this `uninstall`. Chuck and I both felt that this is the more intuitive name for end-users who just ran `install`, and people have already been using `node reset`, so we have a bit of an A/B test if people mention they dislike this new naming.

ajp-io (commented):
Let's use these usage descriptions (can't comment in the PR since these lines didn't change):
```
$ ./gitea-mastodon uninstall -h
NAME:
   gitea-mastodon uninstall - Uninstall gitea-mastodon from the current node

$ ./gitea-mastodon join -h
NAME:
   gitea-mastodon join - Join the current node and install gitea-mastodon
```

cmd/embedded-cluster/reset.go
laverya: ok

ricardomaraschini (approved):

cmd/embedded-cluster/reset.go
ajp-io: Okay, sorry to flip flop. Had a conversation, and while uninstall works well for single-node, I don't want to over-optimize for that. Uninstalling five times in a multi-node cluster does seem weird.

So let's switch back to `reset`.

For usage descriptions, Chris and I came up with these.
```
$ ./gitea-mastodon reset -h
NAME:
   gitea-mastodon reset - Uninstall gitea-mastodon on the current node

$ ./gitea-mastodon join -h
NAME:
   gitea-mastodon join - Join the current node and install gitea-mastodon
```

PR #410, opened 2024-02-22
feat: ensure we always process add-ons the same order

ricardomaraschini:
maps dont guarantee we are going to read the properties in the same order they are created. use an slice of add-ons instead.

### DIFF
--- a/pkg/addons/applier.go
+++ b/pkg/addons/applier.go
@@ -123,23 +123,25 @@ func (a *Applier) HostPreflights() (*v1beta2.HostPreflightSpec, error) {
 }
 
 // load instantiates all enabled addons.
-func (a *Applier) load() (map[string]AddOn, error) {
-	addons := map[string]AddOn{}
+func (a *Applier) load() ([]AddOn, error) {
+	addons := []AddOn{}
 	obs, err := openebs.New()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/addons/applier.go
```diff
@@ -149,16 +151,18 @@ func (a *Applier) Versions(additionalCharts []v1beta1.Chart) (map[string]string,
...
 		return nil, fmt.Errorf("unable to load addons: %w", err)
 	}
+
 	versions := map[string]string{}
-	for name, addon := range addons {
+	for _, addon := range addons {
 		version, err := addon.Version()
 		if err != nil {
-			return nil, fmt.Errorf("unable to get version (%s): %w", name, err)
+			return nil, fmt.Errorf("unable to get version (%T): %w", addon, err)
```
laverya, line 159: Can we do addon.Name() here instead?

laverya (approved):

pkg/addons/applier.go
ricardomaraschini: I was thinking the same.

laverya:
```suggestion
			return nil, fmt.Errorf("unable to get version (%s): %w", addon.Name(), err)
```

laverya (approved):

PR #414, opened 2024-02-23
fix ci with new kotsadm

laverya:


e2e/scripts/check-installation-state.sh
```diff
@@ -25,7 +25,49 @@ wait_for_installation() {
...
+        ready=$(kubectl get pods -n kotsadm -o jsonpath='{.items[*].metadata.name} {.items[*].status.phase}' | grep "nginx" | grep -c Running || true)
+        kubectl get pods -n nginx 2>&1 || true
+        echo "ready: $ready"
+    done
+}
+
+ensure_app_deployed() {
+    local version="$1"
+
+    kubectl kots get versions -n kotsadm embedded-cluster-smoke-test-staging-app
```
sgalsaleh, line 51: you can probably use `kubectl kots get apps` and check that the status is `Ready`, but this works too i guess.

sgalsaleh: actually, nevermind. that doesn't check a specific version.

sgalsaleh (approved):

e2e/scripts/check-installation-state.sh
laverya: I'm checking the version because we upgrade to `sha-upgrade` after this, and I wanted to make sure that we're actually still pinning properly

PR #431, opened 2024-03-05
deploy admin console without minio

cbodonnell:
**NOTE:** Depends on changes in https://github.com/replicatedhq/kots-helm/pull/37

This PR updates the admin console add-on to deploy the KOTS Helm chart without MinIO.


laverya (dismissed):

laverya:
I do still want to see a "install previous version and upgrade" test though

cbodonnell:
> I do still want to see a "install previous version and upgrade" test though

Yup, planning to do that in this PR. It looks like there's some other test fixes needed since we're using a `-build.x` suffix.

laverya:
> > I do still want to see a "install previous version and upgrade" test though
> 
> Yup, planning to do that in this PR. It looks like there's some other test fixes needed since we're using a `-build.x` suffix.

Apparently I added that to the tests since the last time we used a `build` suffix!

.github/workflows/pull-request.yaml
```diff
@@ -73,6 +73,7 @@ jobs:
 
         replicated release create --yaml-dir e2e/kots-release-install --promote CI --version "${SHORT_SHA}"
         replicated release create --yaml-dir e2e/kots-release-upgrade --promote CI --version "${SHORT_SHA}-upgrade"
+        replicated release create --yaml-dir e2e/kots-release-install-old-version --promote CI --version "${SHORT_SHA}-old-version"
```
laverya, line 76: why not create the 'old release' once with a version like 'pre-minio-removal', and then download that version from the endpoint?

.github/workflows/pull-request.yaml
```diff
@@ -1,6 +1,9 @@
 name: Pull request
 on:
 - pull_request
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
```
cbodonnell, line 6: added this since i noticed that runs from old commits were still going. any reason not to do this?

laverya: sure, why not

.github/workflows/pull-request.yaml
```diff
@@ -125,6 +128,7 @@ jobs:
           - TestInstallWithoutEmbed
           - TestResetAndReinstall
           - TestCollectSupportBundle
+          - TestOldVersionUpgrade
```
laverya, line 131: please add this to on-merge too

laverya: (but not the concurrency change)

laverya (approved):

laverya (approved):

PR #434, opened 2024-03-07
Add docs for controlling resources

danj-replicated:
Adds some documentation on custom resources that are key to embedded cluster.


chris-sanders (approved):
Great thanks!

PR #435, opened 2024-03-07
add airgap support to installer

laverya:


pkg/airgap/testfiles/tiny-airgap-noimages.airgap
laverya, line 1: 4.9kb small enough @chris-sanders ?

chris-sanders: Is there really no way to retrieve the binary file at runtime? You can write a lot of lines of code in 4.7kb...
Removing a binary file after commit requires a force push of main, that's why I don't like it. Everyone that clones this repo forever will download a binary file that I suspect, won't be needed in short order.

chris-sanders: Wait, there aren't images here, so this is literally a tar file of yaml files. Why don't you just store the file, and tar them up as part of the test. Then you can add/remove and edit files in the future without the binary file problem.

Just put them in a couple folders and tar them as the first step in the test.

laverya: all right then

laverya: streaming layered tarballs is fun...

laverya: done

cmd/embedded-cluster/install.go
```diff
@@ -179,7 +181,73 @@ func checkLicenseMatches(c *cli.Context) error {
...
 	return nil
+}
+
+func checkAirgapMatches(c *cli.Context) error {
+	rel, err := release.GetChannelRelease()
+	if err != nil {
+		return fmt.Errorf("failed to get release from binary: %w", err) // this should only be if the release is malformed
+	}
+	if rel == nil {
+		return nil // TODO make this an error when done testing
```
laverya, line 192: yes I know about this one

laverya: I'll remove it before release, possibly not before merging this though

pkg/defaults/defaults.go
```diff
@@ -23,6 +23,8 @@ var (
 	K0sBinaryURL = ""
 )
 
+const KOTSADM_NAMESPACE = "kotsadm"
```
ricardomaraschini, line 26: As per convention shouldn't this be named KotsadmNamespace or something like it ?

pkg/release/release.go
```diff
@@ -217,3 +232,41 @@ func (r *ReleaseData) parse() error {
 		}
 	}
 }
+
+// SetReleaseDataForTests should only be called from tests. It sets the release information based on the supplied data.
+func SetReleaseDataForTests(data map[string][]byte) error {
```
ricardomaraschini, line 237: Are we able to place this in a _test file ?

pkg/airgap/config.go
```diff
@@ -0,0 +1,11 @@
+package airgap
+
+import "github.com/k0sproject/k0s/pkg/apis/k0s/v1beta1"
+
+func SetAirgapConfig(cfg *v1beta1.ClusterConfig) {
```
ricardomaraschini, line 5: Can we have this function to live inside config package instead ?

pkg/airgap/materialize.go
```diff
@@ -0,0 +1,112 @@
...
+	"compress/gzip"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/replicatedhq/embedded-cluster/pkg/defaults"
+)
+
+const K0S_IMAGE_PATH = "/var/lib/k0s/images/install.tar.gz"
```
ricardomaraschini, line 14: As per conventions, can we name this K0sImagePath instead ?

pkg/airgap/materialize.go
```diff
@@ -0,0 +1,112 @@
...
+		nextFile, err = tarreader.Next()
+		if err != nil {
+			if err == io.EOF {
+				return fmt.Errorf("embedded-cluster.tar.gz not found in airgap file")
+			}
+			return fmt.Errorf("failed to read airgap file: %w", err)
+		}
+
+		if nextFile.Name == "embedded-cluster/images-amd64.tar" {
+			err = writeOneFile(tarreader, K0S_IMAGE_PATH)
```
ricardomaraschini, line 39: Isn't this going to write a `tar` file with `tar.gz` suffix ?

pkg/addons/registry/registry.go
```diff
@@ -0,0 +1,231 @@
...
+// Outro is executed after the cluster deployment.
+func (o *Registry) Outro(ctx context.Context, cli client.Client) error {
+	if !o.isAirgap {
+		return nil
+	}
+
+	loading := spinner.Start()
+	loading.Infof("Waiting for Registry to be ready")
+
+	err := os.Chmod("/var/lib/embedded-cluster/registry", 0775) // we may wish to run registry as root instead of doing this
```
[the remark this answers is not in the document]
laverya: I think we're going to have to have registry run as root actually

pkg/defaults/defaults.go
laverya: Globals are named like this everywhere in my experience?

pkg/airgap/materialize.go
laverya: Good catch

pkg/defaults/defaults.go
laverya: will do

pkg/release/release.go
laverya: I can't 😢

laverya: (files ending in _test are only compiled for tests in the package in question)

pkg/airgap/config.go
laverya: I'm pretty ambivalent on this - does it fit better as part of airgap or as part of config?

cmd/embedded-cluster/install.go
```diff
@@ -179,7 +181,73 @@ func checkLicenseMatches(c *cli.Context) error {
...
+	}
+
+	// read file from path
+	rawfile, err := os.Open(c.String("airgap-bundle"))
+	if err != nil {
+		return fmt.Errorf("failed to open airgap file: %w", err)
+	}
+	defer rawfile.Close()
+
+	appSlug, channelID, airgapVersion, err := airgap.AirgapBundleVersions(rawfile)
```
sgalsaleh, line 202:
```suggestion
	appSlug, channelID, versionLabel, err := airgap.ChannelReleaseMetadata(rawfile)
```

cmd/embedded-cluster/install.go
```diff
@@ -179,7 +181,73 @@ func checkLicenseMatches(c *cli.Context) error {
...
+	// Check if the airgap bundle matches the application version data
+	if rel.AppSlug != appSlug {
+		// if the app is different, we will not be able to provide the correct vendor supplied charts and k0s overrides
+		return fmt.Errorf("airgap bundle app %s does not match binary app %s, please provide the correct bundle", appSlug, rel.AppSlug)
+	}
+	if rel.ChannelID != channelID {
+		// if the channel is different, we will not be able to install the pinned vendor application version within kots
+		return fmt.Errorf("airgap bundle channel %s does not match binary channel %s, please provide the correct bundle", channelID, rel.ChannelID)
+	}
+	if rel.VersionLabel != airgapVersion {
```
sgalsaleh, line 216: Hmm, should this check channel sequence instead of version label? As version labels aren't always unique. Might be an edge case though.

laverya: We're not compiling in the channel sequence right now is the problem

cmd/embedded-cluster/install.go
laverya: Are you suggesting I rename the function?

cmd/embedded-cluster/install.go
laverya: (we could, but it would still leave the problem of "semver mismatch makes for a far better error message than inscrutable sequence numbers")

pkg/addons/registry/registry.go
```diff
@@ -0,0 +1,229 @@
...
+var registryPassword = helpers.RandString(20)
+var registryAddress = ""
+
+var helmValues = map[string]interface{}{
+	"replicaCount":     1,
+	"fullnameOverride": "registry",
+	"image": map[string]interface{}{
+		"tag": ImageVersion,
+	},
+	"storage": "hostpath", // this is not a recognized option but gets us past all the storage options
```
sgalsaleh, line 43: Where does the data end up in this case?

cmd/embedded-cluster/install.go
laverya: I do like this new name better

pkg/addons/registry/registry.go
sgalsaleh: This would be a `filesystem` then. I think you're getting that value by default.

sgalsaleh: Might be easier to switch to s3/seaweedfs with your current format later though :)

laverya: filesystem created a PVC IIRC

ricardomaraschini (approved):

PR #439, opened 2024-03-08
include k0s images within metadata output

laverya:


pkg/config/config.go
```diff
@@ -34,25 +32,15 @@ func ReadConfigFile(cfgPath string) (dig.Mapping, error) {
...
-	output, err := cmd.Output()
-	if err != nil {
-		return nil, fmt.Errorf("unable to generate default config: %w", err)
-	}
-	var cfg k0sconfig.ClusterConfig
-	if err := k8syaml.Unmarshal(output, &cfg); err != nil {
-		return nil, fmt.Errorf("unable to unmarshal default config: %w", err)
-	}
+func RenderK0sConfig() *k0sconfig.ClusterConfig {
+	cfg := k0sconfig.DefaultClusterConfig()
```
laverya, line 36: I'm a littttle skeptical of actually doing this
I think it'll be fine, but it couples the k0s images to the go.mod instead of the embedded binary

laverya: I've changed our `make` command to pull that version of k0s so that we always have a match

sgalsaleh (approved):

PR #446, opened 2024-03-14
feature: implement artifact puller

ricardomaraschini:
the artifact puller pulls assets from a remote registry and store them locally under `/var/lib/embedded-cluster`. this command works hand in hand with an installation object as the latter provides the source from where the first pulls the artifacts.

the idea is that we will have a job that will run the following command:

```
local-artifact-mirror pull binaries <installation name>
local-artifact-mirror pull images <installation name>
local-artifact-mirror pull helmcharts <installation name>
```

the artifact puller will then read the installation `<installation name>` and fetch the artifacts, storing them in the right place under `/var/lib/embedded-cluster`.

this is an example of an installation object that provides source for such artifacts:

```yaml
apiVersion: embeddedcluster.replicated.com/v1beta1
kind: Installation
metadata:
  creationTimestamp: "2024-03-12T15:30:08Z"
  generation: 8
  name: "20240312153008"
  resourceVersion: "637286"
  uid: 6558c016-c14b-4f59-9a08-691fb5daedfd
spec:
  airGap: true
  artifacts:
    embeddedClusterBinary: 10.110.90.56:5000/embedded-cluster-bundle:v1
    helmCharts: 10.110.90.56:5000/embedded-cluster-charts:v1
    images: 10.110.90.56:5000/embedded-cluster-images:v1
  clusterID: f863807b-957e-4f26-afcc-204d755d2c85
  metricsBaseURL: https://replicated.app
```

and this is a skeleton of the job that will be executed by the operator whenever a new installation object is created:

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: embedded-cluster-updater
  namespace: embedded-cluster
spec:
  backoffLimit: 0
  template:
    spec:
      serviceAccount: embedded-cluster-operator
      serviceAccountName: embedded-cluster-operator
      volumes:
      - name: host
        hostPath:
          path: /var/lib/embedded-cluster
          type: Directory
      restartPolicy: Never
      containers:
      - name: embedded-cluster-updater
        image: busybox:latest
        env:
        - name: INSTALLATION
          value: "20240312153008"
        volumeMounts:
        - mountPath: /var/lib/embedded-cluster
          name: host
          readOnly: false
        command:
        - /bin/sh
        - -e
        - -c
        - |
          /var/lib/embedded-cluster/bin/local-artifact-mirror pull binaries $INSTALLATION
          /var/lib/embedded-cluster/bin/local-artifact-mirror pull images $INSTALLATION
          /var/lib/embedded-cluster/bin/local-artifact-mirror pull helmcharts $INSTALLATION
```

### DIFF
--- a/cmd/local-artifact-mirror/artifact.go
+++ b/cmd/local-artifact-mirror/artifact.go
@@ -2,24 +2,60 @@ package main
 
 import (
 	"context"
-	"crypto/tls"
+	"encoding/json"
 	"fmt"
-	"net/http"
 	"os"
-	"time"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"oras.land/oras-go/v2"
 	"oras.land/oras-go/v2/content/file"
 	"oras.land/oras-go/v2/registry"
 	"oras.land/oras-go/v2/registry/remote"
 	"oras.land/oras-go/v2/registry/remote/auth"
+	"oras.land/oras-go/v2/registry/remote/credentials"
 )
 
-// registryAuth returns the credentials to reach the registry. These credentials are
-// read from the cluster. XXX It is still not defined the secret name from where we
-// are going to read the credentials so this returns empty credentials instead.
-func registryAuth(_ context.Context) (auth.Credential, error) {
-	return auth.Credential{}, nil
+// DockerConfig represents the content of the '.dockerconfigjson' secret.
+type DockerConfig struct {
+	Auths map[string]DockerConfigEntry `json:"auths"`
+}
+
+// DockerConfigEntry represents the content of the '.dockerconfigjson' secret.
+type DockerConfigEntry struct {
+	Username string `json:"username"`
+	Password string `json:"password"`
+	Auth     string `json:"auth"`
+}
+
+// registryAuth returns the authentication store to be used when reaching the
+// registry. The authentication store is read from the cluster secret named
+// 'registry-creds' in the 'kotsadm' namespace.
+func registryAuth(ctx context.Context) (credentials.Store, error) {
+	nsn := types.NamespacedName{Name: "registry-creds", Namespace: "kotsadm"}
+	var sct corev1.Secret
+	if err := kubecli.Get(ctx, nsn, &sct); err != nil {
+		return nil, fmt.Errorf("unable to get secret: %w", err)
+	}
+
+	data, ok := sct.Data[".dockerconfigjson"]
+	if !ok {
+		return nil, fmt.Errorf("unable to find secret .dockerconfigjson")
+	}
+
+	var cfg DockerConfig
+	if err := json.Unmarshal(data, &cfg); err != nil {
+		return nil, fmt.Errorf("unable to unmarshal secret: %w", err)
+	}
+
+	creds := credentials.NewMemoryStore()
+	for addr, entry := range cfg.Auths {
+		creds.Put(ctx, addr, auth.Credential{
+			Username: entry.Username,
+			Password: entry.Password,
+		})
+	}
+	return creds, nil
 }
 
 // pullArtifact fetches an artifact from the registry pointed by 'from'. The artifact
@@ -27,7 +63,7 @@ func registryAuth(_ context.Context) (auth.Credential, error) {
 // Callers are responsible for removing the temporary directory when it is no longer
 // needed. In case of error, the temporary directory is removed here.
 func pullArtifact(ctx context.Context, from string) (string, error) {
-	creds, err := registryAuth(ctx)
+	store, err := registryAuth(ctx)
 	if err != nil {
 		return "", fmt.Errorf("unable to get registry auth: %w", err)
 	}
@@ -53,23 +89,9 @@ func pullArtifact(ctx context.Context, from string) (string, error) {
 	}
 	defer fs.Close()
 
-	// XXX for now we are using a custom transport to skip the certificate validation
-	//     because the test environment is using a self-signed certificate. This should
-	//     be changed.
-	repo.Client = &auth.Client{
-		Credential: auth.StaticCredential(imgref.Registry, creds),
-		Client: &http.Client{
-			Transport: &http.Transport{
-				Proxy:                 http.ProxyFromEnvironment,
-				TLSClientConfig:       &tls.Config{InsecureSkipVerify: true},
-				ForceAttemptHTTP2:     true,
-				MaxIdleConns:          100,
-				IdleConnTimeout:       90 * time.Second,
-				TLSHandshakeTimeout:   10 * time.Second,
-				ExpectContinueTimeout: 1 * time.Second,
-			},
-		},
-	}
+	// setup the pull options. XXX we are using a registry over http.
+	repo.Client = &auth.Client{Credential: store.Get}
+	repo.PlainHTTP = true
 
 	tag := imgref.Reference
 	if _, err := oras.Copy(ctx, repo, tag, fs, tag, oras.DefaultCopyOptions); err != nil {
--- a/cmd/local-artifact-mirror/artifact.go
+++ b/cmd/local-artifact-mirror/artifact.go
@@ -1,0 +1,80 @@
+package main
+
+import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	"net/http"
+	"os"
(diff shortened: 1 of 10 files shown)


cmd/local-artifact-mirror/artifact.go
```diff
@@ -0,0 +1,80 @@
...
+	"oras.land/oras-go/v2/content/file"
+	"oras.land/oras-go/v2/registry"
+	"oras.land/oras-go/v2/registry/remote"
+	"oras.land/oras-go/v2/registry/remote/auth"
+)
+
+// registryAuth returns the credentials to reach the registry. These credentials are
+// read from the cluster. XXX It is still not defined the secret name from where we
+// are going to read the credentials so this returns empty credentials instead.
+func registryAuth(_ context.Context) (auth.Credential, error) {
```
ricardomaraschini, line 21: This is the piece we are missing here. Read the registry credentials from the cluster. We need to give enough permissions for the "embedded-cluster-operator" service account.

cmd/local-artifact-mirror/artifact.go
```diff
@@ -0,0 +1,108 @@
...
+// 'registry-creds' in the 'kotsadm' namespace.
+func registryAuth(ctx context.Context) (credentials.Store, error) {
+	nsn := types.NamespacedName{Name: "registry-creds", Namespace: "kotsadm"}
+	var sct corev1.Secret
+	if err := kubecli.Get(ctx, nsn, &sct); err != nil {
+		if errors.IsNotFound(err) {
+			// if we can't locate a secret then returns an empty
+			// credentials store so we attempt to fetch the assets
+			// without authentication.
+			return credentials.NewMemoryStore(), nil
```
cbodonnell, line 42: should we log something indicating that we will pull anonymously?

cmd/local-artifact-mirror/pull.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+
+		logrus.Infof("embedded cluster binaries materialized")
+		return nil
+	},
+}
+
+// fetchAndValidateInstallation fetches an Installation object from its name and
+// checks if it is valid for an airgap cluster deployment.
+func fetchAndValidateInstallation(ctx context.Context, iname string) (*v1beta1.Installation, error) {
```
cbodonnell, line 198: should we also validate that the necessary artifacts are provided? that shouldn't happen, but we decided to defer the validation to EC in https://github.com/replicatedhq/kots/pull/4524#discussion_r1539950845

cmd/local-artifact-mirror/artifact.go
ricardomaraschini: makes sense. done.

cmd/local-artifact-mirror/pull.go
ricardomaraschini: the fields are [required](https://github.com/replicatedhq/embedded-cluster-operator/blob/main/charts/embedded-cluster-operator/charts/crds/templates/resources.yaml#L238-L242) so they can't be empty. if that is what you mean.

cmd/local-artifact-mirror/pull.go
```diff
@@ -0,0 +1,213 @@
...
+		defer func() {
+			logrus.Infof("removing temporary directory %s", location)
+			os.RemoveAll(location)
+		}()
+
+		dst := defaults.EmbeddedClusterChartsSubDir()
+		src := filepath.Join(location, HelmChartsArtifactName)
+		logrus.Infof("uncompressing %s", src)
+		if err := tgzutils.Uncompress(src, dst); err != nil {
+			return fmt.Errorf("unable to uncompress images: %w", err)
```
laverya, line 133:
```suggestion
			return fmt.Errorf("unable to uncompress helm charts: %w", err)
```

pkg/tgzutils/tgz.go
```diff
@@ -0,0 +1,61 @@
...
+	"archive/tar"
+	"compress/gzip"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+)
+
+// Uncompress decompresses a .tgz file into a directory.
+func Uncompress(tgz, dst string) error {
```
laverya, line 13: why 'uncompress' and not 'decompress'?

'untar' would work too

'uncompress' just sounds wrong to my ear

cmd/local-artifact-mirror/artifact.go
```diff
@@ -0,0 +1,110 @@
...
+		})
+	}
+	return creds, nil
+}
+
+// pullArtifact fetches an artifact from the registry pointed by 'from'. The artifact
+// is stored in a temporary directory and the path to this directory is returned.
+// Callers are responsible for removing the temporary directory when it is no longer
+// needed. In case of error, the temporary directory is removed here.
+func pullArtifact(ctx context.Context, from string) (string, error) {
```
laverya, line 73: why are we writing temporary files here and not returning a reader?

cmd/local-artifact-mirror/serve.go
```diff
@@ -0,0 +1,74 @@
...
+	"syscall"
+	"time"
+
+	"github.com/urfave/cli/v2"
+
+	"github.com/replicatedhq/embedded-cluster/pkg/defaults"
+)
+
+// serveCommand starts a http server that serves files from the /var/lib/embedded-cluster
+// directory. This servers listen only on localhost and is used to serve files needed by
```
laverya, line 19:
```suggestion
// directory. This server listens only on localhost and is used to serve files needed by
```

pkg/goods/goods.go
```diff
@@ -30,19 +30,38 @@ func K0sBinarySHA256() (string, error) {
 //go:embed bins/*
 var binfs embed.FS
 
-// materializeBinaries materializes all binary files from inside bins directory.
+// materializeBinaries materializes all binary files from inside bins directory. If the
+// file already exists a copy of it is made first before overwriting it, this is done
+// because we can't overwrite a running binary. Copies are removed.
```
laverya, line 35: why would a binary be running? I could see this being the case on reset+reinstall, but in that case we should fix our reset code

laverya: in fact, "the file name already exists" should be a panic-worthy error IMO

pkg/tgzutils/tgz.go
laverya: this would be a bit simpler if it took a stream as input 😄

ricardomaraschini: For the use case this is fine. We can always break this into two functions later and make one operate with streams instead. Renamed it to `Decompress`.

cmd/local-artifact-mirror/artifact.go
ricardomaraschini: Each artifact has a different file name so we just return the path to the directory where the it has been stored.

pkg/goods/goods.go
laverya: oh ok, that makes sense

do we restart local-artifact-mirror after rewriting that binary?

ricardomaraschini: No, as far as I know we can't do such a thing from within a container. Do you have any idea how we could achieve that ?

cmd/local-artifact-mirror/artifact.go
laverya: But we're taking everything from that directory, decompressing it elsewhere, and then deleting it, right? Why do we need to write them in the first place?

pkg/goods/goods.go
ricardomaraschini: I think I know how we can do that. Let me craft a different PR for that. I will ping back and we can discuss this there.

cmd/local-artifact-mirror/artifact.go
ricardomaraschini: Ah, I see what you mean. Not sure `oras` supports that. Can we get this in and then I can look into it in a different PR ?

ricardomaraschini: More specifically [this](https://github.com/replicatedhq/embedded-cluster/pull/446/files#diff-26bb60c863d75ab37f05b667b9cf4cc37feec0a1f8b972854197b4d1a8300bddR105) is where this would need to change. Unfortunately I could not find yet a `Target` that implements a stream.

laverya (approved):

PR #447, opened 2024-03-15
add a rudimentary supplemental troubleshoot spec

laverya:


laverya:
... I should put this in the helm chart instead

ricardomaraschini (changes requested):
I think you can add them here: https://github.com/replicatedhq/embedded-cluster-operator/tree/main/charts/embedded-cluster-operator/troubleshoot

laverya:
> I think you can add them here: https://github.com/replicatedhq/embedded-cluster-operator/tree/main/charts/embedded-cluster-operator/troubleshoot

Yep, that's what I should do instead

PR #448, opened 2024-03-18
the initial seaweedfs + registry implementation

laverya:


pkg/addons/seaweedfs/seaweedfs.go
```diff
@@ -0,0 +1,208 @@
...
+		return nil
+	}
+
+	loading := spinner.Start()
+	loading.Infof("Waiting for SeaweedFS to be ready")
+	if err := kubeutils.WaitForNamespace(ctx, cli, namespace); err != nil {
+		loading.Close()
+		return err
+	}
+	loading.Infof("SeaweedFS namespace is ready")
```
sgalsaleh, line 149: How would SeaweedFS become ready without the s3 secret below? I remember it crashes if it cannot find it on startup.

pkg/addons/seaweedfs/seaweedfs.go
```diff
@@ -0,0 +1,208 @@
...
+		},
+		StringData: map[string]string{
+			"seaweedfs_s3_config": accessSecretString,
+		},
+		Type: "Opaque",
+	}
+	err := cli.Create(ctx, &accessSecret)
+	if err != nil {
+		loading.Close()
+		return fmt.Errorf("unable to create seaweedfs-s3-access-secret: %w", err)
```
sgalsaleh, line 170: To match the above suggestion
```suggestion
		return fmt.Errorf("unable to create secret-seaweedfs-s3: %w", err)
```

pkg/addons/seaweedfs/seaweedfs.go
sgalsaleh: Oh this is the namespace, never mind!

pkg/addons/seaweedfs/seaweedfs.go
```diff
@@ -0,0 +1,208 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "seaweedfs-s3-access-secret",
+			Namespace: namespace,
+		},
+		StringData: map[string]string{
+			"seaweedfs_s3_config": accessSecretString,
+		},
+		Type: "Opaque",
+	}
+	err := cli.Create(ctx, &accessSecret)
```
sgalsaleh, line 167: Would this be executed/fail on upgrades since the secret will already exist?

pkg/addons/registry/registry.go
```diff
@@ -181,6 +174,29 @@ func (o *Registry) Outro(ctx context.Context, cli client.Client) error {
...
+			Name:      "seaweedfs-s3-rw",
+			Namespace: namespace,
+		},
+		StringData: map[string]string{
+			"s3AccessKey": rwKey,
+			"s3SecretKey": rwSecret,
+		},
+		Type: "Opaque",
+	}
+	err = cli.Create(ctx, &s3AccessSecret)
```
sgalsaleh, line 194: Ditto: would this be executed/fail on upgrades since the secret will already exist?

pkg/addons/seaweedfs/seaweedfs.go
laverya: No code from this repo runs on upgrades

sgalsaleh: Ah, I see.

pkg/addons/seaweedfs/seaweedfs.go
```diff
@@ -0,0 +1,208 @@
...
+
+	accessSecretTemplate := `{"identities":[{"name":"anvAdmin","credentials":[{"accessKey":"%s","secretKey":"%s"}],"actions":["Admin","Read","Write"]},{"name":"anvReadOnly","credentials":[{"accessKey":"%s","secretKey":"%s"}],"actions":["Read"]}]}`
+	accessSecretString := fmt.Sprintf(accessSecretTemplate, rwAccessKey, rwSecretKey, roAccessKey, roSecretKey)
+	accessSecret := corev1.Secret{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Secret",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "seaweedfs-s3-access-secret",
```
sgalsaleh, line 159:
```suggestion
			Name:      "secret-seaweedfs-s3",
```

pkg/addons/seaweedfs/seaweedfs.go
```diff
@@ -0,0 +1,208 @@
...
+			"size": "1Gi",
+		},
+		"logs": map[string]interface{}{
+			"type": "persistentVolumeClaim",
+			"size": "1Gi",
+		},
+		"s3": map[string]interface{}{
+			"enabled":              true,
+			"enableAuth":           true,
+			"existingConfigSecret": "seaweedfs-s3-access-secret",
```
sgalsaleh, line 70:
```suggestion
			"existingConfigSecret": "secret-seaweedfs-s3",
```

pkg/addons/seaweedfs/seaweedfs.go
```diff
@@ -0,0 +1,208 @@
...
+
+// Overwritten by -ldflags in Makefile
+var (
+	ChartURL  = "https://url"
+	ChartName = "name"
+	Version   = "v0.0.0"
+)
+
+var helmValues = map[string]interface{}{
+	"master": map[string]interface{}{
```
sgalsaleh, line 34: we need to set the `volumeSizeLimitMB` for master to be `30000` (30GB) because it's only `1000` (1GB) by default and it cannot be changed later.

pkg/addons/seaweedfs/seaweedfs.go
```diff
@@ -0,0 +1,208 @@
...
+		"replicas": 1,
+		"data": map[string]interface{}{
+			"type": "persistentVolumeClaim",
+			"size": "1Gi",
+		},
+		"logs": map[string]interface{}{
+			"type": "persistentVolumeClaim",
+			"size": "1Gi",
+		},
+		"disableHttp": true,
```
sgalsaleh, line 44: Based on my comment above:
```suggestion
		"disableHttp": true,
		"volumeSizeLimitMB": 30000,
```

pkg/addons/seaweedfs/seaweedfs.go
```diff
@@ -0,0 +1,208 @@
...
+	},
+	"volume": map[string]interface{}{
+		"replicas": 1,
+		"dataDirs": []map[string]interface{}{
+			{
+				"name":         "data1",
+				"type":         "persistentVolumeClaim",
+				"size":         "10Gi",
+				"storageClass": "openebs-hostpath",
```
sgalsaleh, line 53: Why is storage class only specified for the `volume`? We should either set it for all or remove it for all to use the default.

pkg/addons/seaweedfs/seaweedfs.go
```diff
@@ -0,0 +1,208 @@
...
+				"maxVolumes":   0,
```
sgalsaleh, line 54: Default `maxVolumes` is `7` which would limit how much data can be stored even if the underlying disk has space left. Let's set it to a high number so we don't have to worry about it in the future:
```suggestion
				"maxVolumes":   100,
```
That would roughly allow up to `100 * 30GB = 3TB` disk space to be consumed by the registry bucket. I don't think vendors would ever need more than that, right?

sgalsaleh: No reason not to make it even a bigger number like `1000` 🤷‍♂️. Since we want it to be able to consume whatever storage is available.

pkg/addons/seaweedfs/seaweedfs.go
```diff
@@ -0,0 +1,208 @@
...
+			"type": "persistentVolumeClaim",
+			"size": "1Gi",
+		},
+		"disableHttp": true,
+	},
+	"volume": map[string]interface{}{
+		"replicas": 1,
+		"dataDirs": []map[string]interface{}{
+			{
+				"name":         "data1",
```
sgalsaleh, line 50:
```suggestion
				"name":         "data",
```

pkg/addons/seaweedfs/seaweedfs.go
```diff
@@ -0,0 +1,208 @@
...
+var (
+	ChartURL  = "https://url"
+	ChartName = "name"
+	Version   = "v0.0.0"
+)
+
+var helmValues = map[string]interface{}{
+	"master": map[string]interface{}{
+		"replicas": 1,
+		"data": map[string]interface{}{
```
sgalsaleh, line 36: We should probably use Hostpath volumes for the master because it only has a "soft" state and so that it can be rescheduled to other nodes if the node becomes unhealthy. Even in HA mode, we'll probably only run a single master replica, and a Hostpath volume would allow it to move around.

Or, maybe not even a hostpath? Just use the ephemeral storage? IDK how much space that gives us.

pkg/addons/seaweedfs/seaweedfs.go
```diff
@@ -0,0 +1,208 @@
...
+			"type": "persistentVolumeClaim",
+			"size": "1Gi",
+		},
+		"logs": map[string]interface{}{
```
sgalsaleh, line 40: ditto ^

emosbaugh:
Closing in favor of https://github.com/replicatedhq/embedded-cluster/pull/639

PR #455, opened 2024-03-21
Adds a reboot flag to `reset`

crdant:
TL;DR
-----

Simplifies resetting a node iwth a reboot flag

Details
-------

Encourages better behavior by making it easier to reboot a node after resetting it. Instead of having to run two commands to assure a clean reset:

```
$ sudo ./embedded-cluster reset
$ sudo reboot
```

you just have to

```
$ sudo ./embedded-cluster reset --reboot
```

to do the same.

My implementation shells out to the `reboot` command which takes a little longer than making the minimum required syscalls (`sync(2)` and `reboot(2)`) because of the niceties it brings along like sending a wall message and signaling processes more gracefully.

My first implementation just made the syscalls and it was a lot faster but I decided it probably makes more sense to play nice. I could be convinced it's not needed since the node shouldn't be running any other workloads and the app and cluster are already nicely taken care of.

### DIFF
--- a/cmd/embedded-cluster/uninstall.go
+++ b/cmd/embedded-cluster/uninstall.go
@@ -233,7 +233,6 @@ func stopAndResetK0s() error {
 	if err != nil {
 		return fmt.Errorf("could not reset k0s: %w, %s", err, string(out))
 	}
-	logrus.Infof("Node has been reset. Please reboot to ensure transient configuration is also reset.")
 	return nil
 }
 
@@ -382,7 +381,11 @@ var resetCommand = &cli.Command{
 		err = stopAndResetK0s()
 		if !checkErrPrompt(c, err) {
 			return err
-		}
+		} 
+
+    if !c.Bool("reboot") {
+      logrus.Infof("Node has been reset. Please reboot to ensure transient configuration is also reset.")
+    }
 
 		if _, err := os.Stat(defaults.PathToK0sConfig()); err == nil {
 			if err := os.Remove(defaults.PathToK0sConfig()); err != nil {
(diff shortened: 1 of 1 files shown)


laverya (approved):

cmd/embedded-cluster/uninstall.go
```diff
@@ -404,6 +409,16 @@ var resetCommand = &cli.Command{
...
 
+		if c.Bool("reboot") {
+			// this is a bit more graceful that using sync(2) and reboot(2) but
+			// can take a bit longer
+			if _, err := exec.Command("reboot").Output(); err != nil {
+				return err
+			}
+		} else {
+			return fmt.Errorf("Aborting reboot")
+		}
```
laverya, line 420:
```suggestion
		}
```

laverya: as it is right now, this exits with an error unless you pass in the "--reboot" flag

laverya (changes requested):
Please don't make the command exit nonzero when not passing the reboot flag

crdant:
Oops. Moved the else to the wrong place e when I switched implementations.

ajp-io:
After reset, there's a message that says, "Node has been reset. Please reboot to ensure transient configuration is also reset." Let's make sure we don't show the reboot part if `--reboot` was passed.

cmd/embedded-cluster/uninstall.go
crdant: Definitely wasn't the plan. I thought I had put the `else` on the wrong `if`, which essentially I had, but it was actually that I took out a prompt for "Continue with reboot" that had been in there. Should be fixed up now.

crdant:
> After reset, there's a message that says, "Node has been reset. Please reboot to ensure transient configuration is also reset." Let's make sure we don't show the reboot part if `--reboot` was passed.

I moved the info logging to account for this. Not sure if it was the best approach since it took the logging out of the function that does the reset.

cmd/embedded-cluster/uninstall.go
```diff
@@ -404,6 +412,18 @@ var resetCommand = &cli.Command{
 			}
 		}
 
+		if c.Bool("reboot") {
+			if prompts.New().Confirm("Continue with reboot?", false) {
```
crdant, line 416: I'm not sure we really need this, but it felt like the responsible thing to do.

laverya: if you want to add a confirm here, you'll need to only do so when we don't pass "--no-prompts" IMO

ajp-io: Given that someone passed the reboot flag, I don't think we need a confirmation for this.

laverya (approved):

PR #457, opened 2024-03-25
test airgap in CI

laverya:
This ensures that the airgap flag works with an airgap bundle, but does not yet validate that:
1. the app installs (because it doesn't)
2. that things work without an internet connection (because we're not disabling that yet)

### DIFF
--- a/e2e/scripts/single-node-airgap-install.sh
+++ b/e2e/scripts/single-node-airgap-install.sh
@@ -146,11 +146,11 @@ main() {
         exit 1
     fi
     if ! grep -q "Admin Console is ready!" /tmp/log; then
-        echo "Failed to install embedded-cluster"
+        echo "Failed to validate that the Admin Console is ready"
         exit 1
     fi
     if ! wait_for_healthy_node; then
-        echo "Failed to install embedded-cluster"
+        echo "Failed to wait for healthy node"
         exit 1
     fi
     if ! ensure_node_config; then
@@ -158,7 +158,7 @@ main() {
         exit 1
     fi
     if ! wait_for_pods_running 900; then
-        echo "Failed to install embedded-cluster"
+        echo "Failed to wait for pods to be running"
         exit 1
     fi
     if ! check_openebs_storage_class; then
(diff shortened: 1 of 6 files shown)


e2e/scripts/single-node-airgap-install.sh
```diff
@@ -0,0 +1,191 @@
...
+    if ! embedded-cluster install --no-prompt --license /tmp/license.yaml --airgap-bundle /tmp/release.airgap 2>&1 | tee /tmp/log ; then
+        echo "Failed to install embedded-cluster"
+        exit 1
+    fi
+    if ! grep -q "Admin Console is ready!" /tmp/log; then
+        echo "Failed to install embedded-cluster"
+        exit 1
+    fi
+    if ! wait_for_healthy_node; then
+        echo "Failed to install embedded-cluster"
```
sgalsaleh, line 153:
```suggestion
        echo "Failed to wait for healthy node"
```

e2e/scripts/single-node-airgap-install.sh
```diff
@@ -0,0 +1,191 @@
...
+        exit 1
+    fi
+    if ! ensure_node_config; then
+        echo "Cluster did not respect node config"
+        exit 1
+    fi
+    if ! wait_for_pods_running 900; then
+        echo "Failed to install embedded-cluster"
```
sgalsaleh, line 161:
```suggestion
        echo "Failed to wait for pods to be running"
```

e2e/scripts/single-node-airgap-install.sh
```diff
@@ -0,0 +1,191 @@
...
+    done
+}
+
+main() {
+    if ! embedded-cluster install --no-prompt --license /tmp/license.yaml --airgap-bundle /tmp/release.airgap 2>&1 | tee /tmp/log ; then
+        echo "Failed to install embedded-cluster"
+        exit 1
+    fi
+    if ! grep -q "Admin Console is ready!" /tmp/log; then
+        echo "Failed to install embedded-cluster"
```
sgalsaleh, line 149:
```suggestion
        echo "Failed to validate that the Admin Console is ready"
```

ricardomaraschini (approved):

PR #461, opened 2024-03-28
feat: restart mirror on binary change

ricardomaraschini:
if the binary has changed the mirror should stop so systemd can restart it using the new binary.

### DIFF
--- a/cmd/embedded-cluster/uninstall.go
+++ b/cmd/embedded-cluster/uninstall.go
@@ -402,9 +402,6 @@ var resetCommand = &cli.Command{
 			if err := os.RemoveAll(lamPath); err != nil {
 				return err
 			}
-			if err := os.RemoveAll(defaults.LocalArtifactMirrorPath()); err != nil {
-				return err
-			}
 		}
 
 		if _, err := os.Stat(defaults.EmbeddedClusterHomeDirectory()); err == nil {
--- a/cmd/local-artifact-mirror/serve.go
+++ b/cmd/local-artifact-mirror/serve.go
@@ -37,6 +37,10 @@ var serveCommand = &cli.Command{
 		stop := make(chan os.Signal, 1)
 		signal.Notify(stop, os.Interrupt, syscall.SIGTERM)
 
+		if err := startBinaryWatcher(stop); err != nil {
+			panic(err)
+		}
+
 		server := &http.Server{Addr: "127.0.0.1:50000"}
 		go func() {
 			fmt.Println("Starting server on 127.0.0.1:50000")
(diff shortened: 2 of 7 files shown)


cmd/embedded-cluster/uninstall.go
```diff
@@ -402,9 +402,6 @@ var resetCommand = &cli.Command{
 			if err := os.RemoveAll(lamPath); err != nil {
 				return err
 			}
-			if err := os.RemoveAll(defaults.LocalArtifactMirrorPath()); err != nil {
-				return err
-			}
```
laverya, line 407: shouldn't we still remove this?

cmd/local-artifact-mirror/serve.go
```diff
@@ -60,6 +64,34 @@ var serveCommand = &cli.Command{
...
+func startBinaryWatcher(stop chan os.Signal) error {
+	fpath := defaults.PathToEmbeddedClusterBinary("local-artifact-mirror")
+	stat, err := os.Stat(fpath)
+	if err != nil {
+		return fmt.Errorf("unable to stat %s: %s", fpath, err)
+	}
+	lastmod := stat.ModTime()
+	go func() {
+		fmt.Println("Watching for changes in the binary")
+		ticker := time.NewTicker(5 * time.Second)
```
laverya, line 79: why not https://pkg.go.dev/github.com/fsnotify/fsnotify#section-readme

e2e/local-artifact-mirror_test.go
```diff
@@ -59,5 +60,30 @@ func TestLocalArtifactMirror(t *testing.T) {
...
+
+	command = []string{"journalctl", "-u", "local-artifact-mirror"}
+	if _, _, err := RunCommandOnNode(t, tc, 0, command); err != nil {
+		t.Fatalf("fail to get journalctl logs: %v", err)
+	}
+
+	expected := []string{
+		"Binary changed, sending signal to stop",
+		"Scheduled restart job, restart counter is at",
+	}
```
laverya, line 80: Are there logs that are created upon startup? Could we somehow look for that showing up after the "binary changed" message?

cmd/embedded-cluster/uninstall.go
laverya: oops I see we changed the location, never mind

e2e/local-artifact-mirror_test.go
ricardomaraschini: The `Scheduled restart job, restart counter is at` is logged upon restart.

cmd/local-artifact-mirror/serve.go
ricardomaraschini: Over complicated for the task at hand IMO.

cmd/embedded-cluster/uninstall.go
ricardomaraschini: We were moving the binary to `/usr/local/bin` and using it from there. Now we are using the binary from `/var/lib/embedded-cluster/bin` directly and the whole directory is already removed during resets.

laverya (dismissed):

laverya (approved):

PR #462, opened 2024-03-28
test with real airgap

laverya:


laverya:
TODO: drop the proxy entirely and just disable internet using the same methods

pkg/addons/applier.go
```diff
@@ -101,6 +102,19 @@ func (a *Applier) GetAirgapCharts() ([]v1beta1.Chart, []v1beta1.Repository, erro
 	return regChart, regRepo, nil
 }
 
+func (a *Applier) GetAirgapImages() ([]string, error) {
```
sgalsaleh, line 105: how about naming it the same here as well?
```suggestion
func (a *Applier) GetAdditionalImages() ([]string, error) {
```

sgalsaleh (approved):

sgalsaleh (dismissed):

PR #465, opened 2024-03-28
feat: create version metadata config map

ricardomaraschini:
during install we should create a config map in the embedded cluster operator namespace containing the version metadata (add-on versions, values, images, etc). this is specially useful for airgap environments.


ricardomaraschini:
I still have to adjust a test to check the cm is created correctly so moving this to a Draft.

sgalsaleh (dismissed):

ricardomaraschini:
OK, this is worse than I thought, apparently version is empty in our CI.

laverya (dismissed):

laverya (approved):

PR #468, opened 2024-03-29
use pvc for registry storage

laverya:


laverya:
Until we implement s3 backing, we should use pvc-backed registry instead of hostpath to avoid having the pod rescheduled away from its storage

ricardomaraschini (approved):

PR #487, opened 2024-04-08
allow compiling embedded-cluster with overridden operator image

laverya:
this was already done for admin-console

### DIFF
--- a/pkg/addons/embeddedclusteroperator/embeddedclusteroperator.go
+++ b/pkg/addons/embeddedclusteroperator/embeddedclusteroperator.go
@@ -35,10 +35,11 @@ const (
 
 // Overwritten by -ldflags in Makefile
 var (
-	ChartURL   = "https://url"
-	ChartName  = "name"
-	Version    = "v0.0.0"
-	UtilsImage = "busybox:latest"
+	ChartURL      = "https://url"
+	ChartName     = "name"
+	Version       = "v0.0.0"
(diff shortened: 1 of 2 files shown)


pkg/addons/embeddedclusteroperator/embeddedclusteroperator.go
```diff
@@ -48,6 +49,21 @@ var helmValues = map[string]interface{}{
 	"utilsImage":                UtilsImage,
 }
 
+func init() {
+	if ImageOverride != "" {
+		// split ImageOverride into the image and tag
+		parts := strings.Split(ImageOverride, ":")
```
ricardomaraschini, line 55: What if one uses an image reference by `sha` ? 

Can we not `panic` ? I mean, I guess we can assess this later on and return an error instead.

laverya: this is only for dev use
if you are compiling in an image that won't work with this, I think panicking in the build process in CI is appropriate (this will cause a panic when getting version metadata, which happens in CI)

ricardomaraschini: Ah, I see. This will explode upon calling `embedded-cluster version metadata` AND it is intended only for whoever knows what they are doing. OK. LGTM.

ricardomaraschini (approved):

PR #498, opened 2024-04-09
feat: improve finalizing step feedback

ricardomaraschini:
provide users with more feedback during the finalizing step of the installation. shows the images being pushed.

### DIFF
--- a/pkg/addons/adminconsole/adminconsole.go
+++ b/pkg/addons/adminconsole/adminconsole.go
@@ -226,8 +226,7 @@ func (a *AdminConsole) MaskKotsOutputForOnline() spinner.MaskFn {
 // function replaces some of the messages being printed to the user so the output looks
 // nicer.
 func (a *AdminConsole) MaskKotsOutputForAirgap() spinner.MaskFn {
-
-	current := "Starting the airgap bundle upload process"
+	current := "Uploading air gap bundle"
 	return func(message string) string {
 		switch {
 		case strings.Contains(message, "Pushing application images"):
--- a/pkg/addons/adminconsole/adminconsole.go
+++ b/pkg/addons/adminconsole/adminconsole.go
@@ -7,6 +7,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/k0sproject/dig"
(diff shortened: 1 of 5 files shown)


laverya (approved):

pkg/addons/adminconsole/adminconsole.go
```diff
@@ -244,28 +245,50 @@ func (a *AdminConsole) Outro(ctx context.Context, cli client.Client) error {
...
 	}
 
+	loading.Closef("Admin Console is ready!")
 	if a.licenseFile == "" {
-		loading.Closef("Admin Console is ready!")
 		return nil
 	}
 
-	loading.Infof("Finalizing")
+	lmsg := "Pushing application images"
```
ajp-io, line 253: Here are my ideal messages (excluding all the earlier messages):

**Online**
* Admin console is ready!
* Finalizing
* Finished! (This overwrites "Finalizing")

So after those rewrites, the final terminal output will be:
* Admin console is ready!
* Ready!

**Air Gap**
* Admin console is ready!
* Pushing application images
* Pushing image x (this overwrites "Pushing application images")
* Application images are ready! (this overwrites "Pushing image x")
* Finalizing
* Finished! (This overwrites "Finalizing")

So after those rewrites, the final terminal output will be:
* Admin console is ready!
* Application images are ready!
* Finished!

ajp-io: Couple other changes I've considered that are unrelated to this:

* Change "Waiting for… to be ready" messages to "Deploying…" So "Waiting for storage to be ready" becomes "Deploying storage". This is more concise and shows that we're doing stuff, not just waiting around.
* Lowercase "registry", "storage", "admin console", "embedded cluster operator", etc., when they're in the middle of the line. They're not really proper nouns. (Of course when they start a line like "Admin console is ready!" we would capitalize the first letter.)
  * Deploying storage
  * Storage is ready!
* Change "Node installation finished" to "Node is ready!" so it matches the other lines
* Change "Waiting for Admin Console to deploy: 1/2 ready" to "Deploying admin console (1/2)" so it better matches the other lines
* Change the initial messages like "Materializing files" to "Initializing". Same for air gap (I think it says "Materializing host files" or things like that). This mirrors the "Finalizing" that we do at the end. Make it say "Initialized!" when it's done.
* Add a space before and after the "Visit the admin console" line so that it's really obvious to a user in the output.
Example:
```
✔  Node is ready!
✔  Storage is ready!
✔  Embedded cluster operator is ready!
✔  Admin console is ready!
✔  Finished!

Visit the admin console to configure and install gitea-mastodon: http://35.222.8.61:30000

alex@alex-ec:~$
```

ricardomaraschini: The current code base does not support breaking a single command execution in multiple spinners. With the code in place we can have only one "spinner" while the `kots install` app is running. This will require more work.

ajp-io: Does all of this happen during the `kots install`, so these all have to overwrite each other right now?
```
Pushing application images
Pushing image x
Application images are ready!
Finalizing
Finished!
```

ajp-io: Also I updated my first comment now that you add the x/y progress to the kots CLI

ricardomaraschini: This is the recording of an [online install](https://www.loom.com/share/1d18ce1c5cf340afa812c69a040d0691?sid=a402ca05-df9b-43ea-9e94-d0afab55632c) and this is a recording of an [airgap install](https://www.loom.com/share/a14a9216614b4f188d5c4fd4617dde63?sid=56cbdf01-f71b-4a76-b417-445e0d3df10f). Some notes on the airgap installation:

- There is a delay between calling kots cli and the start of the images uploads, I have chosen to add a message there.
- Artifacts and Images are two separated steps inside kots so I have chosen to keep them separated as well. We could unify them on the same step (images and artifacts on a single spinner) or we could simply omit the artifacts section and let the user wondering, these are the options.

ajp-io: Thank you for these Looms. So helpful for me!

This looks nearly perfect. My only comments:
* I'm fine with the embedded cluster artifacts lines. Thanks for calling that out and presenting options.
* There's a sizable pause between "Finished!" and printing the admin console URL. Why is that?
* "Uploading air gap bundle" would be a good message instead of "Starting air gap bundle upload process", just to keep the wording more similar to other lines.
* Doesn't look like you did the additional rewording I put in my second comment above, but that's okay because it's not directly related to this. If there's time, great, if not, let's track it and do it another time.

ricardomaraschini: 1.  OK.
2. Let's have a different ticket so we can spend time investigating. My assumption is that is due to the time we spend trying to get the valid IP address of the server. For airgap that is just a waste of time.
3. Adressed.
4. Same as 2, as it does not relate we better have a different ticket for it.

laverya (approved):

PR #508, opened 2024-04-11
bug: show success in the previous step instead of current

ricardomaraschini:
when we break the line for a new message we need to flag the previous as success and not the current one. this commit addresses it.


laverya (approved):

laverya:
you know your use case, I suppose

PR #526, opened 2024-04-19
Support airgap updates from the CLI

sgalsaleh:


cmd/embedded-cluster/update.go
```diff
@@ -0,0 +1,58 @@
...
+var updateCommand = &cli.Command{
+	Name:  "update",
+	Usage: fmt.Sprintf("Update %s", binName),
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:   "airgap-bundle",
+			Usage:  "Path to the airgap bundle.",
+			Hidden: true,
+		},
+	},
```
laverya, line 24: IMO make the entire command hidden for now, but make the airgap-bundle flag required

pkg/kotscli/kotscli.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	loading.Closef("Finished!")
+	return nil
+}
+
+// MaskKotsOutputForOnline masks the kots cli output during online installations. For
+// online installations we only want to print "Finalizing" until it is done and then
+// print "Finished!".
+func MaskKotsOutputForOnline() spinner.MaskFn {
```
laverya, line 117: should these commands be internal only? I don't see a reason to export them

laverya (approved):

PR #530, opened 2024-04-22
Feat: Add velero to EC

laverya:


pkg/addons/velero/velero.go
```diff
@@ -0,0 +1,121 @@
...
+	"backupsEnabled":   false,
+	"snapshotsEnabled": false,
+	"deployNodeAgent":  true,
+	"nodeAgent": map[string]interface{}{
+		"podVolumePath": "/var/lib/k0s/kubelet/pods",
+	},
+	"image": map[string]interface{}{
+		"tag": VeleroTag,
+	},
+}
```
cbodonnell, line 39: Do we need to set `uploaderType` to `restic`?
https://github.com/vmware-tanzu/helm-charts/blob/main/charts/velero/values.yaml#L394-L395

Not entirely sure of this myself, but referencing our current install docs: https://docs.replicated.com/enterprise/snapshots-configuring-hostpath#install-velero-and-configure-host-path-storage-in-online-environments

cbodonnell, line 39: Would we need to also install the AWS plugin to support s3 backups? 

Not entirely sure how to do it in the chart, but this look like something? https://github.com/vmware-tanzu/helm-charts/blob/main/charts/velero/values.yaml#L105-L119

pkg/addons/velero/velero.go
```diff
@@ -0,0 +1,121 @@
...
+// Outro is executed after the cluster deployment.
+func (o *Velero) Outro(ctx context.Context, cli client.Client) error {
+	if !o.isEnabled {
+		return nil
+	}
+
+	loading := spinner.Start()
+	loading.Infof("Waiting for Velero to be ready")
+
+	if err := kubeutils.WaitForDeployment(ctx, cli, namespace, "velero"); err != nil {
```
cbodonnell, line 109: Should we wait for node-agent too?

laverya: will do

ricardomaraschini (approved):

PR #540, opened 2024-04-23
chore: remove /usr/local/bin/k0s on reset

ricardomaraschini:
we need to remove the k0s binary when resetting the server.

### DIFF
--- a/cmd/embedded-cluster/uninstall.go
+++ b/cmd/embedded-cluster/uninstall.go
@@ -434,6 +434,12 @@ var resetCommand = &cli.Command{
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


cmd/embedded-cluster/uninstall.go
```diff
@@ -434,6 +434,12 @@ var resetCommand = &cli.Command{
 			}
 		}
 
+		if _, err := os.Stat("/usr/local/bin/k0s"); err == nil {
+			if err := os.RemoveAll("/usr/local/bin/k0s"); err != nil {
```
banjoh, line 438: Nit: Maybe `os.Remove`? I usually avoid `os.RemoveAll` if not a dir

PR #543, opened 2024-04-24
feat: allow overrides on add-on values at install time

ricardomaraschini:
when installing take the user provided overrides for all add-ons.

### DIFF
--- a/pkg/addons/adminconsole/adminconsole.go
+++ b/pkg/addons/adminconsole/adminconsole.go
@@ -110,7 +110,7 @@ func (a *AdminConsole) Version() (map[string]string, error) {
 }
 
 func (a *AdminConsole) Name() string {
-	return releaseName
+	return "AdminConsole"
 }
 
 // GetProtectedFields returns the helm values that are not overwritten when upgrading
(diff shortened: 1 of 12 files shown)


pkg/addons/adminconsole/adminconsole.go
```diff
@@ -110,7 +110,7 @@ func (a *AdminConsole) Version() (map[string]string, error) {
 }
 
 func (a *AdminConsole) Name() string {
-	return "AdminConsole"
+	return releaseName
 }
```
laverya, line 114: I have a feeling this is going to break something, let's see the version test

pkg/addons/velero/velero.go
```diff
@@ -135,7 +135,11 @@ func (o *Velero) Outro(ctx context.Context, cli client.Client) error {
 	return nil
 }
 
+type Options struct {
+	IsEnabled bool
+}
```
laverya, line 140: why the change to pass an options struct? personally I want to have to fix the compile errors for each call to 'new' when I add a new field, as it's almost always going to be required

pkg/addons/applier.go
```diff
@@ -24,15 +22,14 @@ import (
...
-	"github.com/replicatedhq/embedded-cluster/pkg/spinner"
 )
 
 // AddOn is the interface that all addons must implement.
 type AddOn interface {
 	Version() (map[string]string, error)
 	Name() string
 	HostPreflights() (*v1beta2.HostPreflightSpec, error)
-	GenerateHelmConfig(onlyDefaults bool) ([]v1beta1.Chart, []v1beta1.Repository, error)
+	GenerateHelmConfig(onlyDefaults bool) (*v1beta1.Chart, *v1beta1.Repository, error)
```
laverya, line 32: you seem to be doing this in order to allow overriding helm configs by addon name, but the operator code (that's been merged already) overrides helm configs by chart name - I would greatly prefer if we just overrode configs by chart name everywhere

pkg/addons/adminconsole/adminconsole.go
laverya: surprisingly it didn't break unit tests!

ricardomaraschini: restored anyways.

pkg/addons/velero/velero.go
ricardomaraschini: restored the original arguments.

pkg/addons/applier.go
ricardomaraschini: fixed.

laverya (approved):

PR #546, opened 2024-04-24
specify velero credentials secret name

cbodonnell:
This PR adds the `credentials.name` [value](https://github.com/vmware-tanzu/helm-charts/blob/main/charts/velero/values.yaml#L487) to the Velero add-on to specify the secret name used for backup storage location credentials. Without this set, it will fall back to a secret named `velero`, which does not work with KOTS as the credentials secret name there is expected to be `cloud-credentials` ([reference](https://github.com/replicatedhq/kots/blob/v1.108.9/pkg/snapshot/store.go#L51)).

### DIFF
--- a/pkg/addons/velero/velero.go
+++ b/pkg/addons/velero/velero.go
@@ -54,9 +54,6 @@ var helmValues = map[string]interface{}{
 			},
 		},
 	},
-	"podLabels": map[string]interface{}{
-		"component": "velero",
-	},
 	"credentials": map[string]interface{}{
 		"existingSecret": credentialsSecretName,
 	},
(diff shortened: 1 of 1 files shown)


laverya (dismissed):

laverya:
Approved, if it passes

pkg/addons/velero/velero.go
```diff
@@ -51,6 +54,12 @@ var helmValues = map[string]interface{}{
 			},
 		},
 	},
+	"podLabels": map[string]interface{}{
+		"component": "velero",
+	},
```
cbodonnell, line 59: KOTS uses this in a label selector when checking if the Velero/NodeAgent pods have restarted post-configuration.

laverya (dismissed):

laverya (approved):

PR #554, opened 2024-04-29
use openebs-archive chart repo

laverya:


chris-sanders (approved):
Best we can do _right now_

laverya:
see: https://github.com/openebs-archive/charts/pull/350

sgalsaleh (approved):

laverya:
The "TestOldVersionUpgrade" failure is unfortunately expected - all old versions are failing to install because the chart doesn't exist where we expect it to.

Abhinandan-Purkait:
Hi @laverya ,

As per the prior communication regarding OpenEBS re-structuring, (refer this [message](https://kubernetes.slack.com/archives/CUAKPFU78/p1712079300995919)), yesterday we migrated all OpenEBS legacy repositories to a CNCF GitHub [openebs-archive](https://github.com/openebs-archive) organization. In addition to the repositories, the parent Helm charts for the OpenEBS v3.10 release and below, were also moved.

This seems to have caused issues to the existing installations, and users have voiced their concerns.
We have decided to revert the legacy Helm chart repo back to it's earlier location to ensure things continue to run smoothly for a deprecation notice period of SIX months. Post this period (Oct 30), we will re-initiate the Helm repo transfer to the openebs-archive organization.

Hope these SIX months should suffice for users of legacy OpenEBS to plan their migration to the latest OpenEBS v4.x editions.

Please note that after completion of this period and the transfer, the legacy Helm charts will still continue to function only from the openebs-archive organization.

Although it seems you have plans to move to 4.0.0, which is great and the way forward. But till then you can keep using the openebs org, you might need to revert this change if applicable.

Thanks.

laverya:
That's great, thank you! I'm sorry to make you revert though!

PR #559, opened 2024-04-30
update old-version-upgrade test with a k0s spec override to fix the repository issue

laverya:


laverya:
... of course as soon as I write this, OpenEBS moves the repo back (for the next 6 months)
so!

PR #560, opened 2024-04-30
Revert "use openebs-archive chart repo"

laverya:
Reverts replicatedhq/embedded-cluster#554

OpenEBS has moved the chart repo back for the next few months - we'll want to move to the 4.0 chart long before then


diamonwiggins (approved):
lgtm

PR #563, opened 2024-04-30
do not configure proxy packages in CI as we do not use them

laverya:


sgalsaleh (approved):

laverya:
Apparently we do use this right now, to configure the kots CLI in airgap
we should find a better way to do things

PR #571, opened 2024-05-02
Add CLI DR restore command for online mode

sgalsaleh:
This PR:

- Implements the disaster recovery / restore functionality for single-node and multi-node online installs.
- Implements the Playwright setup necessary to start moving testIM tests to Playwright.
- Uses Playwright instead of testIM for the disaster recovery test.
- Moves some instances of the `deploy-kots-application` testIM test to Playwright.

[SC-103324](https://app.shortcut.com/replicated/story/103324)


cmd/embedded-cluster/restore.go
```diff
@@ -0,0 +1,585 @@
...
+	if len(appsVersions) > 1 {
+		return false, "has more than one application"
+	}
+	if _, ok := appsVersions[rel.AppSlug]; !ok {
+		return false, fmt.Sprintf("does not contain the %q application", rel.AppSlug)
+	}
+	if versionLabel := appsVersions[rel.AppSlug]; versionLabel != rel.VersionLabel {
+		return false, fmt.Sprintf("has a different app version (%q) than the current version (%q)", versionLabel, rel.VersionLabel)
+	}
+	return true, ""
```
laverya, line 190: we'll need to add a 'get the correct version with this curl command' message here (and earlier) at some point in the future
not a blocker

cmd/embedded-cluster/restore.go
```diff
@@ -0,0 +1,585 @@
...
+	rel, err := release.GetChannelRelease()
+	if err != nil {
+		return nil, fmt.Errorf("unable to get release from binary: %w", err)
+	}
+	if rel == nil {
+		return nil, fmt.Errorf("no release found in binary")
+	}
+
+	for i := 0; i < 30; i++ {
+		time.Sleep(5 * time.Second)
```
laverya, line 219: we should use expbackoff for this like we do in the util waits, or just an actual k8s wait

pkg/addons/velero/velero.go
```diff
@@ -163,6 +163,9 @@ func (o *Velero) Outro(ctx context.Context, cli client.Client) error {
...
 
 // New creates a new Velero addon.
-func New(isEnabled bool) (*Velero, error) {
-	return &Velero{isEnabled: isEnabled}, nil
+func New(namespace string, isEnabled bool) (*Velero, error) {
+	return &Velero{
+		namespace: namespace,
+		isEnabled: isEnabled,
+	}, nil
 }
```
laverya, line 171: is there a reason to allow configuring the velero namespace?

laverya: I would rather have the entire set of changes in this file reverted honestly

laverya (approved):
only minor things, really - if you want to merge and then fix them I'd be perfectly fine with that

pkg/addons/velero/velero.go
sgalsaleh: Not particularly, but I wanted to follow the pattern we have for the admin console addon.

cmd/embedded-cluster/restore.go
sgalsaleh: Good suggestion.

cmd/embedded-cluster/restore.go
sgalsaleh: Can update in a separate PR. This step usually only takes a few seconds anyways.

pkg/addons/velero/velero.go
laverya: I think allowing us to configure the admin console namespace is also a mistake
so if one should change it's the admin console 😄

PR #574, opened 2024-05-02
set EMBEDDED_CLUSTER_ID for KOTS CLI commands

cbodonnell:
Shortcut: https://app.shortcut.com/replicated/story/104230/ensure-that-necessary-kots-configmaps-secrets-are-restored-as-part-of-embedded-cluster-dr

Depends on KOTS CLI changes in: https://github.com/replicatedhq/kots/pull/4593

This PR sets the `EMBEDDED_CLUSTER_ID` environment variable when executing KOTS CLI commands. This ensures that the KOTS CLI is aware it is running within an embedded cluster.


pkg/helpers/command.go
```diff
@@ -4,14 +4,19 @@ import (
...
 )
 
+type RunCommandWithWriterOptions struct {
+	Env map[string]string
+}
+
 // RunCommandWithWriter runs a the provided command. The stdout of the command is
 // written to the provider writer.
-func RunCommandWithWriter(to io.Writer, bin string, args ...string) error {
+func RunCommandWithWriter(opts RunCommandWithWriterOptions, to io.Writer, bin string, args ...string) error {
```
sgalsaleh, line 19: why not just pass it as another param and support it for both `RunCommandWithWriter` and `RunCommand`?

cbodonnell: ultimately reverted the above and made a more generic `RunCommandWithOptions` in https://github.com/replicatedhq/embedded-cluster/pull/574/commits/ed3b18e817688dbe7236d55f11a0dd4cb7aa8143

sgalsaleh (approved):

e2e/scripts/check-pre-minio-removal-installation-state.sh
cbodonnell, line 1: Created this as a separate script as we need to check for different things in the old version vs. the current version (ex: these DR labels won't exist in the old version)

cbodonnell, line 1: Created this as a separate script as we need to check for different things in the old version vs. the current version (ex: these DR labels won't exist in an old version install), so the checks should just be basic things

sgalsaleh (approved):

PR #577, opened 2024-05-03
Wait for installation status on install/restore

laverya:


sgalsaleh (approved):

laverya:
example updated output:
```
✔  Host files materialized
? Enter an Admin Console password: ****
? Confirm password: ****
✔  Node installation finished
✔  Storage is ready!
✔  Embedded Cluster Operator is ready!
✔  Velero is ready!
✔  Admin Console is ready!
✔  Finished!
✔  Installation is complete!
Visit the admin console to configure and install embedded-cluster-smoke-test-staging-app: http://IP:30000
```

pkg/addons/applier.go
```diff
@@ -65,6 +66,18 @@ func (a *Applier) Outro(ctx context.Context) error {
 			return err
 		}
 	}
+
+	err = spinForInstallation(ctx, kcli)
+	if err != nil {
+		return err
+	}
+
+	if a.licenseFile != "" {
```
sgalsaleh, line 75: maybe add a comment here about why we don't print a success message if there's no license file?

laverya: Sure

ajp-io:
I have a few questions.

* Is this step waiting for vendor-supplied Helm extensions to install? Does it do anything else?
* We chose the finalizing wording because it was at the end, so it's weird to say finalized and then go do more stuff. Maybe we need to adjust that to something else. "Validating license" and "License validated" maybe. I think that step just applies the license in KOTS.
* Can you share a Loom of the output here? When we change that, I'd like to review it. I can see the final output, but I'm interested in what we show as we're installing.

laverya:
Updated output: https://www.terminalizer.com/view/976e113f5939

pkg/addons/applier.go
```diff
@@ -282,6 +298,86 @@ func (a *Applier) waitForKubernetes(ctx context.Context) error {
...
+			select {
+			case <-bgCtx.Done():
+				return
+			case meta, ok := <-ch:
+				if !ok {
+					return // channel closed
+				}
+
+				// figure out what to log
+				if meta.State != embeddedclusterv1beta1.InstallationStatePendingChartCreation {
```
ricardomaraschini, line 322: Can we just `continue/break` on this if statement ? That would keep the happy path left padded. Also, can `meta` be renamed to something more meaningful on this context?

pkg/addons/applier.go
```diff
@@ -282,6 +298,86 @@ func (a *Applier) waitForKubernetes(ctx context.Context) error {
...
+						chartNames = strings.Join(meta.PendingCharts[:len(meta.PendingCharts)-1], ", ") + " and " + meta.PendingCharts[len(meta.PendingCharts)-1]
+					}
+
+					installSpin.Infof("Waiting for additional components %s to be ready", chartNames)
+				}
+			}
+		}
+	}()
+
+	err := kubeutils.WaitForInstallation(ctx, cli, ch)
```
ricardomaraschini, line 345: Why not pass the `installSpin` here instead of the channel ?

e2e/scripts/check-installation-state.sh
```diff
@@ -68,13 +68,16 @@ ensure_app_not_upgraded() {
...
+    sleep 10 # wait for kubectl to become available
 
     echo "pods"
     kubectl get pods -A
 
     echo "ensure that installation is installed"
-    wait_for_installation
+    if echo "$version" | grep "pre-minio-removal"; then
+        echo "waiting for installation as this is pre-minio-removal"
+        wait_for_installation
```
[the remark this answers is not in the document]
laverya: it's for the old-version-upgrade test, and so the installed version won't wait for the installation to be ready, so we need to wait for it here

pkg/addons/applier.go
```diff
@@ -65,6 +68,19 @@ func (a *Applier) Outro(ctx context.Context) error {
 			return err
 		}
 	}
+
+	err = spinForInstallation(ctx, kcli)
+	if err != nil {
+		return err
+	}
+
+	// this is a link to the admin console, and that is only installed if there is a license provided
```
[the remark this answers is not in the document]
laverya: I see we do install it now, but we still didn't print the success message before if you didn't provide a license

pkg/addons/applier.go
laverya: I could do either one

pkg/addons/applier.go
laverya: yeah will do

e2e/scripts/check-installation-state.sh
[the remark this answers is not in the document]
laverya: clarified the echo

pkg/addons/applier.go
laverya: refactored

ricardomaraschini (approved):

laverya:
https://www.terminalizer.com/view/6c3453965941 for the final demo

PR #579, opened 2024-05-06
chore: capture and analyse ipv4 interfaces

ricardomaraschini:
gather and analyses the amount of network interfaces with ipv4. generates an error if no ipv4 interfaces are found.


chris-sanders (approved):
Looks like a good start to me. I'm unclear if "any one interface" is sufficient but it will catch the failure we've seen recently for sure.

PR #590, opened 2024-05-07
remove embedded-cluster from additional components installation output

laverya:


pkg/kubeutils/kubeutils.go
```diff
@@ -148,27 +147,26 @@ func WaitForInstallation(ctx context.Context, cli client.Client, writer *spinner
...
-
-	writer.Infof("Waiting for additional components %s to be ready", chartNames)
+	}
+
+	// remove 'admin-console' from this
+	pendingChartsMap := map[string]struct{}{}
+	for _, chartName := range install.Status.PendingCharts {
+		pendingChartsMap[chartName] = struct{}{}
+	}
+	delete(pendingChartsMap, "admin-console")
```
sgalsaleh, line 164: instead of deleting the admin-console chart explicitly, how about removing charts that aren't part of the extensions?

laverya: hmmm
should be doable

sgalsaleh: yeah, then you wouldn't have to check for negatives below and would also handle charts other than the admin-console

laverya: done

pkg/kubeutils/kubeutils.go
```diff
@@ -148,27 +147,40 @@ func WaitForInstallation(ctx context.Context, cli client.Client, writer *spinner
...
+	numPendingCharts := 0
+	for _, ch := range install.Spec.Config.Extensions.Helm.Charts {
+		if _, ok := pendingChartsMap[ch.Name]; ok {
+			numPendingCharts++
+		}
 	}
 
-	writer.Infof("Waiting for additional components %s to be ready", chartNames)
+	numCompletedCharts := 0
+	if numDesiredCharts-numPendingCharts > 0 {
```
sgalsaleh, line 174: would this ever be true given that the above code filters by extensions now?

pkg/kubeutils/kubeutils.go
```diff
@@ -148,27 +147,40 @@ func WaitForInstallation(ctx context.Context, cli client.Client, writer *spinner
...
+		// it is possible that infra charts are present here in the pending chart count, so we need to make sure this is not negative
+		numCompletedCharts = numDesiredCharts - numPendingCharts
+	}
+
+	if numCompletedCharts < numDesiredCharts {
```
sgalsaleh, line 179: ditto

laverya: yes, this happens when all of the charts are completed but admin console is still applying

sgalsaleh: ah right

sgalsaleh (approved):

PR #597, opened 2024-05-09
feat: implement airgap restore

laverya:
This adds a `--airgap-bundle <path>` flag to the restore command, allowing it to be used in airgap environments. (only with backups taken from airgap installs, which is validated!)

The current test for this installs in airgapped mode, but the instance has internet access to allow usage of s3. This should be changed in a followup PR.


cmd/embedded-cluster/restore.go
```diff
@@ -561,12 +605,24 @@ var restoreCommand = &cli.Command{
...
 
 		logrus.Debugf("restoring infra from backup %q", backup.Name)
-		if err := restoreFromBackup(c.Context, backup, DisasterRecoveryComponentInfra); err != nil {
+		if err := restoreFromBackup(c.Context, backup, DisasterRecoveryComponentInfra, "", nil); err != nil {
 			return err
 		}
 
+		if c.String("airgap-bundle") != "" {
+			logrus.Debugf("restoring embedded cluster registry from backup %q", backup.Name)
+			err := restoreFromBackup(c.Context, backup, DisasterRecoveryComponentChart, "registry", &metav1.LabelSelector{
```
sgalsaleh, line 614: how about just adding a new `DisasterRecoveryComponentChartRegistry` type and keep the logic of determining the label selector in the `restoreFromBackup` function? that way, you also wouldn't have to pass the two extra parameters.

e2e/install_test.go
```diff
@@ -876,6 +876,111 @@ func TestSingleNodeDisasterRecovery(t *testing.T) {
...
+	for _, envVar := range requiredEnvVars {
+		if os.Getenv(envVar) == "" {
+			t.Fatalf("missing required environment variable: %s", envVar)
+		}
+	}
+
+	testArgs := []string{}
+	for _, envVar := range requiredEnvVars {
+		if envVar == "DR_AWS_S3_PREFIX" {
+			testArgs = append(testArgs, os.Getenv(envVar)+"-airgap")
```
sgalsaleh, line 899: instead of mutating the value here, how about just passing a `DR_AWS_S3_AIRGAP_PREFIX` env var from the github action?

e2e/install_test.go
```diff
@@ -928,15 +1033,15 @@ func TestInstallSnapshotFromReplicatedApp(t *testing.T) {
 	t.Logf("%s: test complete", time.Now().Format(time.RFC3339))
 }
 
-func downloadAirgapBundle(t *testing.T, versionLabel string, destPath string) string {
+func downloadAirgapBundle(t *testing.T, versionLabel string, destPath string, license string) string {
```
sgalsaleh, line 1036:
```suggestion
func downloadAirgapBundle(t *testing.T, versionLabel string, destPath string, licenseID string) string {
```

e2e/scripts/restore-installation-airgap.exp
```diff
@@ -0,0 +1,198 @@
...
+expect {
+    -timeout 60 "Infrastructure restored!" {}
+    timeout {
+      puts "\n\nFailed to restore infrastructure."
+      exit 1
+    }
+}
+
+expect {
+    -timeout 600 "Registry restored!" {}
```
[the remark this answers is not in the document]
laverya: no I should change that back down to something more sane

laverya:
this PR was included in https://github.com/replicatedhq/embedded-cluster/pull/614 and is now outdated

PR #613, opened 2024-05-14
Pre-create kots password secret

danj-replicated:
Creates kots password secret before kots starts without storing the password in chart values

depends on:
- https://github.com/replicatedhq/kots-helm/pull/45
- kots PR to leverage ^

### DIFF
--- a/pkg/addons/adminconsole/adminconsole.go
+++ b/pkg/addons/adminconsole/adminconsole.go
@@ -43,6 +43,7 @@ var (
 	ImageOverride           = ""
 	MigrationsImageOverride = ""
 	CounterRegex            = regexp.MustCompile(`(\d+)/(\d+)`)
+	Password                = ""
 )
 
 // protectedFields are helm values that are not overwritten when upgrading the addon.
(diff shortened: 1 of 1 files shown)


pkg/addons/adminconsole/adminconsole.go
```diff
@@ -218,6 +209,8 @@ func (a *AdminConsole) Outro(ctx context.Context, cli client.Client) error {
 	loading.Infof("Waiting for Admin Console to deploy")
 	defer loading.Close()
 
+	createKotsPasswordSecret(ctx, cli, a.namespace, Password)
```
ricardomaraschini, line 212: We should check the returned error here.

pkg/addons/adminconsole/adminconsole.go
```diff
@@ -330,10 +323,47 @@ func createRegistrySecret(ctx context.Context, cli client.Client, namespace stri
...
 }
+
+func createKotsPasswordSecret(ctx context.Context, cli client.Client, namespace string, password string) error {
+	if err := kubeutils.WaitForNamespace(ctx, cli, namespace); err != nil {
+		return err
+	}
+
+	passwordBcrypt, err := bcrypt.GenerateFromPassword([]byte(password), 10)
+	if err != nil {
+		return fmt.Errorf("unable to create kotsadm-password secret: %w", err)
```
ricardomaraschini, line 342: I think this error message could be more specific. The "unable to create secret" would wrap this.

ricardomaraschini (approved):

PR #614, opened 2024-05-14
feat: implement airgap restore

cbodonnell:
Based off existing PR: https://github.com/replicatedhq/embedded-cluster/pull/597

This adds a --airgap-bundle <path> flag to the restore command, allowing it to be used in airgap environments. (only with backups taken from airgap installs, which is validated!)

The current test for this installs in airgapped mode, but the instance has internet access to allow usage of s3. This should be changed in a followup PR.


pkg/addons/registry/registry.go
```diff
@@ -120,6 +121,19 @@ func (o *Registry) GenerateHelmConfig(onlyDefaults bool) ([]v1beta1.Chart, []v1b
...
 
+	// use a static cluster IP for the registry service based on the cluster CIDR range
+	serviceCIDR := v1beta1.DefaultNetwork().ServiceCIDR
+	if o.config.Spec != nil && o.config.Spec.Network != nil {
+		serviceCIDR = o.config.Spec.Network.ServiceCIDR
+	}
+	registryServiceIP, err := helpers.GetLowerBandIP(serviceCIDR, 10)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get cluster IP for registry service: %w", err)
+	}
```
cbodonnell, line 132: Right now this would pull the IP at index 10, so for the k0s default service CIDR of `10.96.0.0/12` would be `10.96.0.11` and follow a similar patter for different ranges. Is this sufficient to specify a static "index" like this? Do we need to make it dynamic at this time? If so, how could we go about that?

Makefile
```diff
@@ -5,7 +5,7 @@ APP_NAME = embedded-cluster
 ADMIN_CONSOLE_CHART_URL = oci://registry.replicated.com/library
 ADMIN_CONSOLE_CHART_NAME = admin-console
 ADMIN_CONSOLE_CHART_VERSION = 1.109.3
-ADMIN_CONSOLE_IMAGE_OVERRIDE =
+ADMIN_CONSOLE_IMAGE_OVERRIDE = ttl.sh/cbodonnell/kotsadm:24h
```
laverya, line 8: is this still required?

cbodonnell: Good call out. Definitely not needed and was in here because the tests depend on the KOTS changes and KOTS CI was blocked yesterday. Should be unblocked now, so I'll get that released and the new version incorporated into this PR before merging.

e2e/scripts/check-airgap-installation-state.sh
```diff
@@ -2,20 +2,19 @@
 set -euox pipefail
 
 wait_for_nginx_pods() {
-    ready=$(kubectl get pods -n kotsadm -o jsonpath='{.items[*].metadata.name} {.items[*].status.phase}' | grep "nginx" | grep -c Running || true)
+    ready=$(kubectl get pods -n kotsadm | grep "nginx" | grep -c Running || true)
```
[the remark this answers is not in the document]
laverya: oh that'd do it

pkg/addons/registry/registry.go
laverya: this seems good to me
if we have a collision in the future, we can deal with it then

laverya (approved):

cmd/embedded-cluster/restore.go
```diff
@@ -437,6 +459,39 @@ func pickBackupToRestore(backups []velerov1.Backup) *velerov1.Backup {
 	return latestBackup
 }
 
+// waitForRestoreCompleted waits for a Velero restore to complete.
+func waitForRestoreCompleted(ctx context.Context, restoreName string) (*velerov1.Restore, error) {
```
sgalsaleh, line 463: I was trying to avoid this function name because it's not clear which "restore" we're waiting for. The EC restore? Velero restore? Any reason it couldn't remain as part of the `waitForDRComponent` function?

cmd/embedded-cluster/restore.go
```diff
@@ -524,6 +578,21 @@ func restoreFromBackup(ctx context.Context, backup *velerov1.Backup, drComponent
 
 	// create a new restore object if it doesn't exist
 	if errors.IsNotFound(err) {
+		var restoreLabelSelector *metav1.LabelSelector
+		if drComponent != disasterRecoveryComponentRegistry {
```
sgalsaleh, line 582: Nit pick: it's probably clearer to handle the positive condition here and flip the assignment. Specially when more components will be added in the future. e.g.:
```suggestion
		if drComponent == disasterRecoveryComponentRegistry {
```

cmd/embedded-cluster/restore.go
```diff
@@ -598,7 +663,13 @@ func waitForAdditionalNodes(ctx context.Context) error {
 var restoreCommand = &cli.Command{
 	Name:  "restore",
 	Usage: fmt.Sprintf("Restore a %s cluster", binName),
-	Flags: []cli.Flag{},
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:   "airgap-bundle",
+			Usage:  "Path to the airgap bundle. If set, the installation will be completed without internet access.",
```
sgalsaleh, line 669:
```suggestion
			Usage:  "Path to the airgap bundle. If set, the restore will be completed without internet access.",
```

cmd/embedded-cluster/restore.go
```diff
@@ -733,10 +810,29 @@ var restoreCommand = &cli.Command{
...
 			}
+
 			logrus.Debugf("restoring infra from backup %q", backupToRestore.Name)
 			if err := restoreFromBackup(c.Context, backupToRestore, disasterRecoveryComponentInfra); err != nil {
 				return err
 			}
+
+			if c.String("airgap-bundle") != "" {
+				logrus.Debugf("restoring embedded cluster registry from backup %q", backupToRestore.Name)
+				err := restoreFromBackup(c.Context, backupToRestore, disasterRecoveryComponentRegistry)
```
sgalsaleh, line 821: Let's make this its own restore state? That way it will also match the logging.

cmd/embedded-cluster/restore.go
cbodonnell: The comment reads `waitForRestoreCompleted waits for a Velero restore to complete`. I could potentially just rename it to `waitForVeleroRestoreCompleted`? I didn't mind it as a a separate function to make `waitForDRComponent` a bit smaller and more readable.

sgalsaleh (dismissed):

sgalsaleh (approved):

PR #620, opened 2024-05-17
update to latest versions of everything besides k0s

laverya:


laverya:
We can't update to k0s 1.30 right now due to the chart type change, but should be able to in the future

ricardomaraschini (approved):

PR #621, opened 2024-05-17
feat: wait for coredns and metrics-server

laverya:


pkg/addons/applier.go
```diff
@@ -319,37 +322,29 @@ func (a *Applier) Versions(additionalCharts []v1beta1.Chart) (map[string]string,
 	return versions, nil
 }
 
-// waitForKubernetes waits until we manage to make a successful connection to the
-// Kubernetes API server.
-func (a *Applier) waitForKubernetes(ctx context.Context) error {
+// waitForKubernetes waits for codedns and metrics-server to be ready in kube-system.
+func (a *Applier) waitForKubernetes(ctx context.Context, cli client.Client) error {
```
laverya, line 326: this function was previously unused, and the only bit that was retained is the name

pkg/addons/applier.go
```diff
@@ -319,37 +322,29 @@ func (a *Applier) Versions(additionalCharts []v1beta1.Chart) (map[string]string,
...
+		loading.Close()
+		return fmt.Errorf("unable to wait for CoreDNS: %w", err)
+	}
+
+	loading.Infof("Waiting for Kubernetes System Infrastructure to be ready 1/2")
+
+	err = kubeutils.WaitForDeployment(ctx, cli, "kube-system", "metrics-server")
+	if err != nil {
+		loading.Errorf("Metrics Server failed to become healthy - this may be a firewall or network issue.")
+		loading.Close()
```
ricardomaraschini, line 342: dito

sgalsaleh (approved):

PR #624, opened 2024-05-17
feat: wait for coredns and metrics-server, mk2

laverya:
Reverts replicatedhq/embedded-cluster#623, but runs in the background so that display is unchanged except in case of errors

### DIFF
--- a/pkg/addons/applier.go
+++ b/pkg/addons/applier.go
@@ -60,14 +60,17 @@ func (a *Applier) Outro(ctx context.Context) error {
 		return fmt.Errorf("unable to load addons: %w", err)
 	}
 
-	err = a.waitForKubernetes(ctx, kcli)
-	if err != nil {
-		return fmt.Errorf("unable to wait for kubernetes: %w", err)
-	}
+	errCh := a.waitForKubernetes(ctx, kcli)
 
 	for _, addon := range addons {
 		if err := addon.Outro(ctx, kcli); err != nil {
+			if len(errCh) != 0 {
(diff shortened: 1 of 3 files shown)


laverya:
Limitations of this: 

1. It only displays in between the other add-ons - it won't interrupt a spinner
2. It'll only show one background error, even if there are multiple

pkg/addons/applier.go
```diff
@@ -62,10 +59,19 @@ func (a *Applier) Outro(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("unable to load addons: %w", err)
 	}
+
+	errCh := a.waitForKubernetes(ctx, kcli)
```
sgalsaleh, line 63: how about running this in a defer statement?

laverya: The 'read from errCh and print that message' bit?
yeah that makes sense to me

laverya:
A demonstration of output with severely broken networking:
```
✔  Host files materialized!
? Enter an Admin Console password: ******
? Confirm password: ******
✔  Node installation finished!
✔  Waiting for Storage to be ready
the Kubernetes Infrastructure failed to become ready: Metrics Server failed to become healthy: timed out waiting for metrics-server to deploy
timed out waiting for openebs-localpv-provisioner to deploy: unable to get deploy openebs-localpv-provisioner status: deployments.apps "openebs-localpv-provisioner" not found
```

sgalsaleh (approved):

PR #632, opened 2024-05-21
airgap disaster recovery CI test

cbodonnell:
This PR modifies the airgap disaster recovery CI test so that it actually runs in an airgapped environment. This changes the test so that it leverages the lxd proxy node for access to the AWS S3 bucket used for the backup storage location. Part of this also includes the addition of a hidden `--proxy` flag that reads proxy environment variables from the system and configures Velero and KOTS accordingly.


e2e/playwright/tests/deploy-airgap-upgrade/test.spec.ts
```diff
@@ -8,7 +8,7 @@ test('deploy airgap upgrade', async ({ page }) => {
   await page.getByRole('button', { name: 'Deploy', exact: true }).click();
   await expect(page.locator('.Modal-body')).toBeVisible();
   await page.getByRole('button', { name: 'Yes, Deploy' }).click();
-  await expect(page.locator('#app')).toContainText('Updating cluster', { timeout: 60000 });
+  await expect(page.locator('#app')).toContainText('Updating cluster', { timeout: 90000 });
```
cbodonnell, line 11: It seems that the deploy time takes more than 60 seconds for the airgap app. This is strange since it's just a basic nginx application. Looking at the logs we can see that ~60 seconds of that deploy time is from KOTS calling troubleshoot functions to try and fetch discovered specs via their upstream `uri` field:
```
{"level":"info","ts":"2024-05-23T13:02:29Z","msg":"deploying app version[{appId 15 0 2grwY2Uoh4yU8NyXQEgLzlsFygT <nil>} {sequence 11 1  <nil>}]"}
...
I0523 13:02:32.618472       1 load.go:82] using upstream reference: https://raw.githubusercontent.com/replicatedhq/embedded-cluster-operator/main/charts/embedded-cluster-operator/troubleshoot/cluster-support-bundle.yaml
...
E0523 13:03:02.619497       1 load.go:85] failed to load upstream supportbundle, falling back
I0523 13:03:02.619996       1 load.go:82] using upstream reference: https://raw.githubusercontent.com/replicatedhq/kots-helm/main/troubleshoot/embedded-cluster-support-bundle.yaml
E0523 13:03:32.620344       1 load.go:85] failed to load upstream supportbundle, falling back
...
{"level":"info","ts":"2024-05-23T13:03:37Z","msg":"started cluster upgrade[]"}
```

This increased timeout will account for this, but a more permanent solution likely needs to be implemented in KOTS, Troubleshoot, or elsewhere.

pkg/addons/adminconsole/adminconsole.go
```diff
@@ -183,6 +184,16 @@ func (a *AdminConsole) GenerateHelmConfig(onlyDefaults bool) ([]v1beta1.Chart, [
...
 		}
+		if len(a.proxyEnv) > 0 {
+			extraEnv := []map[string]interface{}{}
+			for k, v := range a.proxyEnv {
+				extraEnv = append(extraEnv, map[string]interface{}{
+					"name":  k,
+					"value": v,
+				})
+			}
+			helmValues["extraEnv"] = extraEnv
```
laverya, line 195: don't kill me, but this will be wiped on upgrade/helm reapply

you'll likely need to add the proxy values to the installation object and then insert them properly in operator

laverya (approved):

pkg/addons/adminconsole/adminconsole.go
laverya: this isn't a blocker here, since we only really need it on initial install/restore, but if when we eventually want to make proxy a supported flag we'll need to handle it

e2e/playwright/tests/deploy-airgap-upgrade/test.spec.ts
sgalsaleh: It's great that we can now discuss these types of test changes in PRs.

sgalsaleh (approved):

PR #636, opened 2024-05-23
feat: materialize containerd config + allow shell on worker

ricardomaraschini:
- materialize the containerd configuration through the materialize command (when the svc registry exists on the registry namespace).
- allow shell on worker nodes: this may be removed but as i was debugging this it was really handy so i left it.

the recording of the shell on a worker node can be found [here](https://www.loom.com/share/c7dbc0d8b8124c28b79c524c71d21798).


ricardomaraschini:
This isn't going in as we cut the scope of the task. Closing.

PR #638, opened 2024-05-24
feat: Store host preflight results on disk

banjoh:
On install, restore or node join embedded cluster runs host preflights. In order to inspect the result of the last host preflight run after the fact, we need to store this information on disk. This PR introduces this change where the results are stored in /var/lib/embedded-cluster/host-preflights/results.json Only the most resent results will be present.


chris-sanders:
@banjoh can you talk to Dan about how they are collecting logs from the hosts and make sure this is either included or can be added so that we can attempt to collect these files with bundles as well?

banjoh:
> @banjoh can you talk to Dan about how they are collecting logs from the hosts and make sure this is either included or can be added so that we can attempt to collect these files with bundles as well?

I believe you are referring to using `runDaemonSet` collector to collect files from disk

https://github.com/replicatedhq/embedded-cluster-operator/pull/182

pkg/defaults/provider.go
```diff
@@ -202,3 +202,12 @@ func (d *Provider) EmbeddedClusterSupportSubDir() string {
 func (d *Provider) PathToEmbeddedClusterSupportFile(name string) string {
 	return filepath.Join(d.EmbeddedClusterSupportSubDir(), name)
 }
+
+func (d *Provider) PathToHostPreflightsDir() string {
+	path := filepath.Join(d.EmbeddedClusterHomeDirectory(), "host-preflights")
```
ricardomaraschini, line 207: Wouldn't it be cleaner to store this under the `support` directory ?

pkg/preflights/output.go
```diff
@@ -46,6 +49,23 @@ func (o Output) PrintTable() {
 	logrus.Infof("%s\n", tb.Render())
 }
 
+func (o Output) SaveToDisk() error {
+	// Store results on disk of the host that ran the preflights
+	data, err := json.MarshalIndent(o, "", "  ")
+	if err != nil {
+		return fmt.Errorf("failed to marshal preflight results: %w", err)
```
ricardomaraschini, line 56: We have been trying to keep these error messages standardised with "unable to", I might be the only one still doing so as a lot of things are happening so take this with a grain of salt.

e2e/scripts/embedded-preflight.sh
```diff
@@ -177,6 +188,11 @@ main() {
         cat /tmp/log
         exit 1
     fi
+    if ! has_stored_host_preflight_results; then
```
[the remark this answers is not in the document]
banjoh: Good point

pkg/defaults/provider.go
banjoh: Something like `/var/lib/embedded-cluster/support/host-preflights/results.json`?

pkg/preflights/output.go
banjoh: Yer. A number of these are `copilot` generated error messages.

pkg/defaults/provider.go
ricardomaraschini: Oh, we may want to collect this file here as well (in the case the cluster isn't working): https://github.com/replicatedhq/embedded-cluster/blob/main/pkg/goods/support/host-support-bundle.yaml

pkg/preflights/host-preflight.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: troubleshoot.sh/v1beta2
```
banjoh, line 1: This is a subset of the host support bundle specs. Not all apply to host preflights.

ricardomaraschini: Not sure, should we use the `uri` parameter here ?

ricardomaraschini (dismissed):

pkg/preflights/host-preflight.yaml
chris-sanders: I would think so, I'm not aware of any reason to intentionally not use the uri

banjoh:
> Not sure, should we use the uri parameter here ?

Not at this time since `HostPreflight` kind is not updatable. I created a bug to fix this.

https://github.com/replicatedhq/troubleshoot/issues/1557

pkg/preflights/host-preflight.yaml
banjoh: There is one use case to consider here though. If a new analyser/collector is built and added to troubleshoot, then newer versions of EC adopt it in their default host preflight specs, there is a risk of older install/restore/upgrade/join preflights failing. This is because `preflight` command reports failures if an analyser is unknown.

chris-sanders: I would be inclined to call that a bug and fix it. URI should work for this use case and this seems like a blocker to using it.

cmd/embedded-cluster/install.go
```diff
@@ -473,21 +509,21 @@ var installCommand = &cli.Command{
...
-			metrics.ReportApplyFinished(c, err)
-			return err
-		}
 		logrus.Debugf("creating k0s configuration file")
 		if err := ensureK0sConfig(c); err != nil {
 			err := fmt.Errorf("unable to create config file: %w", err)
 			metrics.ReportApplyFinished(c, err)
 			return err
 		}
+		logrus.Debugf("running host preflights")
```
ricardomaraschini, line 518: Isn't this, assuming preflights fail, force a `./embedded-cluster reset` before attempting to install again ?

internal/flagutils/cli-flags.go
```diff
@@ -0,0 +1,15 @@
+package flagutils
```
ricardomaraschini, line 1: I would rather have this as part of the commands themselves than have a entire package just for this.

ricardomaraschini (changes requested):
A couple of small things and one important thing IMHO: users should not need to reset the node before installing if a previous attempt failed due to host preflights. I will defer to @ajp-io on that though.

cmd/embedded-cluster/install.go
ajp-io: If rerunning `install` after host preflights fail is totally fine, then let's not tell them to reset. We should only force a reset if the install already did something that could cause an issue down the line.

e2e/scripts/embedded-preflight.sh
```diff
@@ -161,9 +167,17 @@ main() {
...
+        echo "Install hasn't stored host preflight results to disk"
+        cat /tmp/log
+        exit 1
+    fi
+    rm /var/lib/embedded-cluster/support/host-preflight-results.json
     mv /tmp/log /tmp/log-failure
     embed_preflight "$preflight_with_warning"
-    if ! embedded-cluster install --no-prompt 2>&1 | tee /tmp/log ; then
+    cat /root/preflight.yaml
+    /usr/local/bin/embedded-cluster reset --force
```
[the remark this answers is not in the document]
banjoh: Before running host preflights, we drop k0s configs on disk. This change was introduced in this PR to allow setting the Admin Console password before running preflights. Running `reset` clears these k0s configs dropped by the previous install attempt.

internal/flagutils/cli-flags.go
banjoh: I was trying to remove duplicate code, but I can remove the package. I'm OK with either approach

banjoh: > I was trying to remove duplicate code, but I can remove the package. I'm OK with either approach

The code below will be in 3 different places
```go
&cli.BoolFlag{
		Name:  "skip-host-preflights",
		Usage: "Skip host preflight checks. This is not recommended unless you are sure your system is compatible.",
		Value: false,
	}
```

laverya (approved):

PR #648, opened 2024-05-28
do not update go.mod for PREVIOUS_K0S_VERSION tests

laverya:
This is not perfectly ideal, as it means that we'll possibly be using infrastructure images for the wrong k0s version when upgrading from PREVIOUS_K0S_VERSION in our testing - this means that we will not be changing k0s infra images when doing an airgap upgrade, which is a slight loss of test coverage. We will still be changing chart images, so there's still coverage of that aspect.

On the other hand, this is only in our testing, and not in the release version - and so we won't be shipping k0s using the wrong images.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -50,6 +50,9 @@ jobs:
       uses: actions/setup-go@v5
       with:
         go-version: "1.21.0"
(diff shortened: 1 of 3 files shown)


cbodonnell (approved):

laverya:
I like https://github.com/replicatedhq/embedded-cluster/pull/649 better

PR #657, opened 2024-05-30
prompt for high availability during node join

cbodonnell:
https://app.shortcut.com/replicated/story/105699/add-support-for-ha-registry

This PR adds a hidden CLI flag `--enable-ha` that can be used to enable the HA option when joining additional controller nodes. If passed, this flag will check if the joining node would be the third or more controller node and prompts the user to enable HA if it's not already enabled.

Loom showing CLI: https://www.loom.com/share/a7ff5303331b4fd195b5a7d14d7c969f


cmd/embedded-cluster/join.go
```diff
@@ -368,3 +393,100 @@ func runK0sInstallCommand(fullcmd string) error {
...
+	installation, err := kubeutils.GetLatestInstallation(ctx, kcli)
+	if err != nil {
+		return fmt.Errorf("unable to get latest installation: %w", err)
+	}
+	installation.Spec.HighAvailability = true
+	if err := kcli.Update(ctx, installation); err != nil {
+		return fmt.Errorf("unable to update installation: %w", err)
+	}
+
+	// TODO: how should we wait for HA migration to complete?
```
cbodonnell, line 488: Maybe look for the condition on the installation object? Looking at https://github.com/replicatedhq/embedded-cluster-operator/pull/196

cmd/embedded-cluster/join.go
```diff
@@ -368,3 +393,100 @@ func runK0sInstallCommand(fullcmd string) error {
...
+		return fmt.Errorf("unable to get hostname: %w", err)
+	}
+	numNodes, err := waitForNode(ctx, kcli, hostname)
+	if err != nil {
+		return fmt.Errorf("unable to wait for node: %w", err)
+	}
+	if numNodes < 3 {
+		return nil
+	}
+	shouldEnableHA := prompts.New().Confirm("Do you want to enable high availability?", false)
```
sgalsaleh, line 423: We should probably add a separate CLI command to be able to enable HA without having to join additional nodes in case someone chooses "No" here by mistake.

cmd/embedded-cluster/join.go
sgalsaleh: Yes

cmd/embedded-cluster/join.go
```diff
@@ -368,3 +393,100 @@ func runK0sInstallCommand(fullcmd string) error {
...
+		return fmt.Errorf("unable to check if HA is enabled: %w", err)
+	}
+	if isHA {
+		return nil
+	}
+	hostname, err := os.Hostname()
+	if err != nil {
+		return fmt.Errorf("unable to get hostname: %w", err)
+	}
+	numNodes, err := waitForNode(ctx, kcli, hostname)
```
sgalsaleh, line 416: This should probably be outside of this function with a loading message like: "Waiting for node to join the cluster" or something.

cmd/embedded-cluster/join.go
```diff
@@ -368,3 +393,100 @@ func runK0sInstallCommand(fullcmd string) error {
...
+	installation, err := kubeutils.GetLatestInstallation(ctx, kcli)
+	if err != nil {
+		return false, fmt.Errorf("unable to get latest installation: %w", err)
+	}
+	return installation.Spec.HighAvailability, nil
+}
+
+// waitForNode waits for the node to be registered in the k8s cluster.
+// It returns the total number of nodes in the cluster.
+func waitForNode(ctx context.Context, kcli client.Client, name string) (int, error) {
```
sgalsaleh, line 441: should probably be moved to `kubeutils` package

cmd/embedded-cluster/join.go
cbodonnell: Do you think we should apply this _and_ the `waitForK0s` to all joins instead of it being wrapped in the `enable-ha` condition?

sgalsaleh: Yes

cmd/embedded-cluster/join.go
ajp-io: IMO we can open a story to track that for later. Someone can just remove and re-add the node if that happens, so it's not critical at this stage.

ajp-io: Let's discuss this wording too. I don't have a perfect suggestion now, but we should explain this more to a user. Like the fact that you can't go back, what does HA mean, etc.

ajp-io: And whenever you're able, I'd like to see a Loom of the output on join since we're changing it a bit.

cbodonnell: @ajp-io added a loom the PR description.

I will leave out this extra CLI command for now.

pkg/kubeutils/kubeutils.go
```diff
@@ -210,6 +220,44 @@ func writeStatusMessage(writer *spinner.MessageWriter, install embeddedclusterv1
 	}
 }
 
+func WaitForHAMigration(ctx context.Context, cli client.Client) error {
+	// TODO: determine an appropriate timeout
+	backoff := wait.Backoff{Steps: 60, Duration: 5 * time.Second, Factor: 1.0, Jitter: 0.1}
```
cbodonnell, line 225: Open to suggestions for how long we should wait for the HA migration.

go.mod
```diff
@@ -14,8 +14,8 @@ require (
 	github.com/jedib0t/go-pretty v4.3.0+incompatible
 	github.com/k0sproject/dig v0.2.0
 	github.com/k0sproject/k0s v1.29.6-0.20240527072442-22f6a125e881
-	github.com/replicatedhq/embedded-cluster-kinds v1.3.2
-	github.com/replicatedhq/embedded-cluster-operator v0.33.2
+	github.com/replicatedhq/embedded-cluster-kinds v1.3.3
+	github.com/replicatedhq/embedded-cluster-operator v0.34.2-0.20240531145409-ed39edc4db44
```
cbodonnell, line 18: TODO: change this once https://github.com/replicatedhq/embedded-cluster-operator/pull/196 is merged and operator is released

cmd/embedded-cluster/join.go
```diff
@@ -368,3 +412,69 @@ func runK0sInstallCommand(fullcmd string) error {
...
+	if installation.Spec.HighAvailability {
+		return false, nil
+	}
+	var nodes corev1.NodeList
+	if err := kcli.List(ctx, &nodes); err != nil {
+		return false, fmt.Errorf("unable to list nodes: %w", err)
+	}
+	if len(nodes.Items) < 3 {
+		return false, nil
+	}
```
laverya, line 457: we specifically need three controllers, no?

cbodonnell: Yes, thanks for that call out!

cbodonnell: addressed in https://github.com/replicatedhq/embedded-cluster/pull/657/commits/c71e6b0c8059aab4688f85cdf22652f5f09b6794

cmd/embedded-cluster/join.go
cbodonnell: Copying over a TODO from the loom comments:
> we should discuss how to ask the HA question. The user should know what they're choosing and what the ramifications are (like if you can't go back)

pkg/kubeutils/kubeutils.go
sgalsaleh: an exponential backoff here might not be a good idea for migrations that will take a very long time. i would suggest:

- for loop without a timeout (similar to how we wait for velero restores to complete)
- fixed sleep period (5s?)

cbodonnell: @sgalsaleh thanks for the feedback - definitely makes sense to avoid extra waiting

cbodonnell: Addressed in https://github.com/replicatedhq/embedded-cluster/pull/657/commits/53071a1b1830fc69b0f366120e284ff8a273fe93

cmd/embedded-cluster/join.go
```diff
@@ -368,3 +419,74 @@ func runK0sInstallCommand(fullcmd string) error {
...
+func canEnableHA(ctx context.Context, kcli client.Client) (bool, error) {
+	installation, err := kubeutils.GetLatestInstallation(ctx, kcli)
+	if err != nil {
+		return false, fmt.Errorf("unable to get latest installation: %w", err)
+	}
+	if installation.Spec.HighAvailability {
+		return false, nil
+	}
+	var nodes corev1.NodeList
+	opts := &client.ListOptions{LabelSelector: labels.Set(config.ControllerLabels()).AsSelector()}
```
sgalsaleh, line 463: What if we add more labels to `ControllerLabels` later? Wouldn't that affect existing nodes that don't have the additional labels? How about using the `node-role.kubernetes.io/control-plane` label instead? That's what we're using for node affinity rules for rqlite and seaweedfs.

pkg/kubeutils/kubeutils.go
```diff
@@ -210,6 +220,31 @@ func writeStatusMessage(writer *spinner.MessageWriter, install embeddedclusterv1
...
+		lastInstall, err := GetLatestInstallation(ctx, cli)
+		if err != nil {
+			return fmt.Errorf("unable to get latest installation: %v", err)
+		}
+		migrationStatus := CheckConditionStatus(lastInstall.Status, registry.RegistryMigrationStatusConditionType)
+		if migrationStatus == metav1.ConditionTrue {
+			return nil
+		}
+
+		time.Sleep(5 * time.Second)
```
laverya, line 234: Can you change this to use a select statement instead?
One of the select conditions would be the context closing

cbodonnell: updated this in https://github.com/replicatedhq/embedded-cluster/pull/657/commits/451d58cf08ad79d08537578eb1b045e40c2e5c95

cmd/embedded-cluster/join.go
cbodonnell: updated the selector in https://github.com/replicatedhq/embedded-cluster/pull/657/commits/a6b28b202b1d01742a034c00aebc953238705ca9

pkg/kubeutils/kubeutils.go
```diff
@@ -210,6 +220,36 @@ func writeStatusMessage(writer *spinner.MessageWriter, install embeddedclusterv1
...
+			lastInstall, err := GetLatestInstallation(ctx, cli)
+			if err != nil {
+				return fmt.Errorf("unable to get latest installation: %v", err)
+			}
+			migrationStatus := CheckConditionStatus(lastInstall.Status, registry.RegistryMigrationStatusConditionType)
+			if migrationStatus == metav1.ConditionTrue {
+				return nil
+			}
+
+			time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
cbodonnell: updated in https://github.com/replicatedhq/embedded-cluster/pull/657/commits/584d40f6b33e981319fe3705ad04acd3fb4b61b0

laverya (dismissed):

cmd/embedded-cluster/join.go
```diff
@@ -185,36 +197,75 @@ var joinCommand = &cli.Command{
...
 			return err
 		}
 
+		if err := waitForK0s(); err != nil {
+			err := fmt.Errorf("unable to wait for node: %w", err)
+			metrics.ReportJoinFailed(c.Context, jcmd.MetricsBaseURL, jcmd.ClusterID, err)
+			return err
+		}
+
+		if !strings.Contains(jcmd.K0sJoinCommand, "controller") {
```
sgalsaleh, line 234: does this mean we're not waiting for workers to join, only controllers?

sgalsaleh (approved):

cmd/embedded-cluster/join.go
cbodonnell: Yes, for right now. I had run into an issue with workers where I could not get a k8s client instantiated to do that check. Maybe we need to be looking elsewhere for the kubeconfig for workers?

PR #666, opened 2024-06-03
get apiserver & etcd certificates and their http health checks

adamancini:
bringing in some of the relevant collectors from https://github.com/replicatedhq/troubleshoot-specs/blob/main/host/default.yaml

### DIFF
--- a/pkg/goods/support/host-support-bundle.yaml
+++ b/pkg/goods/support/host-support-bundle.yaml
@@ -11,12 +11,81 @@ spec:
   - hostOS: {}
   - memory: {}
   - blockDevices: {}
+  - certificate:
+      collectorName: k8s-api-keypair
+      certificatePath: /var/lib/k0s/pki/k0s-api.crt
+      keyPath: /var/lib/k0s/pki/k0s-api.key
+  - certificate:
+      collectorName: etcd-keypair
+      certificatePath: /var/lib/k0s/pki/etcd/server.crt
+      keyPath: /var/lib/k0s/pki/etcd/server.key
+    # Disk usage for commonly used directories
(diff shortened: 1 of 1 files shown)


diamonwiggins (approved):

pkg/goods/support/host-support-bundle.yaml
```diff
@@ -5,116 +5,189 @@ metadata:
...
+              message: No IPv4 interfaces detected
+          - pass:
+              when: 'count >= 1'
+              message: IPv4 interface detected
+    - memory:
+        checkName: Amount of Memory
+        outcomes:
+          - warn:
+              when: '< 2G'
+              message: At least 2G of memory is recommended
```
chris-sanders, line 130: Moving forward with EC, we're going to try not having warnings. Either something is important enough to Fail or it isn't. Choose what you think is a fair solution here.

chris-sanders (dismissed):
There's a lot of changes here that I don't think are relevant. I'm guessing your editor ran this through a linter. Can you not lint/change things _or_ submit the lint as it's own with this change after that for easier review.

I tend to prefer just not running a lint, having different people submit different lint configs isn't very helpful. If we really want linting we should just add an enforced lint rule to the tests so it stays consistent.

pkg/goods/support/host-support-bundle.yaml
adamancini: :thumbsup:

banjoh (approved):

adamancini:
@chris-sanders fixed the linter changes, and moved some warns to fails

banjoh (approved):

PR #672, opened 2024-06-04
Add support for HA restores

sgalsaleh:


cmd/embedded-cluster/restore.go
```diff
@@ -39,28 +41,37 @@ import (
...
-	ecRestoreStateRestoreApp       ecRestoreState = "restore-app"
+	ecRestoreStateNew                 ecRestoreState = "new"
+	ecRestoreStateConfirmBackup       ecRestoreState = "confirm-backup"
+	ecRestoreStateRestoreECInstall    ecRestoreState = "restore-ec-install"
+	ecRestoreStateRestoreAdminConsole ecRestoreState = "restore-admin-console"
+	ecRestoreStateWaitForNodes        ecRestoreState = "wait-for-nodes"
+	ecRestoreStateRestoreSeaweedFS    ecRestoreState = "restore-seaweedfs"
+	ecRestoreStateRestoreRegistry     ecRestoreState = "restore-registry"
+	ecRestoreStateRestoreECO          ecRestoreState = "restore-embedded-cluster-operator"
+	ecRestoreStateRestoreApp          ecRestoreState = "restore-app"
```
laverya, line 52: wow that's a lot more states

cmd/embedded-cluster/restore.go
```diff
@@ -496,23 +515,104 @@ func waitForVeleroRestoreCompleted(ctx context.Context, restoreName string) (*ve
...
+	if err != nil {
+		return fmt.Errorf("unable to get registry service IP from backup: %w", err)
+	}
+	seaweedFSS3ServiceIP, err := getSeaweedFSS3ServiceIPFromBackup(backup)
+	if err != nil {
+		return fmt.Errorf("unable to get seaweedfs s3 service IP from backup: %w", err)
+	}
+
+	modifiersYAML := strings.Replace(resourceModifiersYAML, "__REGISTRY_SERVICE_IP__", registryServiceIP, 1)
+	modifiersYAML = strings.Replace(modifiersYAML, "__SEAWEEDFS_S3_SERVICE_IP__", seaweedFSS3ServiceIP, 1)
```
laverya, line 574: IMO leaving the number of replacements as `1` feels like it'll trip someone up if we ever add another `__SEAWEEDFS_S3_SERVICE_IP__` to the modifier yaml
but fine for now

e2e/restore_test.go
```diff
@@ -291,3 +301,431 @@ func TestSingleNodeAirgapDisasterRecovery(t *testing.T) {
...
+	tc := cluster.NewTestCluster(&cluster.Input{
+		T:                   t,
+		Nodes:               3,
+		Image:               "debian/12",
+		LicensePath:         "snapshot-license.yaml",
+		EmbeddedClusterPath: "../output/bin/embedded-cluster",
+	})
+	defer cleanupCluster(t, tc)
+
+	// install "expect" dependency on node 0 as that's where the restore process will be initiated.
```
laverya, line 336: Should we run the restore from something other than node 0?

Just to make sure it works on a different node?

sgalsaleh: I tried that, but hit a bunch of issues and had to give up given the time. It's because playwright is installed on node 0 by default and uses the node 0 IP as base URL. Doable, I don't wanna waste another day battling it though.

laverya (approved):

PR #676, opened 2024-06-06
Add support for proxy servers

jdewinne:


cmd/embedded-cluster/join.go
```diff
@@ -42,6 +48,7 @@ type JoinCommandResponse struct {
 	EndUserK0sConfigOverrides string    `json:"endUserK0sConfigOverrides"`
 	MetricsBaseURL            string    `json:"metricsBaseURL"`
 	AirgapRegistryAddress     string    `json:"airgapRegistryAddress"`
+	Proxy                     *Proxy    `json:"proxy"`
```
laverya, line 51: omitempty?
```suggestion
	Proxy                     *Proxy    `json:"proxy,omitempty"`
```

Makefile
```diff
@@ -66,6 +66,7 @@ LD_FLAGS = -X github.com/replicatedhq/embedded-cluster/pkg/defaults.K0sVersion=$
 	-X github.com/replicatedhq/embedded-cluster/pkg/addons/velero.Version=$(VELERO_CHART_VERSION) \
 	-X github.com/replicatedhq/embedded-cluster/pkg/addons/velero.VeleroTag=$(VELERO_IMAGE_VERSION) \
 	-X github.com/replicatedhq/embedded-cluster/pkg/addons/velero.AwsPluginTag=$(VELERO_AWS_PLUGIN_IMAGE_VERSION)
+SYNC_SERVER ?= root@reliabilityrunner-02
```
laverya, line 69: this seems like it was not supposed to be committed

cmd/embedded-cluster/uninstall.go
```diff
@@ -400,6 +400,13 @@ var resetCommand = &cli.Command{
 			}
 		}
 
+		proxyPath := "/etc/systemd/system/k0scontroller.service.d/http-proxy.conf"
```
laverya, line 403: what about workers?

cmd/embedded-cluster/join.go
jdewinne: Isn't this only needed in case we would marshal this also in embedded cluster? We do this within [kots](https://github.com/replicatedhq/kots/pull/4664/files#diff-416168ed19d2f0fac6d14240569d6e03f063d5ee3ffd5ee0dc6237d325d0040dR36).

.github/workflows/release-dev.yaml
```diff
@@ -168,6 +168,7 @@ jobs:
           - TestMultiNodeHAInstallation
           - TestMultiNodeAirgapHAInstallation
           - TestMultiNodeAirgapUpgradeSameK0s
+          - TestProxiedEnvironment
```
laverya, line 171: this file needs the same build change as `.github/workflows/pull-request.yaml` (to include `K0S_BINARY_SOURCE_OVERRIDE=$(make print-PREVIOUS_K0S_BINARY_SOURCE_OVERRIDE)`)

cmd/embedded-cluster/install.go
```diff
@@ -434,6 +441,21 @@ var installCommand = &cli.Command{
...
+			Hidden: true,
+		},
+		&cli.StringFlag{
+			Name:   "no-proxy",
+			Usage:  "Comma separated list of hosts to bypass the proxy for",
+			Hidden: true,
+		},
 		&cli.BoolFlag{
 			Name:   "proxy",
 			Usage:  "Use the system proxy settings for the install operation. These variables are currently only passed through to Velero and the Admin Console.",
```
laverya, line 461: is this usage info now out of date, since it follows the same codepath as the other proxy args?

cmd/embedded-cluster/join.go
laverya: fair enough

cmd/embedded-cluster/util.go
```diff
@@ -0,0 +1,65 @@
...
+	src := "/etc/systemd/system/k0scontroller.service"
+	if proxy != nil {
+		ensureProxyConfig(fmt.Sprintf("%s.d", src), proxy.HTTPProxy, proxy.HTTPSProxy, proxy.NoProxy)
+	}
+	if isWorker {
+		src = "/etc/systemd/system/k0sworker.service"
+		if proxy != nil {
+			ensureProxyConfig(fmt.Sprintf("%s.d", src), proxy.HTTPProxy, proxy.HTTPSProxy, proxy.NoProxy)
+		}
+	}
```
laverya, line 29:
```suggestion
	src := "/etc/systemd/system/k0scontroller.service"
	if isWorker {
		src = "/etc/systemd/system/k0sworker.service"
	}
	if proxy != nil {
		ensureProxyConfig(fmt.Sprintf("%s.d", src), proxy.HTTPProxy, proxy.HTTPSProxy, proxy.NoProxy)
	}
```

Unless you were intending to create two proxy config files on worker nodes?

pkg/defaults/defaults.go
```diff
@@ -21,6 +21,11 @@ var (
 	LocalArtifactMirrorImage = ""
 )
 
+// Holds the default no proxy values.
+var (
+	DefaultNoProxy = []string{"localhost", "127.0.0.1", ".default", ".local", ".svc", "kubernetes", "kotsadm-rqlite", "10.0.0.0/8"}
```
laverya, line 26: Including `10.0.0.0/8` in this with no way to remove that is going to cause problems

Ideally we would include the actual pod and service CIDRs in this instead

cmd/embedded-cluster/install.go
jdewinne: I'm pretty sure that `--proxy` doesn't work and never worked. It isn't passed through to the operator, so it will not have any effect.
We can clean it up in a follow-up PR, as it is used afaik by some e2e tests to mock an airgap environment.

cmd/embedded-cluster/util.go
jdewinne: I've seen that `/etc/systemd/system/k0sworker.service` isn't doing anything, but it might be that this has been fixed while working on the proxy?

cmd/embedded-cluster/install.go
laverya: it's used in our CI - it gets wiped away by operator, but it worked for long enough for our airgap tests to make a snapshot

laverya (approved):

laverya (approved):

pkg/defaults/defaults.go
```diff
@@ -23,7 +25,7 @@ var (
 
 // Holds the default no proxy values.
 var (
-	DefaultNoProxy = []string{"localhost", "127.0.0.1", ".default", ".local", ".svc", "kubernetes", "kotsadm-rqlite", "10.0.0.0/8"}
+	DefaultNoProxy = []string{"localhost", "127.0.0.1", ".default", ".local", ".svc", "kubernetes", "kotsadm-rqlite", v1beta1.DefaultNetwork().ServiceCIDR, v1beta1.DefaultNetwork().PodCIDR}
```
laverya, line 28: I'm approving despite this still being less than ideal - I'd prefer for this to include the CIDRs for the config being used in the cluster, as they might be overridden by the vendor/user in unsupported overrides

ricardomaraschini: Agree, I think this can be addressed in a follow up PR.

ricardomaraschini (approved):

pkg/defaults/defaults.go
jdewinne: Yes. We mentioned that once overriding CIDRs is supported, we should also support it.

PR #684, opened 2024-06-10
feat: ability to upgrade the cluster via the new binary

emosbaugh:


.github/workflows/release-dev.yaml
```diff
@@ -30,15 +30,15 @@ jobs:
 
       - name: Build and push local-artifact-mirror image
         run: |
-          export SHORT_SHA=dev-$(git rev-parse --short=7 HEAD)
+          export SHORT_SHA=dev-${GITHUB_SHA::7}
```
laverya, line 33: this should match what's in the pull-request.yaml version of things

laverya (commented):
Looks good besides the SHORT_SHA mismatch between pull-request.yaml and release-dev.yaml (we should really combine those soon)

sgalsaleh (approved):

.github/workflows/release-dev.yaml
sgalsaleh: can't use the same thing here though as `github.event.pull_request.head.sha` doesn't apply here, no?

sgalsaleh: i guess we could use `GITHUB_SHA` in `pull-request.yaml`?

laverya (approved):

PR #691, opened 2024-06-11
feat: build and publish local-artifact-mirror image

ricardomaraschini:
build and publish local-artifact-mirror image.

### DIFF
--- a/.github/workflows/release-prod.yaml
+++ b/.github/workflows/release-prod.yaml
@@ -25,6 +25,36 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Login to Registry
+        uses: docker/login-action@v2
+        with:
+          registry: registry.replicated.com
+          username: ${{ secrets.REGISTRY_USERNAME_PRODUCTION }}
+          password: ${{ secrets.REGISTRY_PASSWORD_PRODUCTION }}
+
+      - name: Build and push local-artifact-mirror image (production)
+        run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/release-prod.yaml
```diff
@@ -19,6 +19,24 @@ jobs:
...
+      - name: Login to Registry
+        uses: docker/login-action@v2
+        with:
+          registry: registry.replicated.com
+          username: ${{ secrets.REGISTRY_USERNAME_PRODUCTION }}
+          password: ${{ secrets.REGISTRY_PASSWORD_PRODUCTION }}
+
+      - name: Build and push local-artifact-mirror image
+        run: |
+          export SHORT_SHA=dev-$(git rev-parse --short=7 HEAD)
```
laverya, line 35: Are you sure that this should be using a short version and not the tag? 

Also, you should push the tag version to both prod and staging registries

laverya: Also... I would kind of recommend putting this as a different job, both here and in the onpr/onmerge workflows

ricardomaraschini: That is a good question. I guess we need it published for all releases we create (CI, staging and production). I think @sgalsaleh can answer this.

sgalsaleh: Yeah, let's use the tag when pushing to prod. We can push the tag to staging too, but IDK what would use it since I assume our CI will use the short SHA version.

ricardomaraschini: Done, please take a look.

laverya: > Yeah, let's use the tag when pushing to prod. We can push the tag to staging too, but IDK what would use it since I assume our CI will use the short SHA version.

Any time that we run test apps with real release versions on staging

laverya (approved):

PR #701, opened 2024-06-13
fix: metadata has wrong ec version

emosbaugh:
This is blocking work on new upgrade feature as wrong version is set in the new chart.


laverya:
I'm a little confused by this
Why not change what's output by the metadata command instead of mangling it afterwards?

laverya (dismissed):

sgalsaleh (approved):

PR #703, opened 2024-06-14
feat: push lam image to the embedded cluster registry

ricardomaraschini:
pushes the lam image to the embedde cluster registry.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ K0S_BINARY_SOURCE_OVERRIDE = https://ec-k0s-binaries.s3.amazonaws.com/k0s-v1.29.
 PREVIOUS_K0S_BINARY_SOURCE_OVERRIDE =
 TROUBLESHOOT_VERSION = v0.92.1
 KOTS_VERSION = v$(shell echo $(ADMIN_CONSOLE_CHART_VERSION) | sed 's/\([0-9]\+\.[0-9]\+\.[0-9]\+\).*/\1/')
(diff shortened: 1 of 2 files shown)


ricardomaraschini:
@ajp-io [Here](https://www.loom.com/share/f4e67692ba0144809794d696d8f126fc) is a recording of the change on the user feedback during installation. You might want to skip to 5 minutes. In case you didn't notice it: there is a new step called "Pushing embedded cluster images". Let me know what you think.

ajp-io:
Ideally this would be part of "Pushing embedded cluster artifacts." We've already gotten feedback that there is too much low-level detail in the output for basic non-Kubernetes end-users (what's Velero, what's Embedded Cluster, etc.), so I'd like to minimize how many additional lines we put in there. And since images are artifacts, it would be great to fit this into that message and just have the counter increment once more for this image.

ricardomaraschini:
> Ideally this would be part of "Pushing embedded cluster artifacts." We've already gotten feedback that there is too much low-level detail in the output for basic non-Kubernetes end-users (what's Velero, what's Embedded Cluster, etc.), so I'd like to minimize how many additional lines we put in there. And since images are artifacts, it would be great to fit this into that message and just have the counter increment once more for this image.

Hmmm, I think on this case we want to change the Kots output to include the image as part of the artifacts. Let me see how we can get it done there then.

ricardomaraschini:
After https://github.com/replicatedhq/kots/pull/4703 this became obsolete. Closing.

PR #728, opened 2024-06-24
add 'pod-cidr' and 'service-cidr' install/restore flags

laverya:


laverya:
This requires https://github.com/replicatedhq/kots/pull/4719 to function properly

laverya:
when restoring with a different set of CIDRs, the following message (or similar) is displayed:
```
        Found 1 backup, but it is not restorable:
        "instance-4wlxz" has a different network configuration than the current cluster, please run with '--pod-cidr 10.128.0.0/20 --service-cidr 10.129.0.0/20'
```

laverya:
> Service "registry" is invalid: spec.clusterIPs[0]: Invalid value: []string{"10.96.0.11"}: may not change once set

oh it's this one again

e2e/proxy_test.go
```diff
@@ -102,3 +102,204 @@ func TestProxiedEnvironment(t *testing.T) {
...
+	t.Parallel()
+	tc := cluster.NewTestCluster(&cluster.Input{
+		T:                   t,
+		Nodes:               4,
+		Image:               "debian/12",
+		LicensePath:         "license.yaml",
+		EmbeddedClusterPath: "../output/bin/embedded-cluster",
+	})
+	defer cleanupCluster(t, tc)
+	t.Log("Proxied infrastructure created")
```
sgalsaleh, line 117: infra isn't proxied here

e2e/proxy_test.go
```diff
@@ -102,3 +102,204 @@ func TestProxiedEnvironment(t *testing.T) {
...
+
+	// bootstrap the first node and makes sure it is healthy. also executes the kots
+	// ssl certificate configuration (kurl-proxy).
+	t.Logf("%s: installing embedded-cluster on node 0", time.Now().Format(time.RFC3339))
+	line := []string{"single-node-proxy-install.sh"}
```
sgalsaleh, line 122: is this supposed to be using a proxy install?

sgalsaleh (approved):

Makefile
```diff
@@ -3,9 +3,11 @@ UNAME := $(shell uname)
 ARCH := $(shell uname -m)
 APP_NAME = embedded-cluster
 ADMIN_CONSOLE_CHART_VERSION = 1.109.13
-ADMIN_CONSOLE_IMAGE_OVERRIDE =
+ADMIN_CONSOLE_IMAGE_OVERRIDE = ttl.sh/automated-9765912462/kotsadm:24h
```
sgalsaleh, line 6: I missed this. We should release KOTS first.

sgalsaleh (dismissed):

e2e/proxy_test.go
laverya: no, but the proxy install script already supported arbitrary additional args
I'll fix it

laverya: it is not using a proxy install despite the name of the script here

sgalsaleh (approved):

PR #732, opened 2024-07-01
chore(dev): uploading uncompressed ec release

ricardomaraschini:
uploads the uncompressed binary to the bucket, we are looking to evaluate if there is a performance gain in comparison.

this is temporary and should not be merged.


ricardomaraschini:
Tests indicate that using the uncompressed binary when serving embedded clusters did not bring performance gains. This isn't necessary anymore, closing.

PR #746, opened 2024-07-03
feat: Use filesystem specific checks for host preflights

hedge-sparrow:
sc-107808

### DIFF
--- a/pkg/preflights/host-preflight.yaml
+++ b/pkg/preflights/host-preflight.yaml
@@ -50,8 +50,8 @@ spec:
               when: 'total < 40Gi'
               message: The filesystem at the path /var/lib/k0s has less than 40Gi of total space
           - fail:
-              when: 'available < 10Gi'
-              message: The filesystem at the path /var/lib/k0s has less than 10Gi of disk space available
+              when: 'used/total > 80%'
+              message: The filesystem at the path /var/lib/k0s is more than 80% full
           - pass:
               message: The filesystem at the path /var/lib/k0s has sufficient space
     - diskUsage:
--- a/pkg/preflights/host-preflight.yaml
+++ b/pkg/preflights/host-preflight.yaml
@@ -7,6 +7,18 @@ spec:
     - diskUsage:
         collectorName: root-disk-usage
         path: /
+    - diskUsage:
+        collectorName: embedded-cluster-path-usage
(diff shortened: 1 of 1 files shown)


hedge-sparrow:
image

pkg/preflights/host-preflight.yaml
```diff
@@ -22,20 +34,56 @@ spec:
...
+              message: The filesystem at the path /opt/openebs is more than 80% full
+          - pass:
+              message: The filesystem at the path /opt/openebs has sufficient space
+    - diskUsage:
+        checkName: tmp path usage
+        collectorName: tmp-path-usage
+        outcomes:
+          - fail:
+              when: 'total < 5Gi'
+              message: The filesystem at the path /tmp has less than 40Gi of total space
```
ajp-io, line 81:
```suggestion
              message: The filesystem at the path /tmp has less than 5Gi of total space
```

pkg/preflights/host-preflight.yaml
```diff
@@ -22,20 +34,56 @@ spec:
...
+              message: The filesystem at the path /var/lib/k0s is more than 80% full
+          - pass:
+              message: The filesystem at the path /var/lib/k0s has sufficient space
+    - diskUsage:
+        checkName: openebs path usage
+        collectorName: openebs-path-usage
+        outcomes:
+          - fail:
+              when: 'total < 5Gi'
+              message: The filesystem at the path /opt/openebs has less than 40Gi of total space
```
ajp-io, line 69:
```suggestion
              message: The filesystem at the path /opt/openebs has less than 5Gi of total space
```

ajp-io (commented):
Another thing I discussed with Chris is that the used/total checks don't seem helpful in a few cases. Like for /tmp, if you have the needed 5Gi, you're fine. Checking the percentage doesn't add a lot more value, and it might even cause a problem if someone has a huge disk that is 90% full. The preflight will fail, but they could have plenty of space.

Chris mentioned that Kubernetes does start garbage collecting if a disk is 85% full. So the proposal is to remove the "percent full" checks, except for whichever one Kubernetes cares about. That's presumably /var/lib/k0s, but we should be sure.

ajp-io (approved):
lgtm

PR #757, opened 2024-07-08
feat: pull add on images through replicated proxy

ricardomaraschini:
replace add on image references in order to pull images using replicated proxy for oci artifacts.


Makefile
```diff
@@ -40,13 +43,16 @@ LD_FLAGS = -X github.com/replicatedhq/embedded-cluster/pkg/defaults.K0sVersion=$
 	-X github.com/replicatedhq/embedded-cluster/pkg/addons/embeddedclusteroperator.UtilsImage=$(EMBEDDED_OPERATOR_UTILS_IMAGE) \
 	-X github.com/replicatedhq/embedded-cluster/pkg/addons/embeddedclusteroperator.ImageOverride=$(EMBEDDED_CLUSTER_OPERATOR_IMAGE_OVERRIDE) \
 	-X github.com/replicatedhq/embedded-cluster/pkg/addons/openebs.Version=$(OPENEBS_CHART_VERSION) \
+	-X github.com/replicatedhq/embedded-cluster/pkg/addons/openebs.ImageOverride=$(OPENEBS_IMAGE_VERSION) \
 	-X github.com/replicatedhq/embedded-cluster/pkg/addons/openebs.UtilsVersion=$(OPENEBS_UTILS_VERSION) \
 	-X github.com/replicatedhq/embedded-cluster/pkg/addons/seaweedfs.Version=$(SEAWEEDFS_CHART_VERSION) \
+  -X github.com/replicatedhq/embedded-cluster/pkg/addons/seaweedfs.ImageOverride=$(SEAWEEDFS_IMAGE_VERSION) \
```
ricardomaraschini, line 49: small indentation problem here.

pkg/addons/velero/velero.go
```diff
@@ -56,6 +56,12 @@ var helmValues = map[string]interface{}{
 	"credentials": map[string]interface{}{
 		"existingSecret": credentialsSecretName,
 	},
+	"kubectl": map[string]interface{}{
+		"image": map[string]interface{}{
+			"repository": "proxy.replicated.com/anonymous/docker.io/bitnami/kubectl",
```
ricardomaraschini, line 61: this is on me, i don't think we need the `docker.io` prefix here.

cmd/buildtools/openebs.go
```diff
@@ -75,6 +76,66 @@ var updateOpenEBSAddonCommand = &cli.Command{
...
+		// openebs localpv image
+		// filter out image we want to pin
+		desired := []string{}
+		filterString := "openebs/provisioner-localpv"
+		desired = slices.Filter(desired, images, func(a string) bool {
+			i := strings.Split(a, ":")
+			return i[0] == filterString
+		})
+		// error if there's less or more than one image that matches our filter
+		fmt.Println(desired)
```
banjoh, line 95: I'd be confused by this print statement where a number just gets printed

cmd/buildtools/openebs.go
```diff
@@ -75,6 +76,66 @@ var updateOpenEBSAddonCommand = &cli.Command{
...
+
+		// openebs utils image (busybox)
+		desired = []string{}
+		filterString = "openebs/linux-utils"
+		desired = slices.Filter(desired, images, func(a string) bool {
+			i := strings.Split(a, ":")
+			return i[0] == filterString
+		})
+		// error if there's less or more than one image that matches our filter
+		fmt.Println(desired)
```
banjoh, line 121: ...and this one

cmd/buildtools/seaweedfs.go
```diff
@@ -34,6 +36,38 @@ var updateSeaweedFSAddonCommand = &cli.Command{
...
+		}
+		// filter out image we want to pin
+		desired := []string{}
+		filterString := "chrislusf/seaweedfs"
+		desired = slices.Filter(desired, images, func(a string) bool {
+			i := strings.Split(a, ":")
+			return i[0] == filterString
+		})
+		// error if there's less or more than one image that matches our filter
+		fmt.Println(desired)
```
banjoh, line 53: and this one

Makefile
```diff
@@ -11,14 +11,18 @@ EMBEDDED_OPERATOR_CHART_NAME = embedded-cluster-operator
 EMBEDDED_OPERATOR_CHART_VERSION = 0.36.5
 EMBEDDED_OPERATOR_UTILS_IMAGE = busybox:1.36.1
 EMBEDDED_CLUSTER_OPERATOR_IMAGE_OVERRIDE =
-OPENEBS_CHART_VERSION = 4.1.0
-OPENEBS_UTILS_VERSION = 4.1.0
+OPENEBS_CHART_VERSION = 4.0.1
+OPENEBS_IMAGE_VERSION = 4.0.0@sha256:9e45763b9d067bfef4b367b7b29a4d79d477897a8c7f9a16f4839a3a7e4b31a8
```
ricardomaraschini, line 15: This caught my eye, I looked into the helm chart and it seems to me that this should be `4.1.0` instead of `4.0.0`.

Makefile
```diff
@@ -11,14 +11,18 @@ EMBEDDED_OPERATOR_CHART_NAME = embedded-cluster-operator
 EMBEDDED_OPERATOR_CHART_VERSION = 0.36.5
 EMBEDDED_OPERATOR_UTILS_IMAGE = busybox:1.36.1
 EMBEDDED_CLUSTER_OPERATOR_IMAGE_OVERRIDE =
-OPENEBS_CHART_VERSION = 4.1.0
-OPENEBS_UTILS_VERSION = 4.1.0
+OPENEBS_CHART_VERSION = 4.0.1
```
ricardomaraschini, line 14: Somehow this seems to be downgrading OpenEBS. This does not feel right, maybe this is from before `4.1.0` became available.

cmd/buildtools/helm.go
```diff
@@ -188,3 +195,59 @@ func (h *Helm) Push(path, dst string) error {
...
+	// defaults
+	client.KubeVersion = &chartutil.KubeVersion{
+		Version: fmt.Sprintf("v%s.%s.0", k8sVersionMajor, k8sVersionMinor),
+		Major:   k8sVersionMajor,
+		Minor:   k8sVersionMinor,
+	}
+
+	chartRequested, err := loader.Load(chartPath)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to load chart")
```
ricardomaraschini, line 225: I think we are using `fmt.Errorf` everywhere so maybe we should use it on this file as well.

cmd/buildtools/helm.go
```diff
@@ -188,3 +195,59 @@ func (h *Helm) Push(path, dst string) error {
 	}
 	return up.UploadTo(path, dst)
 }
+
+func (h *Helm) Render(chartName string, chartPath string, vals map[string]interface{}, namespace string) ([]string, error) {
+
+	const k8sVersionMajor, k8sVersionMinor = "1", "26"
```
ricardomaraschini, line 201: Maybe we should use 1.28 as this is the oldest ec version we have. I recognise that I don't fully understand the impact of this, I assume it will render only types that are known to 1.28 (maybe do some conversion?). Is that it ?

cmd/buildtools/helm.go
```diff
@@ -188,3 +195,59 @@ func (h *Helm) Push(path, dst string) error {
...
+	rel, err := client.Run(chartRequested, vals)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to render chart")
+	}
+
+	var manifests bytes.Buffer
+	fmt.Fprintln(&manifests, strings.TrimSpace(rel.Manifest))
+	for _, m := range rel.Hooks {
+		fmt.Fprintf(&manifests, "---\n# Source: %s\n%s\n", m.Path, m.Manifest)
+	}
```
ricardomaraschini, line 243: IIUC we are sorta removing all spaces from `rel.Manifest` but we don't do the same with `m.Manifest`. Maybe the why will become clear as I keep reading the code.

cmd/buildtools/utils.go
```diff
@@ -114,6 +117,89 @@ func LatestChartVersion(repo, name string) (string, error) {
 	return hcli.Latest(repo, name)
 }
 
+func GetImageDigest(image string) (string, error) {
+	ref, err := docker.ParseReference("//" + image)
+	if err != nil {
+		panic(err)
```
ricardomaraschini, line 123: Can we return an error here ?

cmd/buildtools/utils.go
```diff
@@ -114,6 +117,89 @@ func LatestChartVersion(repo, name string) (string, error) {
...
+	}
+	ctx := context.Background()
```
ricardomaraschini, line 125: Can we receive this context as a function argument instead ?

pkg/addons/seaweedfs/seaweedfs.go
```diff
@@ -70,6 +71,12 @@ func (o *SeaweedFS) GenerateHelmConfig(onlyDefaults bool) ([]v1beta1.Chart, []v1
 		Order:     2,
 	}
 
+	if ImageVersion != "" {
```
ricardomaraschini, line 74: Doesn't this always evaluate to true ?

Makefile
```diff
@@ -2,23 +2,34 @@ VERSION ?= $(shell git describe --tags --dirty)
...
 OPENEBS_CHART_VERSION = 4.1.0
-OPENEBS_UTILS_VERSION = 4.1.0
+OPENEBS_IMAGE_TAG = 4.1.0@sha256:7a9bb247229564c74b5c4981304bf614f1b0157797a52c6d882d5f1cb057b838
+OPENEBS_KUBECTL_IMAGE_TAG = 1.25.15@sha256:da199debb691621e0dc6954fefdaaf833900e58e28b96a4d805ff01ac397a979
+OPENEBS_UTILS_IMAGE_TAG = 4.1.0@sha256:a0f8fd9457388d5728e8e9c41f1cf5dc3b21ef8f00d7b019520101812ae602dc
 SEAWEEDFS_CHART_VERSION = 4.0.0
+SEAWEEDFS_IMAGE_TAG = 3.69@sha256:88eaab1398cd95b1e49f7210a36af389b4ebc896cccd2ef6f4baff3fa1b245fd
 REGISTRY_CHART_VERSION = 2.2.3
-REGISTRY_IMAGE_VERSION = 2.8.3
+REGISTRY_IMAGE_TAG = 2.8.1@sha256:85d4cd48754d4e8e4d4fe933bbde017d969e0e20775507add9faca39064bf9b0
```
ricardomaraschini, line 27: It seems the automation erroneously bumped this in the past. The latest registry chart (`v2.2.3`) ships with registry `2.8.1`.  I don't expect to see any problems once this is in I need to test an upgrade just to be sure.

Makefile
```diff
@@ -30,28 +41,41 @@ KOTS_VERSION = v$(shell echo $(ADMIN_CONSOLE_CHART_VERSION) | sed 's/\([0-9]\+\.
...
+	-X github.com/replicatedhq/embedded-cluster/pkg/addons/openebs.OpenEBSKubectlImageTag=$(OPENEBS_KUBECTL_IMAGE_TAG) \
+	-X github.com/replicatedhq/embedded-cluster/pkg/addons/seaweedfs.SeaweedFSChartVersion=$(SEAWEEDFS_CHART_VERSION) \
+	-X github.com/replicatedhq/embedded-cluster/pkg/addons/seaweedfs.SeaweedFSImageTag=$(SEAWEEDFS_IMAGE_TAG) \
+	-X github.com/replicatedhq/embedded-cluster/pkg/addons/registry.RegistryChartVersion=$(REGISTRY_CHART_VERSION) \
+	-X github.com/replicatedhq/embedded-cluster/pkg/addons/registry.RegistryImageTag=$(REGISTRY_IMAGE_TAG) \
+	-X github.com/replicatedhq/embedded-cluster/pkg/addons/velero.VeleroChartVersion=$(VELERO_CHART_VERSION) \
+	-X github.com/replicatedhq/embedded-cluster/pkg/addons/velero.VeleroImageTag=$(VELERO_IMAGE_TAG) \
+	-X github.com/replicatedhq/embedded-cluster/pkg/addons/velero.VeleroAWSPluginImageTag=$(VELERO_AWS_PLUGIN_IMAGE_TAG) \
+	-X github.com/replicatedhq/embedded-cluster/pkg/addons/velero.VeleroKubectlImageTag=$(VELERO_KUBECTL_IMAGE_TAG) \
+	-X github.com/replicatedhq/embedded-cluster/pkg/addons/velero.VeleroRestoreHelperImageTag=$(VELERO_RESTORE_HELPER_IMAGE_TAG)
```
ricardomaraschini, line 78: I am inclined to make the automation to write this down to a `json` file in each add-on directory that then can be embedded. What do you think ? We would not need to have any of these tags.

Makefile
laverya: I would strongly prefer to keep using the 2.8.3 image version

Makefile
emosbaugh: along the same lines it would be pretty cool since buildtools discovers images in charts if it were to generate the values yaml file in each add-on directory to override images

emosbaugh:
lgtm pending andrew's comment and merge of the other pr

pkg/addons/adminconsole/static/values.tpl.yaml
```diff
@@ -0,0 +1,19 @@
+images:
+    kotsadm: 'proxy.replicated.com/anonymous/kotsadm/kotsadm:{{ index .Images "kotsadm/kotsadm" }}'
```
emosbaugh, line 2: I think you missed my point here but I realize what I said was very unclear. Right now there is a hardcoded list of images in this values file. But your new code discovers images in helm charts. What if a new image gets added? I thought it may be cool to dynamically generate the helm values.yaml overriding images so we would automatically proxy new images. That way we would never make a release with an image we didn't know about and therefore break our rules around whitelisted domains. It was just a thought and I'm happy to approve this pull request as is.

pkg/addons/embeddedclusteroperator/embeddedclusteroperator.go
```diff
@@ -138,7 +143,15 @@ func (e *EmbeddedClusterOperator) GenerateHelmConfig(onlyDefaults bool) ([]v1bet
 }
 
 func (e *EmbeddedClusterOperator) GetAdditionalImages() []string {
-	return []string{UtilsImage}
+	for image, tag := range Metadata.Images {
```
emosbaugh, line 146: Do you really need a loop here since this is a map?

pkg/addons/adminconsole/static/values.tpl.yaml
emosbaugh: Oh I see why this does not make sense now. Since you are parsing the rendered yaml and not the values files. But I do wonder if we could compare the list of discovered images with the list of expect images and somehow warn.

emosbaugh: oh i see this is not possible since you are extracting images from the rendered yaml but not the values files. as a follow up it may be nice to provide a sanity check, since we have a hardcoded list of images we are expecting, and if we find are any images we do not know about in the rendered template produce an error.

emosbaugh (approved):

Makefile
ricardomaraschini: done.

pkg/addons/embeddedclusteroperator/embeddedclusteroperator.go
ricardomaraschini: No, that is not necessary. Solved.

cmd/buildtools/utils.go
```diff
@@ -240,12 +257,200 @@ func SetMakefileVariable(name, value string) error {
...
+	}
+	return "", fmt.Errorf("failed to locate linux/amd64 manifest")
+}
+
+// easier than walking through map[interface{}]interface{}
+type ReducedContainer struct {
+	Image string `yaml:"image"`
+}
+
+type RecudedNestedSpec struct {
```
sgalsaleh, line 315: typo
```suggestion
type ReducedNestedSpec struct {
```

cmd/buildtools/velero.go
```diff
@@ -13,43 +16,83 @@ var updateVeleroAddonCommand = &cli.Command{
 	UsageText: environmentUsageText,
 	Action: func(c *cli.Context) error {
 		logrus.Infof("updating velero addon")
-		latest, err := GetLatestGitHubTag(c.Context, "vmware-tanzu", "velero")
-		if err != nil {
-			return fmt.Errorf("failed to get latest velero release: %w", err)
+
+		// XXX remove-me.
```
sgalsaleh, line 20: is this something that should be done now or later at some point?

pkg/addons/openebs/static/metadata.yaml
```diff
@@ -0,0 +1,42 @@
+#
+# this file is automatically generated by buildtools. manual edits are not recommended.
+# to regenerate this file, run the following commands:
+#
+# $ make buildtools
+# $ output/bin/buildtools update addon <addon name>
+#
+version: 4.1.0
+location: oci://proxy.replicated.com/anonymous/ttl.sh/ricardox/openebs
```
sgalsaleh, line 9: this image location uses ttl.sh, this doesn't seem expected

pkg/addons/openebs/static/metadata.yaml
```diff
@@ -0,0 +1,42 @@
...
+images:
```
sgalsaleh, line 10: i assume these are all the images found in the chart, but not necessarily the ones we actually end up using?

pkg/addons/openebs/static/metadata.yaml
ricardomaraschini: These references to ttl are there for testing purposes. :-) Let me clean this up.

cmd/buildtools/velero.go
ricardomaraschini: Yeah, this should be removed before merging.

pkg/addons/openebs/static/metadata.yaml
sgalsaleh: no problem. why are there so many images? the airgap builder doesn't include all of these images in the airgap bundle.

ricardomaraschini: solved as discussed.

cmd/buildtools/velero.go
ricardomaraschini: removed.

cmd/buildtools/utils.go
ricardomaraschini: done.

emosbaugh (approved):

sgalsaleh (approved):

PR #758, opened 2024-07-08
chore: Add host preflights to check k0s system requirements

banjoh:
Craft host preflights for checking `k0s` [system requirements](https://docs.k0sproject.io/stable/system-requirements/). This will match what `k0s sysinfo` inspects and reports errors on


e2e/scripts/common.sh
```diff
@@ -0,0 +1,21 @@
+wait_for_healthy_node() {
```
laverya, line 1: thank you for finally making a common script file
there's a lot of stuff that should get moved here

laverya (approved):

PR #771, opened 2024-07-11
Add host-preflight to check for default route

hedge-sparrow:
This prevents a situation where an airgaped cluster install can be left hanging waiting for calico to initialize.

### DIFF
--- a/pkg/preflights/host-preflight.yaml
+++ b/pkg/preflights/host-preflight.yaml
@@ -24,7 +24,22 @@ spec:
     - time: {}
     - ipv4Interfaces: {}
     - kernelConfigs: {}
+    - run:
+        collectorName: "ip-route-table"
+        command: "ip"
+        args: ["route"]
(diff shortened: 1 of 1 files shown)


ricardomaraschini (approved):

pkg/preflights/host-preflight.yaml
```diff
@@ -24,7 +24,22 @@ spec:
...
+        args: ["route"]
   analyzers:
+    - textAnalyze:
+        checkName: Default route
+        fileName: host-collectors/run-host/ip-route-table.txt
+        regex: "default"
+        outcomes:
+          - fail:
+              when: "false"
+              message: "No default route found. A default route is required for cluster initialization"
```
ajp-io, line 39:
```suggestion
              message: "No default route found. A default route is required for cluster initialization."
```

ricardomaraschini (approved):

PR #786, opened 2024-07-18
delete contents of symlinked dirs before deleting the symlink when resetting

laverya:


chris-sanders:
From standup today. While this is an easy fix for this particular bug. The use of symlinks will represent a continual progression of edge cases that we don't want to deal with. We also know that Ceph will not allow it, there are hard limits to how far we can allow people to use symlinks.

I don't want to send the message that we're supporting using symlinks, the work around to just rm the contents of the folder is an acceptable one and we have a path to a proper solution. I'm going to close this one, not because it doesn't work but because we don't want to encourage the continued use of symlinks which will cause loss of trust in the product.

PR #794, opened 2024-07-18
feat: use chainguard for OpenEBS images

emosbaugh:


Makefile
```diff
@@ -24,7 +24,14 @@ EMBEDDED_OPERATOR_UTILS_IMAGE_VERSION ?= $(subst +,-,$(VERSION))
 EMBEDDED_OPERATOR_UTILS_IMAGE_LOCATION = proxy.replicated.com/anonymous/$(EMBEDDED_OPERATOR_UTILS_IMAGE):$(EMBEDDED_OPERATOR_UTILS_IMAGE_VERSION)
 EMBEDDED_CLUSTER_OPERATOR_IMAGE_OVERRIDE =
 OPENEBS_CHART_VERSION = 4.1.0
-OPENEBS_UTILS_VERSION = 4.1.0
+OPENEBS_IMAGE_VERSION = 4.1.0
+OPENEBS_KUBECTL_IMAGE_VERSION = 1.29
```
emosbaugh, line 28: I think i still have some work here to figure out exactly what version will end up as the input here. the default version is 1.25.15 but this is not available in chainguard.

https://github.com/openebs/openebs/blob/main/charts/values.yaml#L58

cmd/buildtools/openebs.go
```diff
@@ -78,3 +78,115 @@ var updateOpenEBSAddonCommand = &cli.Command{
 		return nil
 	},
 }
+
+type addonComponent struct {
```
sgalsaleh, line 82: can this be moved to a separate file as it'll be used by other addons? maybe `addon.go`?

cmd/buildtools/openebs.go
```diff
@@ -78,3 +78,115 @@ var updateOpenEBSAddonCommand = &cli.Command{
...
+	getWolfiPackageName         func(upstreamVersion string) string
+	upstreamVersionMakefileVar  string
+	upstreamVersionFlagOverride string
+}
+
+var openebsComponents = []addonComponent{
+	{
+		name:                        "openebs-provisioner-localpv",
+		wolfiPackageName:            "dynamic-localpv-provisioner",
+		upstreamVersionMakefileVar:  "OPENEBS_IMAGE_VERSION",
```
sgalsaleh, line 94: should we just extract these versions from the charts instead of tracking them in the Makefile? we already have the code for that.

cmd/buildtools/openebs.go
```diff
@@ -78,3 +78,115 @@ var updateOpenEBSAddonCommand = &cli.Command{
 		return nil
 	},
 }
+
+type addonComponent struct {
+	name                        string
+	wolfiPackageName            string
```
sgalsaleh, line 84: nit: how about just always using `getWolfiPackageName` and changing its function header to something like:
```golang
func(args ...string) string
```

cmd/buildtools/openebs.go
```diff
@@ -78,3 +78,115 @@ var updateOpenEBSAddonCommand = &cli.Command{
...
+			return fmt.Errorf("failed to apko login: %w", err)
+		}
+
+		wolfiAPKIndex, err := GetWolfiAPKIndex()
+		if err != nil {
+			return fmt.Errorf("failed to get APK index: %w", err)
+		}
+
+		for _, component := range openebsComponents {
+			upstreamVersion, err := getAddonComponentUpstreamVersion(c, component)
```
sgalsaleh, line 137: nit: can we make this a field or a method of the `addonComponent` struct and maybe call it `getUpstreamVersion`? seems inconsistent with how we're handling `getWolfiPackageName` for example.

cmd/buildtools/openebs.go
```diff
@@ -78,3 +78,115 @@ var updateOpenEBSAddonCommand = &cli.Command{
...
+			if err != nil {
+				return fmt.Errorf("failed to get upstream version for %s: %w", component.name, err)
+			}
+
+			packageName := component.wolfiPackageName
+			if component.getWolfiPackageName != nil {
+				packageName = component.getWolfiPackageName(upstreamVersion)
+			}
+			packageVersion := upstreamVersion
+			if packageName != "" {
```
sgalsaleh, line 147: should this just fail if the package name is empty?

cmd/buildtools/openebs.go
```diff
@@ -78,3 +78,115 @@ var updateOpenEBSAddonCommand = &cli.Command{
...
+			if err := ApkoBuildAndPublish(component.name, packageVersion, extraArgs...); err != nil {
+				return fmt.Errorf("failed to apko build and publish for %s: %w", component.name, err)
+			}
+
+			digest, err := GetDigestFromBuildFile()
+			if err != nil {
+				return fmt.Errorf("failed to get digest from build file: %w", err)
+			}
+
+			makefileVar := getAddonComponentImageTagMakefileVar(component)
```
sgalsaleh, line 169: nit: ditto on making it a field/method of the `addonComponent` struct

cmd/buildtools/openebs.go
```diff
@@ -78,3 +78,115 @@ var updateOpenEBSAddonCommand = &cli.Command{
...
+				if err != nil {
+					return fmt.Errorf("failed to get package version for %s: %w", component.name, err)
+				}
+			}
+
+			logrus.Infof("building and publishing %s=%s", component.name, packageVersion)
+
+			extraArgs := []string{}
+			if packageName != "" {
+				extraArgs = append(extraArgs, fmt.Sprintf("PACKAGE_NAME=%s", packageName))
```
sgalsaleh, line 158: why not just always set this env var instead of passing it as an extra arg?

cmd/buildtools/openebs.go
emosbaugh: there may not be a wolfi package for example openebs/linux-utils

cmd/buildtools/openebs.go
sgalsaleh: i think you maybe misunderstood my point. I'm saying why not just set `PACKAGE_NAME` here similar to `PACKAGE_VERSION` instead of passing it as an extra arg?: https://github.com/replicatedhq/embedded-cluster/pull/794/files#diff-1d4bdab5fd148a75dead93758db106572c100974fc9a02efbb9125e9522e5db5R193

that way, the `PACKAGE_NAME` var would be automatically templated for other addons/packages as well when used in the apko file without having to explicitly pass it as an extra arg here.

cmd/buildtools/openebs.go
```diff
@@ -57,25 +98,49 @@ var updateOpenEBSAddonCommand = &cli.Command{
...
 		for _, image := range images {
-			sha, err := GetImageDigest(c.Context, image)
+			logrus.Infof("updating image %s", image)
+
+			upstreamVersion := TagFromImage(image)
+			image = RemoveTagFromImage(image)
+
+			componentName, ok := openebsImageComponents[image]
+			if !ok {
+				logrus.Warnf("no component found for image %s", image)
```
emosbaugh, line 117: should this be an error?

ricardomaraschini: I think this is a good idea.

Makefile
ricardomaraschini: What about using the latest for the K0S major version ?

cmd/buildtools/openebs.go
sgalsaleh: it probably should so we can know if the chart adds a new image that we aren't building with chainguard

sgalsaleh: looks like it already fails below if the component isn't found, so...

cmd/buildtools/openebs.go
```diff
@@ -88,3 +152,104 @@ var updateOpenEBSAddonCommand = &cli.Command{
...
+		}
+
+		current := openebs.Metadata
+		newmeta := release.AddonMetadata{
+			Version:  current.Version,
+			Location: current.Location,
+			Images:   make(map[string]string),
+		}
+
+		values, err := release.GetValuesWithOriginalImages("openebs")
```
sgalsaleh, line 192: nit: correct me if i'm wrong, but from this line onwards, the logic is the same as `updateOpenEBSAddonCommand`, right?

cmd/buildtools/utils.go
```diff
@@ -73,65 +112,92 @@ func GetWolfiPackageVersion(wolfiAPKIndex []byte, pkgName string, pinnedVersion
...
+// FindWolfiPackageVersion returns the latest version and revision of a package in the wolfi APK
+// index that matches the semver constraints.
+func FindWolfiPackageVersion(wolfiAPKIndex []byte, pkgName string, constraints *semver.Constraints) (string, error) {
+	versions, err := listMatchingWolfiPackageVersions(wolfiAPKIndex, pkgName, constraints)
+	if err != nil {
+		return "", fmt.Errorf("list package versions: %w", err)
+	}
+
+	if len(versions) == 0 {
+		return "", fmt.Errorf("package %q not found", pkgName)
```
sgalsaleh, line 162: nit: it could be that the package exists, but not with the provided constraints

sgalsaleh (approved):

PR #798, opened 2024-07-19
fix HA cluster upgrades

laverya:


pkg/highavailability/registry.go
```diff
@@ -0,0 +1,74 @@
...
+		Data: map[string][]byte{
+			"s3AccessKey": []byte(sfsCreds.AccessKey),
+			"s3SecretKey": []byte(sfsCreds.SecretKey),
+		},
+	}
+
+	obj.ObjectMeta.Labels = applySeaweedFSLabels(obj.ObjectMeta.Labels, "s3")
+
+	err := cli.Create(ctx, obj)
+	if err != nil {
```
sgalsaleh, line 70:
```suggestion
	if err != nil && !k8serrors.IsAlreadyExists(err) {
```

pkg/highavailability/seaweed.go
```diff
@@ -0,0 +1,202 @@
...
+			Selector: map[string]string{
+				"app.kubernetes.io/component": "filer",
+				"app.kubernetes.io/name":      "seaweedfs",
+			},
+		},
+	}
+	obj.ObjectMeta.Labels = applySeaweedFSLabels(obj.ObjectMeta.Labels, "s3")
+
+	err = cli.Create(ctx, obj)
+	if err != nil {
```
sgalsaleh, line 137:
```suggestion
	if err != nil && !k8serrors.IsAlreadyExists(err) {
```

pkg/highavailability/seaweed.go
```diff
@@ -0,0 +1,202 @@
...
+		ObjectMeta: metav1.ObjectMeta{Name: seaweedfsS3SecretName, Namespace: defaults.SeaweedFSNamespace},
+		Data: map[string][]byte{
+			"seaweedfs_s3_config": configData,
+		},
+	}
+
+	obj.ObjectMeta.Labels = applySeaweedFSLabels(obj.ObjectMeta.Labels, "s3")
+
+	err = cli.Create(ctx, obj)
+	if err != nil {
```
sgalsaleh, line 178:
```suggestion
	if err != nil && !k8serrors.IsAlreadyExists(err) {
```

pkg/highavailability/registry.go
laverya: I think failing if it already exists here is actually desired - this migration should only be run once, and if the resource already exists that means it's being run in parallel

sgalsaleh: what if creating the seaweed resources succeeded but creating the registry resources failed, so the `HA` field didn't get updated in the installation object. then it will always fail on retry unless the resource gets deleted manually.

sgalsaleh (approved):

PR #808, opened 2024-07-23
chore: Add hostname mismatch support bundle check

nvanthao:
Fixes: https://app.shortcut.com/replicated/story/108807/analyzers-in-default-spec-not-compatible-with-k0s-deployments
Demo: https://asciinema.org/a/jX7izIBEaFbWZP9pTUiWvvmap


xavpaice (approved):
Change makes sense.  Our previous analyzer relies on a different service name, which doesn't work for k0s.

ricardomaraschini (approved):

PR #816, opened 2024-07-23
feat: use chainguard for velero

emosbaugh:


cmd/buildtools/utils.go
```diff
@@ -255,6 +255,38 @@ func GetLatestGitHubTag(ctx context.Context, owner, repo string) (string, error)
 	return tags[0].GetName(), nil
 }
 
+func GetBestGitHubTag(ctx context.Context, owner, repo string, constrants *semver.Constraints) (string, error) {
+	client := github.NewClient(nil)
+	tags, _, err := client.Repositories.ListTags(ctx, owner, repo, &github.ListOptions{})
+	if err != nil {
+		return "", fmt.Errorf("unable to list tags: %w", err)
```
sgalsaleh, line 262:
```suggestion
		return "", fmt.Errorf("list tags: %w", err)
```

cmd/buildtools/utils.go
```diff
@@ -255,6 +255,38 @@ func GetLatestGitHubTag(ctx context.Context, owner, repo string) (string, error)
 	return tags[0].GetName(), nil
 }
 
+func GetBestGitHubTag(ctx context.Context, owner, repo string, constrants *semver.Constraints) (string, error) {
```
sgalsaleh, line 258: What does "Best" mean? :)

sgalsaleh: baybe just `GetGitHubTag`?

sgalsaleh (approved):

cmd/buildtools/utils.go
emosbaugh: more descriptive would be something like GetGreatestGitHubTag

PR #817, opened 2024-07-23
Build kube-proxy image with Chainguard

sgalsaleh:


cmd/buildtools/k0s.go
```diff
@@ -13,30 +13,56 @@ import (
...
 	},
+	"kube-proxy": {
+		getWolfiPackageName: func(k0sVersion *semver.Version, upstreamVersion string) string {
+			return fmt.Sprintf("kube-proxy-%d.%d-default", k0sVersion.Major(), k0sVersion.Minor())
+		},
+		getWolfiPackageVersionComparison: func(k0sVersion *semver.Version, upstreamVersion string) string {
+			// match the greatest patch version of the same minor version
+			return fmt.Sprintf(">=%d.%d, <%d.%d", k0sVersion.Major(), k0sVersion.Minor(), k0sVersion.Major(), k0sVersion.Minor()+1)
+		},
+		makefileVar: "KUBE_PROXY_VERSION",
```
emosbaugh, line 64: eventuallly we should move these to a metadata.yaml file as ricardo has for the addons

cmd/buildtools/k0s.go
```diff
@@ -13,30 +13,56 @@ import (
...
+		},
 		makefileVar: "METRICS_SERVER_VERSION",
 	},
+	"kube-proxy": {
+		getWolfiPackageName: func(k0sVersion *semver.Version, upstreamVersion string) string {
+			return fmt.Sprintf("kube-proxy-%d.%d-default", k0sVersion.Major(), k0sVersion.Minor())
+		},
+		getWolfiPackageVersionComparison: func(k0sVersion *semver.Version, upstreamVersion string) string {
+			// match the greatest patch version of the same minor version
+			return fmt.Sprintf(">=%d.%d, <%d.%d", k0sVersion.Major(), k0sVersion.Minor(), k0sVersion.Major(), k0sVersion.Minor()+1)
```
emosbaugh, line 62: it looks like this version tracks the k0s version upstream. i assume the issue here is 1.29.6 is not available?

cmd/buildtools/k0s.go
```diff
@@ -92,58 +135,35 @@ var updateK0sImagesCommand = &cli.Command{
...
 			}
 		}
 
 		return nil
 	},
 }
 
-func makeK0s(k0sVersion string) error {
-	if k0sVersion != "" {
-		cmd := exec.Command("make", "pkg/goods/bins/k0s", fmt.Sprintf("K0S_VERSION=%s", k0sVersion), "K0S_BINARY_SOURCE_OVERRIDE=")
```
emosbaugh, line 105: why lose this?

emosbaugh (approved):

cmd/buildtools/k0s.go
sgalsaleh: yes

cmd/buildtools/k0s.go
sgalsaleh: you mean `K0S_BINARY_SOURCE_OVERRIDE=`?

emosbaugh: yes

sgalsaleh: if we're overriding the k0s binary, why wouldn't we want to get the images from the custom binary?

emosbaugh: but it makes no sense when passing in k0s_verison as input

sgalsaleh: how so? custom k0s binaries aren't downloadable from the k0s github project, only via the `K0S_BINARY_SOURCE_OVERRIDE` url.

so this workflow would break if we ever set `K0S_VERSION` to a custom build version.

sgalsaleh: i guess the conflict is that i'm reading the `K0S_VERSION` from the Makefile if it's not passed as an input: https://github.com/replicatedhq/embedded-cluster/pull/817/files#diff-b7c04cb7731e67d487af4a6bf8367637e1da8e995aaf361cae5c41dc110eae90R79

maybe i should change that

sgalsaleh: but i use that version is used later on. hmm. i'll think about it

cmd/buildtools/k0s.go
sgalsaleh: i'll do that on the next pr

cmd/buildtools/k0s.go
sgalsaleh: will fix on the following pr

PR #821, opened 2024-07-24
feat: remove unused k0s images from airgap bundle

emosbaugh:


pkg/config/images.go
```diff
@@ -22,11 +24,22 @@ func init() {
 	}
 }
 
-func OverrideK0sImages(cfg *k0sv1beta1.ClusterConfig) error {
-	if len(Metadata.Images) == 0 {
-		return fmt.Errorf("no images found in metadata")
+func ListK0sImages(cfg *k0sconfig.ClusterConfig) []string {
+	var images []string
+	for _, image := range airgap.GetImageURIs(cfg.Spec, true) {
```
sgalsaleh, line 29: hmm, now i get why you were concerned about the pause image and `GetImageURIs`. as this function will always returned the original image even if we override it

sgalsaleh (approved):

PR #822, opened 2024-07-24
check that etcdserver database size is not too large

adamancini:


pkg/goods/support/host-support-bundle.yaml
```diff
@@ -202,3 +202,16 @@ spec:
         - pass:
             when: "false"
             message: "No signs of hostname changes found"
+  - textAnalyze:
+        checkName: Etcdserver Database Size Exceeded
+        exclude: ""
+        ignoreIfNoFiles: true
+        fileName: *
```
diamonwiggins, line 209: lgtm, just curious if this works? never knew if so.

diamonwiggins (approved):

PR #827, opened 2024-07-24
feat: scan images for CVEs

emosbaugh:


cmd/embedded-cluster/metadata.go
```diff
@@ -0,0 +1,148 @@
...
+	charts := append([]eckinds.Chart{}, chtconfig...)
+	for _, chart := range builtinCharts {
+		repos = append(repos, chart.Repositories...)
+		charts = append(charts, chart.Charts...)
+	}
+	images, err := config.ExtractImagesFromHelmExtensions(repos, charts)
+	if err != nil {
+		return nil, fmt.Errorf("unable to extract images from helm extensions: %w", err)
+	}
+	meta.Images = append(meta.Images, images...)
```
emosbaugh, line 140: This is the only real new code, otherwise the rest was moved. This needs to have an internet connection because it pulls charts. Ideally we embed all charts in the binary and make it truly "embedded".

cmd/embedded-cluster/version.go
```diff
@@ -56,105 +54,6 @@ var versionCommand = &cli.Command{
 	},
 }
 
-var metadataCommand = &cli.Command{
```
emosbaugh, line 59: this was moved to a new file.

cmd/buildtools/metadata.go
```diff
@@ -0,0 +1,133 @@
...
+	if err := json.Unmarshal(b, &metadata); err != nil {
+		return nil, fmt.Errorf("unmarshal metadata: %w", err)
+	}
+
+	return &metadata, nil
+}
+
+func extractImagesFromHelmExtensions(repos []embeddedclusterv1beta1.Repository, charts []embeddedclusterv1beta1.Chart) ([]string, error) {
+	hcli, err := helm.NewHelm(helm.HelmOptions{
+		K0sVersion: "1.29.6", // TODO
```
sgalsaleh, line 101: can get this from `defaults.K0sVersion`.

cmd/buildtools/utils.go
```diff
@@ -688,3 +574,14 @@ func RunCommand(cmd *exec.Cmd) error {
 	cmd.Stderr = os.Stderr
 	return cmd.Run()
 }
+
+func NewHelm() (*helm.Helm, error) {
+	k0sVersion, err := GetMakefileVariable("K0S_VERSION")
```
sgalsaleh, line 579: should use the existing `getK0sVersion` function since that accounts for `INPUT_K0S_VERSION`

cmd/embedded-cluster/list_images.go
```diff
@@ -0,0 +1,25 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/urfave/cli/v2"
+)
+
+var listImagesCommand = &cli.Command{
+	Name:   "list-images",
```
sgalsaleh, line 10: how is this different from the metadata list-images command?

cmd/buildtools/metadata.go
emosbaugh: this is in buildtools so probably not

cmd/embedded-cluster/list_images.go
emosbaugh: its not just two different tools. metadata list-images is unused

cmd/embedded-cluster/metadata.go
sgalsaleh: this is only needed before we release though, right?

cmd/buildtools/metadata.go
sgalsaleh: ah in that case, maybe can use the `getK0sVersion` function?

cmd/embedded-cluster/list_images.go
sgalsaleh: can it be removed then? do we need it?

emosbaugh: removed

cmd/buildtools/utils.go
emosbaugh: done

cmd/buildtools/metadata.go
emosbaugh: fixed to use the metadata

pkg/addons/seaweedfs/seaweedfs.go
```diff
@@ -93,6 +93,14 @@ func (o *SeaweedFS) GenerateHelmConfig(onlyDefaults bool) ([]eckinds.Chart, []ec
 	return []eckinds.Chart{chartConfig}, nil, nil
 }
 
+func (a *SeaweedFS) GetImages() []string {
+	var images []string
+	for image, tag := range Metadata.Images {
+		images = append(images, fmt.Sprintf("proxy.replicated.com/anonymous/%s:%s", image, tag))
```
sgalsaleh, line 99: this will need to use the `AddonImageFromComponentName ` function now that we build seaweedfs with Chainguard

pkg/addons/embeddedclusteroperator/embeddedclusteroperator.go
```diff
@@ -40,7 +40,11 @@ var (
 	rawmetadata []byte
 	// Metadata is the unmarshal version of rawmetadata.
 	Metadata release.AddonMetadata
-	// Overwritten by -ldflags in Makefile
+)
+
+// Overwritten by -ldflags in Makefile
+var (
+	UtilsImage                    = "busybox:latest"
```
sgalsaleh, line 47: why does this sometimes come from the makefile and sometimes from the metadata.yaml file?

emosbaugh: Im just trying to unrevert this. ultimately we should move this to a makefile but i will follow up with that.

emosbaugh: i will add a metadata in a follow up when i add image gcr.io/kubebuilder/kube-rbac-proxy

sgalsaleh (approved):

PR #836, opened 2024-07-25
chore: add an e2e test with 5 airgapped nodes

ricardomaraschini:
add an e2e test with 5 controller nodes in an airgap environment.

### DIFF
--- a/e2e/install_test.go
+++ b/e2e/install_test.go
@@ -1871,6 +1871,13 @@ func maybeDownloadAirgapBundle(t *testing.T, versionLabel string, destPath strin
 	return airgapBundlePath, size
 }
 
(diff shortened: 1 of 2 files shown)


e2e/install_test.go
```diff
@@ -1973,3 +1980,130 @@ func cleanupCluster(t *testing.T, tc *cluster.Output) {
 	generateAndCopySupportBundle(t, tc)
 	copyPlaywrightReport(t, tc)
 }
+
+func TestMultiNodeAirgapUpgradeLarge(t *testing.T) {
```
laverya, line 1984: Any chance this could be named something like "TestFiveNodeAirgapUpgrade" instead?

ricardomaraschini: Done.

laverya (approved):

PR #856, opened 2024-07-26
Ensure k8s version matches the target version after upgrades

laverya:


laverya:
closed by https://github.com/replicatedhq/embedded-cluster/pull/894

PR #898, opened 2024-07-31
chore(ci): add pre-upgrade k0s verison check to e2e tests

emosbaugh:
Checks k8s version is as expected after installation and prior to upgrade

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -105,6 +105,9 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
     needs: [git-sha, build-images]
+    outputs:
+      k0s_version: ${{ steps.export.outputs.k0s_version }}
+      k0s_version_previous: ${{ steps.export.outputs.k0s_version_previous }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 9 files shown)


e2e/utils.go
```diff
@@ -162,9 +162,18 @@ func injectString(original, injection, after string) string {
 
 func k8sVersion() string {
 	// split the version string (like 'v1.29.6+k0s.0') into the k8s version and the k0s revision
-	verParts := strings.Split(os.Getenv("K0S_VERSION"), "+")
+	verParts := strings.Split(os.Getenv("EXPECT_K0S_VERSION"), "+")
```
emosbaugh, line 165: this wasnt actually working since actions have a limit of 10 inputs. it was just using the default from the makefile. i've changed this variable name to avoid the collision.

.github/workflows/ci.yaml
```diff
@@ -380,7 +388,8 @@ jobs:
           license: ${{ secrets.STAGING_EMBEDDED_CLUSTER_LICENSE }}
           dr-aws-access-key-id: ${{ secrets.TESTIM_AWS_ACCESS_KEY_ID }}
           dr-aws-secret-access-key: ${{ secrets.TESTIM_AWS_SECRET_ACCESS_KEY }}
-          k0s-version: ${{ steps.build.outputs.k0s_version }}
```
emosbaugh, line 383: needs

e2e/utils.go
laverya: That would do it

laverya (approved):

sgalsaleh (approved):

PR #899, opened 2024-07-31
fix: use k0s kubectl

emosbaugh:
Symlinks `kubectl` to `k0s`. This works because of https://github.com/k0sproject/k0s/blob/5d48d20767851fe8e299aacd3d5aae6fcfbeab37/main.go#L40

Fixes auto-completion for `kubectl` command alias. Only supports bash.


cmd/embedded-cluster/shell.go
```diff
@@ -99,7 +99,12 @@ var shellCommand = &cli.Command{
 
 		// if /etc/bash_completion is present enable kubectl auto completion.
 		if _, err := os.Stat("/etc/bash_completion"); err == nil {
-			config = fmt.Sprintf("source <(kubectl completion %s)\n", filepath.Base(shpath))
+			config = fmt.Sprintf("source <(k0s completion %s)\n", filepath.Base(shpath))
+			_, _ = shellpty.WriteString(config)
+			_, _ = io.CopyN(io.Discard, shellpty, int64(len(config)+1))
+
+			comppath := defaults.PathToEmbeddedClusterBinary("kubectl_completion_bash.sh")
```
emosbaugh, line 106: This only supports bash for now but I dont believe we supported anything else since we check for `/etc/bash_completion`

e2e/local-artifact-mirror_test.go
```diff
@@ -34,9 +34,9 @@ func TestLocalArtifactMirror(t *testing.T) {
 		{"systemctl", "stop", "local-artifact-mirror"},
 		{"systemctl", "start", "local-artifact-mirror"},
 		{"systemctl", "status", "local-artifact-mirror"},
-		{"curl", "-o", "/tmp/kubectl-test", "127.0.0.1:50000/bin/kubectl"},
-		{"chmod", "755", "/tmp/kubectl-test"},
-		{"/tmp/kubectl-test", "version", "--client"},
+		{"curl", "-o", "/tmp/kubectl", "127.0.0.1:50000/bin/kubectl"},
```
emosbaugh, line 37: Since kubectl is now a symlink `http.FileServer` returns the linked file with is k0s. Everything will still work it just happens that it is the actual k0s binary rather than a symlink.

Another option is to change kubectl to a script that execs the k0s binary.

laverya (approved):

laverya (approved):

PR #964, opened 2024-08-13
fix(reset): do not drain node if only one in the cluster

emosbaugh:
Draining and leaving etcd cluster can cause issues for reset of a single node and is not needed.

### DIFF
--- a/cmd/embedded-cluster/uninstall.go
+++ b/cmd/embedded-cluster/uninstall.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"regexp"
 	"time"
 
 	autopilot "github.com/k0sproject/k0s/pkg/apis/autopilot/v1beta2"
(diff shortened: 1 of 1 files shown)


cmd/embedded-cluster/uninstall.go
```diff
@@ -379,39 +400,43 @@ var resetCommand = &cli.Command{
...
-		if currentHost.Status.Role == "controller" {
-
-			// delete controlNode object from cluster
-			deleteControlCtx, deleteCancel := context.WithTimeout(c.Context, time.Minute)
-			defer deleteCancel()
-			err := currentHost.deleteControlNode(deleteControlCtx)
+		numControllerNodes, _ := kubeutils.NumOfControlPlaneNodes(c.Context, currentHost.Kclient)
+		// do not drain node if this is the only controller node in the cluster
+		// if there is an error (numControllerNodes == 0), drain anyway to be safe
+		if currentHost.Status.Role != "controller" || numControllerNodes != 1 {
```
emosbaugh, line 406: This is a really bad diff. I didnt change anything inside this conditional. I only wrapped the code in it.

laverya: looks much better when ignoring whitespace

laverya (approved):

PR #968, opened 2024-08-13
feat(troubleshoot): add connectivity preflights for replicated.app and proxy.replicated.com

emosbaugh:


pkg/preflights/host-preflight.yaml
```diff
@@ -291,3 +303,29 @@ spec:
               message: Kernel meets minimum version version of 3.10
           - fail:
               message: Kernel doesn't meet minimum version of 3.10
+    - http:
+        checkName: Can Access API
```
ajp-io, line 307: The way these print out during a failure, this name will make more sense
```suggestion
        checkName: API Access
```

pkg/preflights/host-preflight.yaml
```diff
@@ -291,3 +303,29 @@ spec:
...
+          - fail:
+              when: error
+              message: 'Error connecting to {{ .ReplicatedAPIURL }}'
+          - pass:
+              when: 'statusCode == 200'
+              message: 'Connected to {{ .ReplicatedAPIURL }}'
+          - fail:
+              message: 'Unexpected response from {{ .ReplicatedAPIURL }}'
+    - http:
+        checkName: Can Access Proxy Registry
```
ajp-io, line 320:
```suggestion
        checkName: Proxy Registry Access
```

pkg/preflights/host-preflight.yaml
```diff
@@ -291,3 +303,29 @@ spec:
...
           - fail:
               message: Kernel doesn't meet minimum version of 3.10
+    - http:
+        checkName: Can Access API
+        collectorName: http-replicated-app
+        exclude: '{{ or .IsAirgap (eq .ReplicatedAPIURL "") }}'
+        outcomes:
+          - fail:
+              when: error
+              message: 'Error connecting to {{ .ReplicatedAPIURL }}'
```
ajp-io, line 313: One goal of these is to highlight missing proxy configuration or a misconfigured firewall. I don't know how this will look UX-wise because it's getting long, but could we do something like this? Same for proxy service URL below.
```suggestion
              message: 'Error connecting to {{ .ReplicatedAPIURL }}. Ensure your firewall is properly configured, or use the `--http-proxy`, `--https-proxy`, and `--no-proxy` flags if there is a proxy server.'
```

pkg/preflights/host-preflight.yaml
ajp-io: Think this one got missed. I'll just commit it.

pkg/preflights/embed.go
```diff
@@ -3,20 +3,29 @@ package preflights
...
 	"github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
 	"github.com/replicatedhq/troubleshoot/pkg/loader"
 )
 
-//go:embed host-preflight.yaml
-var clusterHostPreflightYAML string
+var (
+	//go:embed host-preflight.yaml
+	clusterHostPreflightYAML string
+)
```
laverya, line 15: which one is it 😆 
(you changed from multiline to single line in pkg/defaults/defaults.go)

pkg/preflights/host-preflight.yaml
```diff
@@ -47,6 +47,18 @@ spec:
         command: 'sh'
         args: ['-c', 'command -v umount']
     - hostOS: {}
+    - http:
+        collectorName: http-replicated-app
```
laverya, line 51: I have a question here
does the preflights runner inherit proxy settings from the user's shell environment?

laverya (approved):

pkg/preflights/host-preflight.yaml
```diff
@@ -47,6 +47,18 @@ spec:
...
+        get:
+          url: '{{ .ReplicatedAPIURL }}'
+          timeout: 5s
+        exclude: '{{ or .IsAirgap (eq .ProxyRegistryURL "") }}'
```
sgalsaleh, line 55:
```suggestion
        exclude: '{{ or .IsAirgap (eq . ReplicatedAPIURL "") }}'
```

ajp-io: Nice catch!

pkg/preflights/host-preflight.yaml
emosbaugh: So i call setProxyEnv for each command that executes the preflights and they way exec.Command works it will inherit the environment from the parent golang process. I wonder if i should modify that function to set the env vars to empty strings if nil or unset?

https://github.com/replicatedhq/embedded-cluster/pull/968/files#diff-0d2512143545f44ccab24560b90d4413dc15a3dd9e0d1723dc841ce2e94b3554R67-R80

laverya: setting the env vars to empty strings if no proxy flags are passed makes perfect sense to me

laverya (approved):

sgalsaleh (approved):

emosbaugh:
E2E test is failing because there is more than 1 update available now that other prs have created new releases. I will resolve this in a follow up.

```
 [1386](https://github.com/replicatedhq/embedded-cluster/actions/runs/10425473877/job/28898682590?pr=968#step:4:1398)            Error: locator.click: Error: strict mode violation: locator('.available-update-row').filter({ hasText: 'appver-dev-c334ef1-upgrade' }).getByRole('button', { name: 'Deploy', exact: true }) resolved to 2 elements:   
```

0e432f1f14ed57b241afa27083d0fe6e934c55f6

https://github.com/replicatedhq/embedded-cluster/actions/runs/10425473877/job/28898682590?pr=968

PR #981, opened 2024-08-14
fix: pass service node port range to installation for join

emosbaugh:
This sets the computed value of service-node-port-range in the installation object for later use in the join command from the defaults taking into account the unsupported overrides from both the spec and the install flag.

This fixes an issue that could cause node joins to use a different value for the service-node-port-range than on the original node.


cmd/embedded-cluster/restore.go
```diff
@@ -979,6 +960,11 @@ var restoreCommand = &cli.Command{
 			}
 		}
 
+		applier, err := getAddonsApplier(c, "")
```
sgalsaleh, line 963: nit: should this be moved to the `ecRestoreStateNew` case since it's only used there?

sgalsaleh (approved):

PR #997, opened 2024-08-16
Bug: Don't check absolute paths for mount binaries

hedge-sparrow:
Checking for absolute paths breaks these host preflights on systems that don't conform to Ubuntu's binary locations. 

i.e: Alpine and RHEL don't keep `mount` and `umount` in `/usr/bin`

### DIFF
--- a/pkg/preflights/host-preflight.yaml
+++ b/pkg/preflights/host-preflight.yaml
@@ -264,7 +264,7 @@ spec:
     - textAnalyze:
         checkName: "'mount' Command"
         fileName: host-collectors/run-host/check-mount.txt
-        regex: 'mount'
+        regex: '/mount$'
         outcomes:
           - pass:
               when: "true"
--- a/pkg/preflights/host-preflight.yaml
+++ b/pkg/preflights/host-preflight.yaml
@@ -275,7 +275,7 @@ spec:
     - textAnalyze:
         checkName: "'umount' Command"
         fileName: host-collectors/run-host/check-umount.txt
-        regex: 'umount'
+        regex: '/umount$'
         outcomes:
           - pass:
               when: "true"
--- a/pkg/preflights/host-preflight.yaml
+++ b/pkg/preflights/host-preflight.yaml
@@ -253,7 +253,7 @@ spec:
     - textAnalyze:
         checkName: "'modprobe' Command"
         fileName: host-collectors/run-host/check-modprobe.txt
-        regex: '/usr/sbin/modprobe'
+        regex: 'modprobe'
         outcomes:
           - pass:
               when: "true"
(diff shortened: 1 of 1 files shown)


ajp-io (dismissed):

ricardomaraschini (approved):

laverya (dismissed):

pkg/preflights/host-preflight.yaml
```diff
@@ -253,7 +253,7 @@ spec:
     - textAnalyze:
         checkName: "'modprobe' Command"
         fileName: host-collectors/run-host/check-modprobe.txt
-        regex: '/usr/sbin/modprobe'
+        regex: 'modprobe'
```
laverya, line 256: I would _prefer_ to have this be  `/modprobe$` instead

```suggestion
        regex: '/modprobe$'
```

pkg/preflights/host-preflight.yaml
```diff
@@ -264,7 +264,7 @@ spec:
     - textAnalyze:
         checkName: "'mount' Command"
         fileName: host-collectors/run-host/check-mount.txt
-        regex: '/usr/bin/mount'
+        regex: 'mount'
```
laverya, line 267:
```suggestion
        regex: '/mount$'
```

pkg/preflights/host-preflight.yaml
```diff
@@ -275,7 +275,7 @@ spec:
     - textAnalyze:
         checkName: "'umount' Command"
         fileName: host-collectors/run-host/check-umount.txt
-        regex: '/usr/bin/umount'
+        regex: 'umount'
```
laverya, line 278:
```suggestion
        regex: '/umount$'
```

laverya (commented):
These regexes could be better

hedge-sparrow:
@laverya the regex you suggested doesn't work when testing locally. I had initially tried `^.+/mount$` but I'm  not sure the regex collector actually supports this format

laverya (approved):

RangerRick:
If you want to be a bit more specific, you could do something like:

`^(?:/usr)?/s?bin/mount$`

(assuming you can do extended regex... if it's go like most of your stuff, they supposedly support non-capturing groups)

That would catch anything in `/bin`, `/sbin`, `/usr/bin`, and `/usr/sbin` and nowhere else.

hedge-sparrow:
Personally I don't think the analyzer _needs_ to be super specific, the collector will only output the path to the binary if it exists in `$PATH`, so matching on just the binary name will either be true (it's in `$PATH`) or false (it's not and the collector returns an empty string), I don't think it's possible to generate a false positive here

PR #1020, opened 2024-08-21
update tests for node join command UI

miaawong:
update playwright tests after node join UI improvements in KOTS


e2e/playwright/tests/get-join-worker-command/test.spec.ts
```diff
@@ -7,7 +7,11 @@ test('get join worker command', async ({ page }) => {
   await page.getByRole('button', { name: 'Add node', exact: true }).click();
   await expect(page.locator('.Modal-body')).toBeVisible();
   await expect(page.getByRole('heading')).toContainText('Add a Node');
-  await page.locator('.BoxedCheckbox').getByText('abc', { exact: true }).click();
+  await expect(page.getByText('Roles:')).toBeVisible();
+  await expect(page.locator('#controller-testNodeType')).toBeChecked();
+  await page.locator('.nodeType-selector').getByText('controller-test').click()
+  await page.locator('.nodeType-selector').getByText('abc').click()
+  await expect(page.locator('#abcNodeType')).toBeChecked();
```
sgalsaleh, line 14: how about also checking that `controller-testNodeType` isn't checked anymore?

e2e/playwright/tests/shared/deploy-app.ts
```diff
@@ -1,5 +1,7 @@
 export const deployApp = async (page, expect) => {
-  await expect(page.getByRole('button', { name: 'Add node', exact: true })).toBeVisible();
+  await expect(page.getByText('Optionally add nodes to the cluster'),).toBeVisible();
+  const joinCommand = await page.locator('.react-prism.language-bash').first().textContent();
+  console.log(`{"command":"${joinCommand}"}`);
```
sgalsaleh, line 4: we only log this so it's detected by our e2e tests. how about running one or two validations that the command is visible instead?

sgalsaleh (approved):

PR #1043, opened 2024-08-27
Set default value for pod/service CIDRs with flag

ajp-io:
This way the default value shows up in the CLI help menu, making it easier for users to know whether they need to update it.


ajp-io:
What the CLI help menu looks like now:
image

I don't know why there's a commands section just for a help command when there's already a help option. Other commands like reset don't do that. I might make another PR to change that.

PR #1050, opened 2024-08-28
feat: verify network system configuration during preflight

ricardomaraschini:

Make sure the user deliberately sets up a few `sysctl` parameters. K0s sets these automatically but we want to make users deliberately enable them anyways. The goal is to let users aware that these configurations are necessary. These parameters were copied from k0s [source code](https://github.com/k0sproject/k0s/blob/391cfd1b9fc1c6b6221398e5bcef4034ba248149/pkg/component/worker/kernelsetup_linux.go#L62).

https://app.shortcut.com/replicated/story/110749/preflight-for-ip-forwarding

TO BE IMPLEMENTED

```release-note
Start to verify more sysctl configuration prior to installation.
```

### DIFF
--- a/e2e/cluster/cluster.go
+++ b/e2e/cluster/cluster.go
@@ -33,6 +33,18 @@ echo "Internet connection is down"
 exit 1
 `
 
+// setSysctl is a script that sets the sysctl rules required by the embedded
+// cluster. this script implements the behaviour we expect the user to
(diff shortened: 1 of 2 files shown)


ricardomaraschini:
We have decided to postpone this work (see Shortcut). Closing this one.

PR #1124, opened 2024-09-09
remove extra 'force' flag in asset downloads

laverya:

```release-note
Update to Kots v1.116.0
```

### DIFF
--- a/scripts/cache-files.sh
+++ b/scripts/cache-files.sh
@@ -54,11 +54,11 @@ function k0sbin() {
     # if the override is set, we should download this binary and upload it to the bucket so as not to require end users hit the override url
     if [ -n "${k0s_override}" ] && [ "${k0s_override}" != '' ]; then
         echo "K0S_BINARY_SOURCE_OVERRIDE is set to '${k0s_override}', using that source"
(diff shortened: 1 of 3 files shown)


ricardomaraschini (approved):

scripts/cache-files.sh
```diff
@@ -54,11 +54,11 @@ function k0sbin() {
     # if the override is set, we should download this binary and upload it to the bucket so as not to require end users hit the override url
     if [ -n "${k0s_override}" ] && [ "${k0s_override}" != '' ]; then
         echo "K0S_BINARY_SOURCE_OVERRIDE is set to '${k0s_override}', using that source"
-        curl --fail-with-body --retry 5 --retry-all-errors -fL -o "${k0s_version}" "${k0s_override}"
+        curl --fail-with-body --retry 5 --retry-all-errors -L -o "${k0s_version}" "${k0s_override}"
```
emosbaugh, line 57: do you think you should remove fail-with-body instead? im not sure it makes sense with -o

laverya: I have no preference either way
yours worked and was merged so I'm happy to do that instead

PR #1127, opened 2024-09-09
feat: add --private-ca flag

ricardomaraschini:

add --private-ca flag and sets it up on the newest version of kots-helm.

```release-note

```


e2e/utils.go
```diff
@@ -32,6 +32,16 @@ func RequireEnvVars(t *testing.T, envVars []string) {
 
 type RunCommandOption func(cmd *cluster.Command)
 
+func WithECShelEnv() RunCommandOption {
```
sgalsaleh, line 35:
```suggestion
func WithECShellEnv() RunCommandOption {
```

pkg/addons/adminconsole/adminconsole.go
```diff
@@ -172,6 +173,10 @@ func (a *AdminConsole) Outro(ctx context.Context, cli client.Client, k0sCfg *k0s
 		return fmt.Errorf("unable to create kots password secret: %w", err)
 	}
 
+	if err := createKotsCAConfigmap(ctx, cli, a.namespace, a.privateCAs); err != nil {
```
sgalsaleh, line 176: what happens if kotsadm gets deployed by helm before this configmap is created? would the kotsadm pod detect the configmap when it gets created?

sgalsaleh (approved):

PR #1140, opened 2024-09-12
Improvements to dev environment

sgalsaleh:

Adds support for running the dev environment locally on MacOS.

Benefits:

- Much faster feedback loop
- Better caching
- Ability to integrate with and develop against the KOTS project

[SC-111503](https://app.shortcut.com/replicated/story/111503/improvements-to-kots-and-ec-dev-environment)

NONE

```release-note
NONE
```

NONE


pkg/addons/adminconsole/adminconsole.go
```diff
@@ -38,7 +38,7 @@ const (
 )
 
 var (
-	//go:embed static/values.yaml
+	//go:embed static/values.tpl.yaml
```
emosbaugh, line 41: is there some reason admin console doesnt have architecture images? i dont see any changes to the tpl file here. what am i missing?

scripts/ci-cache-files.sh
```diff
@@ -9,6 +9,7 @@ EC_VERSION=${EC_VERSION:-}
 K0S_VERSION=${K0S_VERSION:-}
 AWS_REGION="${AWS_REGION:-us-east-1}"
 S3_BUCKET="${S3_BUCKET:-dev-embedded-cluster-bin}"
+CACHE_BINS=${CACHE_BINS:-1}
```
emosbaugh, line 12: could you change this to UPLOAD_BINARIES perhaps in a follow up pr? really the file should be renamed as its not caching them. It is uploading them so that the release can be installed remotely.

emosbaugh (approved):
one question and one nit but feel free to merge

pkg/addons/adminconsole/adminconsole.go
sgalsaleh: it uses the `ImageString` template function which has been updated in this pr to account for arch.

scripts/ci-cache-files.sh
sgalsaleh: yes, i can update that in a separate pr

PR #1159, opened 2024-09-13
clarify that kots should be in the same parent directory, not a subdirectory

laverya:

```release-note

```


sgalsaleh (approved):

sgalsaleh:
Merging as this is only a README.md change

PR #1160, opened 2024-09-13
chore(ci): build and tag all image architectures

emosbaugh:

Will build for all architectures possible and name the tag with the arch.

```release-note

```


pkg/addons/registry/static/metadata.yaml
```diff
@@ -12,4 +12,4 @@ images:
         repo: proxy.replicated.com/anonymous/replicated/ec-registry
         tag:
             amd64: 2.8.3-r0@sha256:5b76ebd0a362009e31a05ac487c690f5ece0e11f6c4d9261ca63a3f162b57660
```
sgalsaleh, line 14: why wasn't this updated?

emosbaugh: this doesnt work for `replicated/ec-registry` image because its not multi-arch. we need to re-push it

sgalsaleh: how did it work for the arm image then?

pkg/addons/registry/static/metadata.yaml
```diff
@@ -12,4 +12,4 @@ images:
...
-            arm64: 2.8.3-r0@sha256:b6b7345a1455ae1decd79f39e9ee9f8dff4dedf9fc52acebeba0903779c37cc3
+            arm64: 2.8.3-r0-arm64@sha256:e9841478ded46f9ed3f258ea7b291718954cecce2d6907df778d223dded2dd9d
```
sgalsaleh, line 15: hmm, the arm image isn't multi-arch, why did the digest change here?

sgalsaleh (approved):

pkg/addons/registry/static/metadata.yaml
emosbaugh: Screenshot 2024-09-17 at 1 16 22 PM

looks like the tag points at e9

pkg/addons/registry/static/metadata.yaml
emosbaugh: tag 2.8.3-r0 is an arm image

pkg/addons/registry/static/metadata.yaml
sgalsaleh: sounds good

emosbaugh:
Failing test is a flake. Merging

PR #1195, opened 2024-09-17
feat(preflights): include preflight collect results in support bundle

emosbaugh:

The new version of troubleshoot will drop the preflight collected files on disk in the current directory. This change will move the file to the /var/lib/embedded-cluster/logs directory which is included in support bundles.

```release-note

```


pkg/preflights/preflights.go
```diff
@@ -61,6 +63,28 @@ func Run(ctx context.Context, spec *troubleshootv1beta2.HostPreflightSpec, proxy
...
+		return nil
+	}
+	// get the newest bundle
+	src := matches[0]
+	for _, match := range matches {
+		if filepath.Base(match) > filepath.Base(src) {
+			src = match
+		}
+	}
+	dst := filepath.Join(defaults.EmbeddedClusterLogsSubDir(), "preflight-bundle.tar.gz")
```
laverya, line 81: this shouldn't be unique - I think for logs we use the timestamp or something?

That way we don't overwrite each time the install command is run

banjoh (approved):

pkg/preflights/preflights.go
emosbaugh: My thinking here is that if we store all these bundles here it will bloat the support bundle too much so i am only storing the latest.

laverya (approved):

emosbaugh:
Test is flakey. Merging

PR #1201, opened 2024-09-18
feat: ability to override admin console and lam host ports

emosbaugh:

- Adds the ability to override the Admin Console port (default 30000) on installation using the flag `--admin-console-port`
- Adds the ability to override the Local Artifact Mirror port (default 50000) on installation using the flag `--local-artifact-mirror-port`
- Adds the ability to override the Local Artifact Mirror port (default 50000) on restore using the flag `--local-artifact-mirror-port`

```release-note
- Adds the ability to override the Admin Console port (default 30000) on installation using the flag `--admin-console-port`
- Adds the ability to override the Local Artifact Mirror port (default 50000) on installation using the flag `--local-artifact-mirror-port`
- Adds the ability to override the Local Artifact Mirror port (default 50000) on restore using the flag `--local-artifact-mirror-port`
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ include common.mk
 
 APP_NAME = embedded-cluster
 ADMIN_CONSOLE_CHART_REPO_OVERRIDE =
-ADMIN_CONSOLE_IMAGE_OVERRIDE =
+ADMIN_CONSOLE_IMAGE_OVERRIDE = ttl.sh/ethan/kotsadm:24h
 ADMIN_CONSOLE_MIGRATIONS_IMAGE_OVERRIDE =
 ADMIN_CONSOLE_KURL_PROXY_IMAGE_OVERRIDE =
 K0S_VERSION = v1.29.8+k0s.0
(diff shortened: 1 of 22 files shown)


cmd/embedded-cluster/flags.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("invalid admin console port: %w", err)
+	}
+	return port, nil
+}
+
+func getLocalArtifactMirrorPortFlag() cli.Flag {
+	return &cli.StringFlag{
+		Name:   "local-artifact-mirror-port",
+		Usage:  "Port on which the Local Artifact Mirror will be served",
```
ajp-io, line 37: I was considering whether it was worth mentioning the name LAM or not, because an end user will have no idea what that is and doesn't need to. But I don't know what else we'd call it other than `port-2`. So maybe it's fine.

I don't have a strong preference, but we could consider `lam-port` if we feel like `local-artifact-mirror-port` is too long, or use it as an alias. People don't type this much, so I don't care much.

cmd/embedded-cluster/flags.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/replicatedhq/embedded-cluster/pkg/defaults"
+	"github.com/urfave/cli/v2"
+	k8snet "k8s.io/utils/net"
+)
+
+func getAdminColsolePortFlag() cli.Flag {
+	return &cli.StringFlag{
+		Name:   "admin-console-port",
+		Usage:  "Port on which the Admin Console will be served",
+		Value:  strconv.Itoa(defaults.AdminConsolePort),
```
ajp-io, line 16: Nice. Came here looking for this because I like the default value to show in the help menu so users can more easily determine if they need to set it to something else.

ajp-io: Is there a range of ports that that are valid that we should validate this against? Maybe 80-32767 or whatever the node port range is set to?

ajp-io: Oh I see your TODO down lower

Makefile
```diff
@@ -282,7 +283,7 @@ create-node%:
 		-v /var/lib/k0s \
 		-v $(shell pwd):/replicatedhq/embedded-cluster \
 		-v $(shell dirname $(shell pwd))/kots:/replicatedhq/kots \
-		$(if $(filter node0,node$*),-p 30000:30000) \
+		$(if $(filter node0,node$*),-p $(NODE_PORT):$(NODE_PORT)) \
```
sgalsaleh, line 286: Shouldn't the container port be 30000? Otherwise you'd have to pass the admin console port flag right?

cmd/embedded-cluster/restore.go
```diff
@@ -906,6 +916,12 @@ var restoreCommand = &cli.Command{
 				Value:  false,
 				Hidden: true,
 			},
+			&cli.StringFlag{
+				Name:  "local-artifact-mirror-port",
+				Usage: "Port on which the Local Artifact Mirror will be served. If left empty, the port will be retrieved from the snapshot.",
+				// DefaultText: strconv.Itoa(defaults.LocalArtifactMirrorPort),
```
sgalsaleh, line 922: is this intentional?

sgalsaleh (approved):

Makefile
emosbaugh: yes this is intentional

cmd/embedded-cluster/restore.go
emosbaugh: yes im not sure what is right here

PR #1204, opened 2024-09-19
feat: ability to specify network interface

sgalsaleh:

Adds the ability to choose the network interface to use for the cluster node(s).

```release-note
* Adds the ability to choose the network interface to use for the cluster node(s).
```

TBD


cmd/embedded-cluster/join.go
```diff
@@ -190,7 +190,7 @@ var joinCommand = &cli.Command{
 		}
 
 		setProxyEnv(jcmd.Proxy)
-		proxyOK, localIP, err := checkProxyConfigForLocalIP(jcmd.Proxy)
+		proxyOK, localIP, err := checkProxyConfigForLocalIP(jcmd.Proxy, "") // TODO (@salah): pass network interface
```
laverya, line 193: still a todo here

sgalsaleh: yeah, we decided to ship the feature for installs / restores and add it later for joins (multi nodes).

pkg/addons/applier.go
```diff
@@ -52,8 +52,16 @@ type Applier struct {
 	privateCAs      map[string]string
 }
 
+type OutroOptions struct {
+	Ctx              context.Context
```
laverya, line 56: ctx shouldn't be in the options struct from a general code standards department
moving the rest to one potentially makes sense but also makes it dramatically harder to ensure you've updated every caller

pkg/netutils/ips.go
```diff
@@ -4,51 +4,91 @@ import (
...
-		if isPodnetworkInterface(i.Name) {
-			continue
+		if i.Name == iface {
+			return firstValidIPNet(i)
 		}
-		addresses, err := i.Addrs()
-		if err != nil {
-			logrus.Debugf("failed to get addresses for interface %s: %s", i.Name, err.Error())
+	}
+	return nil, fmt.Errorf("interface %s not found or is not valid", iface)
```
laverya, line 39:
```suggestion
	ifaceNames := []string{}
	for _, i := range ifs {
		ifaceNames = append(ifaceNames, i.Name)
	}
	return nil, fmt.Errorf("interface %s not found or is not valid. The following interfaces were detected %v", iface, ifaceNames)
```

this should probably be integrated with the previous loop but github wouldn't let me suggest that

pkg/netutils/ips.go
```diff
@@ -4,51 +4,91 @@ import (
...
-
-	"github.com/sirupsen/logrus"
 )
 
-// GetDefaultIPNet returns the default interface for the node, and the subnet mask for that node, using the same logic
-// as k0s in https://github.com/k0sproject/k0s/blob/v1.30.4%2Bk0s.0/internal/pkg/iface/iface.go#L61
-func GetDefaultIPNet() (*net.IPNet, error) {
-	ifs, err := net.Interfaces()
+// FirstValidAddress returns the first found non-local IPv4 address
+// that's not part of pod network for the given interface.
```
laverya, line 10: This function doesn't care about pod networking at all, that's an interface name filter and this takes an interface name

sgalsaleh: it filters interfaces based on whether or not they have valid ip nets which takes pod networking into account though

laverya (approved):

sgalsaleh:
Bypassing the `TestSingleNodeAirgapUpgradeFromEC18` test failure as it's expected to be flakey until the [upstream k0s fix](https://github.com/k0sproject/k0s/pull/4866) is released.

PR #1205, opened 2024-09-19
feat: add host preflights for all needed ports

ricardomaraschini:

Checks if all needed ports are available on the server prior to start the installation.

### DIFF
--- a/cmd/embedded-cluster/reset.go
+++ b/cmd/embedded-cluster/reset.go
@@ -365,11 +365,6 @@ var resetCommand = &cli.Command{
 			Usage: "Disable interactive prompts",
 			Value: false,
 		},
-		&cli.BoolFlag{
-			Name:  "reboot",
-			Usage: "Reboot system after resetting the node",
-			Value: false,
-		},
 	},
 	Usage: fmt.Sprintf("Remove %s from the current node", binName),
 	Action: func(c *cli.Context) error {
@@ -378,7 +373,7 @@ var resetCommand = &cli.Command{
 		}
 
 		logrus.Info("This will remove this node from the cluster and completely reset it, removing all data stored on the node.")
-		logrus.Info("Do not reset another node until this is complete.")
+		logrus.Info("This action will cause the node to reboot. Do not reset another node until this is complete.")
 		if !c.Bool("force") && !c.Bool("no-prompt") && !prompts.New().Confirm("Do you want to continue?", false) {
 			return fmt.Errorf("Aborting")
 		}
@@ -449,10 +444,6 @@ var resetCommand = &cli.Command{
 			return err
 		}
 
-		if !c.Bool("reboot") {
-			logrus.Infof("Node has been reset. Please reboot to ensure transient configuration is also reset.")
-		}
-
 		if err := helpers.RemoveAll(defaults.PathToK0sConfig()); err != nil {
 			return fmt.Errorf("failed to remove k0s config: %w", err)
 		}
@@ -501,10 +492,8 @@ var resetCommand = &cli.Command{
 			return fmt.Errorf("failed to remove k0s binary: %w", err)
 		}
 
-		if c.Bool("reboot") {
-			if _, err := exec.Command("reboot").Output(); err != nil {
-				return err
-			}
+		if _, err := exec.Command("reboot").Output(); err != nil {
+			return err
 		}
 
 		return nil
--- a/pkg/preflights/host-preflight.yaml
+++ b/pkg/preflights/host-preflight.yaml
@@ -71,6 +71,60 @@ spec:
         operationSize: 2300
         datasync: true
         runTime: "0" # let it run to completion
+    - tcpPortStatus:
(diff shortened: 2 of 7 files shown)


pkg/preflights/host-preflight.yaml
```diff
@@ -370,6 +424,342 @@ spec:
...
+              message: Port 7443/TCP is available
+          - error:
+              message: Port 7443/TCP is required, but an unexpected error occurred when trying to connect to it
+    - tcpPortStatus:
+        checkName: Kotsadm Node Port Availability
+        collectorName: Kotsadm Node Port
+        outcomes:
+          - fail:
+              when: "connection-refused"
+              message: Port 30000/TCP is required, but the connection was refused. This may indicate that the connection is being redirected to somewhere else on your network (perhaps a proxy configuration).
```
ajp-io, line 562: I need to better understand this one to comment on the wording, but this can be fine for now.

pkg/preflights/host-preflight.yaml
```diff
@@ -370,6 +424,342 @@ spec:
...
+          - fail:
+              when: "address-in-use"
+              message: Port 30000/TCP is required, but another process is already listening on it. Relocate the conflicting process to continue.
```
ajp-io, line 565: When we merge Ethan's port relocation changes, we can update this:
```suggestion
              message: Port 30000/TCP is required, but another process is already listening on it. To proceed, relocate the conflicting process or specify an available port using --admin-console-port.
```

pkg/preflights/host-preflight.yaml
```diff
@@ -370,6 +424,342 @@ spec:
...
+    - tcpPortStatus:
+        checkName: Local Artifact Mirror Port Availability
+        collectorName: Local Artifact Mirror Port
+        outcomes:
+          - fail:
+              when: "connection-refused"
+              message: Port 50000/TCP is required, but the connection was refused. This may indicate that the connection is being redirected to somewhere else on your network (perhaps a proxy configuration).
+          - fail:
+              when: "address-in-use"
+              message: Port 50000/TCP is required, but another process is already listening on it. Relocate the conflicting process to continue.
```
ajp-io, line 481: When we merge Ethan's port relocation changes, we can update this:
```suggestion
              message: Port 50000/TCP is required, but another process is already listening on it. To proceed, relocate the conflicting process or specify an available port using --local-artifact-mirror-port.
```

pkg/preflights/host-preflight.yaml
ricardomaraschini: For context: this means that we have been able to start listening on a given port (the port is available) but when we tried to connect to it we failed. This is an IMHO unlikely scenario.

We have discussed this personally and we came to the conclusion that this could be caused by an erroneous proxy configuration but I am not sure this is even possible as we are testing at TCP level, not HTTP.  Some iptables rules may cause this though (redirecting traffic somewhere else) but again, unlikely.

cmd/embedded-cluster/reset.go
```diff
@@ -365,11 +365,6 @@ var resetCommand = &cli.Command{
 			Usage: "Disable interactive prompts",
 			Value: false,
 		},
-		&cli.BoolFlag{
```
emosbaugh, line 368: is this only for testing purposes? or is this a proposed change? im not sure this is desirable

emosbaugh: Got it. I missed that. Makes sense now

pkg/preflights/host-preflight.yaml
```diff
@@ -381,3 +435,339 @@ spec:
               message: 'P99 write latency for the disk at /var/lib/k0s/etcd is {{ "{{" }} .P99 {{ "}}" }}, which is better than the 10 ms requirement.'
           - fail:
               message: 'P99 write latency for the disk at /var/lib/k0s/etcd is {{ "{{" }} .P99 {{ "}}" }}, but it must be less than 10 ms. A higher-performance disk is required.'
+    - tcpPortStatus:
+        checkName: ETCD Internal Port Availability
+        collectorName: ETCD Internal Port
+        outcomes:
+          - fail:
+              when: "connection-refused"
+              message: Connection to 2379/TCP was refused. This may indicate that the connection is being redirected to somewhere else on your network (perhaps a proxy configuration).
```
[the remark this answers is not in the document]
ajp-io: Yeah, we do that ourselves now. Maybe a more generic message like "ensure this port is available" would suffice if it's very unlikely this even occurs.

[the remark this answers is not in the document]
ricardomaraschini: After looking at this again I don't think even an invalid `NO_PROXY` configuration would cause this to happen. The collector operates at the TCP level. Are we OK with the "Please ensure this port is available" message ?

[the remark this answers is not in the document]
ajp-io: Yeah, I can make a change

laverya (approved):

PR #1222, opened 2024-09-24
improve our upgrade architecture

laverya:

This creates a job (using the operator image) to trigger a k0s upgrade, instead of updating an installation object and using the currently installed operator to trigger the upgrade.

This allows us to use the to-be-upgraded-to operator to perform the upgrade, allowing us to upgrade to 1.30.

```release-note

```


laverya:
It looks like the previous-k0s build isn't actually being built with the right k0s import - still working on figuring out why

operator/pkg/k8sutil/clusterVersion.go
```diff
@@ -0,0 +1,34 @@
...
+	"context"
+	"fmt"
+	"github.com/replicatedhq/embedded-cluster/kinds/types"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
+)
+
+// K0sVersionFromMetadata takes versions like v1.30.5+k0s.0 and returns v1.30.5+k0s to match the kubeletVersion in a cluster
+func K0sVersionFromMetadata(meta *types.ReleaseMetadata) string {
```
laverya, line 13: I need to add a unit test for this

operator/pkg/k8sutil/clusterVersion.go
```diff
@@ -0,0 +1,34 @@
...
+	if meta == nil || meta.Versions == nil {
+		return ""
+	}
+	desiredParts := strings.Split(meta.Versions["Kubernetes"], "k0s")
+	desiredVersion := desiredParts[0] + "k0s"
+	return desiredVersion
+}
+
+// ClusterNodesMatchVersion returns true if all nodes in the cluster have kubeletVersion matching the provided version.
+func ClusterNodesMatchVersion(ctx context.Context, cli client.Client, version string) (bool, error) {
```
laverya, line 23: and for this

operator/pkg/upgrade/upgrade.go
```diff
@@ -2,411 +2,248 @@ package upgrade
...
+		return fmt.Errorf("failed to delete successful upgrade plan: %w", err)
 	}
-	return clusterConfig, nil
+	return nil
 }
 
-func getOperatorChart(ctx context.Context, cli client.Client, in *clusterv1beta1.Installation) (*k0sv1beta1.Chart, error) {
-	metadata, err := release.MetadataFor(ctx, in, cli)
+// copied from ReconcileHelmCharts in https://github.com/replicatedhq/embedded-cluster/blob/c6a57a4/operator/controllers/installation_controller.go#L568
+func chartUpgrade(ctx context.Context, cli client.Client, in *clusterv1beta1.Installation) error {
```
laverya, line 135: this function should be deduplicated somehow

Makefile
```diff
@@ -176,10 +176,12 @@ upgrade-release: EC_VERSION = $(VERSION)-$(CURRENT_USER)-upgrade
 upgrade-release: APP_VERSION = appver-dev-$(shell LC_ALL=C tr -dc 'A-Za-z0-9' < /dev/urandom | head -c6)-upgrade
 upgrade-release: check-env-EC_VERSION check-env-APP_VERSION
 	UPLOAD_BINARIES=1 \
+	RELEASE_YAML_DIR=e2e/kots-release-upgrade \
 		./scripts/build-and-release.sh
 
 .PHONY: go.mod
 go.mod: Makefile
+	echo "go get github.com/k0sproject/k0s@$(K0S_GO_VERSION)"
```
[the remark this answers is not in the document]
laverya: this was probably leftover from debugging, will remove

operator/pkg/upgrade/upgrade.go
```diff
@@ -2,411 +2,248 @@ package upgrade
...
-}
-
-func ensureAirgapArtifactsInCluster(ctx context.Context, cli client.Client, in *clusterv1beta1.Installation) error {
-	log := ctrl.LoggerFrom(ctx)
 
-	log.Info("Uploading container images...")
+}
 
-	err := autopilotEnsureAirgapArtifactsPlan(ctx, cli, in)
+func unLockInstallation(ctx context.Context, cli client.Client, in *v1beta1.Installation) error {
```
emosbaugh, line 216: can you explain to me and comment what this does? why do you copy the new spec into the old installation? i thought we create new installation crs for upgrades?

laverya: yeah this definitely needs a comment then
this isn't the 'old' installation, this is the installation created at the beginning of the upgrade process - it's what we're upgrading to
but because that was created before the CRD was updated, it might not have all the fields, and so updating it with the from-configmap version is needed

operator/pkg/upgrade/job.go
```diff
@@ -0,0 +1,400 @@
...
+		err = airgapDistributeArtifacts(ctx, cli, in, localArtifactMirrorImage)
+		if err != nil {
+			return fmt.Errorf("airgap distribute artifacts: %w", err)
+		}
+
+		pullPolicy = corev1.PullNever
+	}
+
+	// create the installation object so that kotsadm can immediately find it and watch it for the upgrade process
+	err = createInstallation(ctx, cli, in)
```
emosbaugh, line 65: should you rename this to createInstallationIfNotExists or something?

operator/pkg/upgrade/job.go
```diff
@@ -0,0 +1,400 @@
...
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "embedded-cluster",
+			Name:      fmt.Sprintf(upgradeJobConfigMap, in.Name),
+		},
+		Data: map[string]string{
+			"installation.yaml": string(installationData),
+		},
+	}
+	if err = cli.Create(ctx, cm); err != nil {
```
emosbaugh, line 85: do you need a similar function, createJobConfigMapIfNotExists?

operator/pkg/upgrade/upgrade.go
emosbaugh: Actually I dont follow. The installation used to be created here

https://github.com/replicatedhq/embedded-cluster/pull/1222/files#diff-1595f9a51283933adb1a95f0d4a42f0576e783a1932c0d0da27fa1e5ca5628b3L69

```
	err = createInstallation(ctx, cli, in)
	if err != nil {
		return fmt.Errorf("apply installation: %w", err)
	}
```

operator/pkg/upgrade/job.go
emosbaugh: ok, i understand the flow now. Does this really belong in this function?

operator/pkg/charts/reconcile.go
```diff
@@ -0,0 +1,134 @@
...
+	v1beta3 "github.com/k0sproject/k0s/pkg/apis/helm/v1beta1"
+	v1beta2 "github.com/k0sproject/k0s/pkg/apis/k0s/v1beta1"
+	"github.com/replicatedhq/embedded-cluster/kinds/apis/v1beta1"
+	"github.com/replicatedhq/embedded-cluster/operator/pkg/release"
+	"sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ReconcileHelmCharts reconciles the helm charts from the Installation metadata with the clusterconfig object.
+func ReconcileHelmCharts(ctx context.Context, cli client.Client, in *v1beta1.Installation) error {
```
laverya, line 17: this is unmodified in terms of logic, just moved and changed to take a client as an argument

operator/pkg/upgrade/job.go
```diff
@@ -0,0 +1,414 @@
...
+		return "", fmt.Errorf("failed to get release metadata: %w", err)
+	}
+	operatorImage := ""
+	for _, image := range meta.Images {
+		if strings.Contains(image, "embedded-cluster-operator-image") {
+			operatorImage = image
+			break
+		}
+	}
+	return operatorImage, nil
```
sgalsaleh, line 219: should this return an error if the operator image is not found?

operator/pkg/upgrade/job.go
```diff
@@ -0,0 +1,414 @@
...
+
+	// find the first installation with a different name than the one we're upgrading to
+	for _, installation := range installations.Items {
+		if installation.Name != in.Name {
+			return &installation, nil
+		}
+	}
+
+	// if we get here, we didn't find a previous installation
+	return nil, fmt.Errorf("installation not found")
```
sgalsaleh, line 413:
```suggestion
	return nil, fmt.Errorf("previous installation not found")
```

pkg/kubeutils/kubeutils.go
```diff
@@ -318,6 +318,29 @@ func WaitForControllerNode(ctx context.Context, kcli client.Client, name string)
 	return nil
 }
 
+// WaitForJob waits for a job to be ready.
```
sgalsaleh, line 321:
```suggestion
// WaitForJob waits for a job to complete.
```

pkg/kubeutils/kubeutils.go
```diff
@@ -318,6 +318,29 @@ func WaitForControllerNode(ctx context.Context, kcli client.Client, name string)
...
+func WaitForJob(ctx context.Context, cli client.Client, ns, name string, maxSteps int, completions int32) error {
```
sgalsaleh, line 322: i might be missing it, but where is this used?

operator/pkg/upgrade/upgrade.go
```diff
@@ -2,411 +2,157 @@ package upgrade
...
+		if command.K0sUpdate != nil {
+			isK0sUpgrade = true
+			break
+		}
 	}
-
-	for _, chart := range cfgs.Charts {
-		if chart.Name == operatorChartName {
-			return &chart, nil
+	// if this was not a k0s upgrade plan, we can just delete the plan and restart the function to get a k0s upgrade
```
sgalsaleh, line 106: should this be done before checking if the plan has ended because it might've been a non-k0s upgrade plan?

pkg/kubeutils/kubeutils.go
laverya: It was initially added when I was having the binary wait for the job to complete
it's not used now

operator/pkg/upgrade/upgrade.go
laverya: I think that even if the plan that's in progress isn't a k0s upgrade plan we should still wait for it to complete
similarly if it failed we should probably avoid messing things up further

sgalsaleh (approved):

PR #1223, opened 2024-09-25
chore: eco is built at release time and is no longer needed in auto updates

emosbaugh:

We now build ECO at time of release and no longer need to update it periodically.

```release-note

```


ricardomaraschini:
Do we still need this ? https://github.com/replicatedhq/embedded-cluster/blob/main/cmd/buildtools/embeddedclusteroperator.go

PR #1238, opened 2024-09-26
feat: consolidate data directories for ec, k0s, and openebs

emosbaugh:

This change consolidates the k0s and openebs data directories under the /var/lib/embedded-cluster directory and adds the ability to configure the root directory with the flag `--data-dir`.

```release-note
- Adds the ability to configure the data directory with the flag `--data-dir` to the install command (default /var/lib/embedded-cluster).
- Data directories are now consolidated under /var/lib/embedded-cluster. Previously the k0s and openebs data was stored in /var/lib/k0s and /var/openebs respectively.
- Logs are now stored in /var/log/embedded-cluster rather than /var/lib/embedded-cluster/log prior to this change.
- Temporary data is now stored in /var/lib/embedded-cluster/tmp rather than /tmp prior to this change.
```


cmd/embedded-cluster/restore.go
```diff
@@ -886,370 +882,364 @@ func installAndWaitForRestoredK0sNode(c *cli.Context, applier *addons.Applier) (
...
-			},
-			&cli.StringFlag{
-				Name:  "local-artifact-mirror-port",
-				Usage: "Port on which the Local Artifact Mirror will be served. If left empty, the port will be retrieved from the snapshot.",
-				// DefaultText: strconv.Itoa(defaults.LocalArtifactMirrorPort),
-				Hidden: false,
+func restoreCommand() *cli.Command {
+	runtimeConfig := ecv1beta1.GetDefaultRuntimeConfig()
+
+	return &cli.Command{
```
sgalsaleh, line 888: i believe there should be a `data-dir` flag for the restore command as well because the cluster is set up before the backup is accessible.

cmd/embedded-cluster/restore.go
```diff
@@ -886,370 +882,364 @@ func installAndWaitForRestoredK0sNode(c *cli.Context, applier *addons.Applier) (
...
-		return fmt.Errorf("unable to parse local artifact mirror port from backup: %w", err)
-	}
-	logrus.Debugf("updating local artifact mirror port from backup to %d", port)
-	if err := updateLocalArtifactMirrorService(port); err != nil {
-		return fmt.Errorf("unable to update local artifact mirror service: %w", err)
+// overrideRuntimeConfigFromBackup will update the runtime config from the backup. These values may
+// be used during the install and set in the Installation object via the
+// restoreReconcileInstallationFromRuntimeConfig function.
+func overrideRuntimeConfigFromBackup(c *cli.Context, runtimeConfig *ecv1beta1.RuntimeConfigSpec, backup *velerov1.Backup) (*ecv1beta1.RuntimeConfigSpec, error) {
+	if !c.IsSet("local-artifact-mirror-port") {
```
sgalsaleh, line 1233: i don't see where this flag is used / updates the runtime config

pkg/kotscli/kotscli.go
```diff
@@ -25,7 +26,8 @@ type InstallOptions struct {
 }
 
 func Install(opts InstallOptions, msg *spinner.MessageWriter) error {
-	kotsBinPath, err := goods.MaterializeInternalBinary("kubectl-kots")
+	materializer := goods.NewMaterializer(defaults.EmbeddedClusterHomeDirectory())
```
sgalsaleh, line 29: shouldn't this use the runtimeconfig?

pkg/kotscli/kotscli.go
```diff
@@ -89,7 +91,8 @@ type AirgapUpdateOptions struct {
 }
 
 func AirgapUpdate(opts AirgapUpdateOptions) error {
-	kotsBinPath, err := goods.MaterializeInternalBinary("kubectl-kots")
+	materializer := goods.NewMaterializer(defaults.EmbeddedClusterHomeDirectory())
```
sgalsaleh, line 94: shouldn't this use the runtimeconfig?

pkg/kotscli/kotscli.go
```diff
@@ -134,7 +137,8 @@ type VeleroConfigureOtherS3Options struct {
 }
 
 func VeleroConfigureOtherS3(opts VeleroConfigureOtherS3Options) error {
-	kotsBinPath, err := goods.MaterializeInternalBinary("kubectl-kots")
+	materializer := goods.NewMaterializer(defaults.EmbeddedClusterHomeDirectory())
```
sgalsaleh, line 140: shouldn't this use the runtimeconfig?

pkg/kotscli/kotscli.go
```diff
@@ -25,7 +26,8 @@ type InstallOptions struct {
 }
 
 func Install(opts InstallOptions, msg *spinner.MessageWriter) error {
```
sgalsaleh, line 28: just a reminder that KOTS puts stuff in the tmp directory, and that needs to be consolidated as well

cmd/embedded-cluster/restore.go
emosbaugh: `getDataDirFlag()` function is unsed

cmd/embedded-cluster/restore.go
emosbaugh: It is set in the urfave/cli flag action

https://github.com/replicatedhq/embedded-cluster/pull/1238/files#diff-114d16ace9fc8606c3b1e9a3ba899010aae12b04b9ab398f4973a52e1e63f55dR55

pkg/kotscli/kotscli.go
emosbaugh: this pr only attempts to consolidate the directories. not allow them to be overwritten. i will do that in a follow up

pkg/kotscli/kotscli.go
emosbaugh: this pr only attempts to consolidate the directories. not allow them to be overwritten. i will do that in a follow up

pkg/kotscli/kotscli.go
emosbaugh: this pr only attempts to consolidate the directories. not allow them to be overwritten. i will do that in a follow up

pkg/kotscli/kotscli.go
emosbaugh: This pr does not address the /tmp dir. i will do that in a follow up

emosbaugh: Done

pkg/kotscli/kotscli.go
emosbaugh: It now allows for overwriting

cmd/embedded-cluster/restore.go
emosbaugh: Done

cmd/embedded-cluster/flags.go
```diff
@@ -4,55 +4,73 @@ import (
...
-	port, err := k8snet.ParsePort(portStr, false)
-	if err != nil {
-		return 0, fmt.Errorf("invalid local artifact mirror port: %w", err)
-	}
-	if portStr == c.String("admin-console-port") {
-		return 0, fmt.Errorf("local artifact mirror port cannot be the same as admin console port")
+func getDataDirFlag(runtimeConfig *ecv1beta1.RuntimeConfigSpec) cli.Flag {
+	return &cli.StringFlag{
+		Name:   "data-dir",
+		Usage:  "Path to the Embedded Cluster data directory",
```
ajp-io, line 64: We try to avoid saying "Embedded Cluster" for the end user. This seems like the only user-facing one, but feel free to change to match on the other commands too.
```suggestion
		Usage:  "Path to the data directory",
```

emosbaugh: done

pkg/preflights/host-preflight.yaml
```diff
@@ -158,39 +158,12 @@ spec:
...
-          - pass:
-              message: The filesystem at /var/openebs has sufficient space
-    - diskUsage:
-        checkName: tmp Disk Space
-        collectorName: tmp-path-usage
-        outcomes:
-          - fail:
-              when: 'total < 5Gi'
-              message: The filesystem at /tmp has less than 5Gi of total space
+              message: The filesystem at {{ .DataDir }} is more than 80% full
```
ajp-io, line 164:
```suggestion
              message: The filesystem at {{ .DataDir }} is more than 80% full. Ensure sufficient space is available, or use the --data-dir flag to specify an alternative data directory.
```

pkg/preflights/host-preflight.yaml
```diff
@@ -158,39 +158,12 @@ spec:
...
-          - pass:
-              message: The filesystem at /var/lib/embedded-cluster has sufficient space
-    - diskUsage:
-        checkName: k0s Disk Space
-        collectorName: k0s-path-usage
-        outcomes:
-          - fail:
-              when: 'total < 40Gi'
-              message: The filesystem at /var/lib/k0s has less than 40Gi of total space
+              message: The filesystem at {{ .DataDir }} has less than 40Gi of total space
```
ajp-io, line 161:
```suggestion
              message: The filesystem at {{ .DataDir }} has less than 40 Gi of total space. Ensure sufficient space is available, or use the --data-dir flag to specify an alternative data directory.
```

.github/workflows/ci.yaml
```diff
@@ -319,8 +319,8 @@ jobs:
         run: |
           chmod +x ./output/bin/buildtools
           chmod +x ./output/bin/embedded-cluster-original
-          ./output/bin/embedded-cluster-original version metadata > version-metadata.json
-          ./output/bin/embedded-cluster-original version list-images > expected.txt
+          sudo ./output/bin/embedded-cluster-original version metadata > version-metadata.json
+          sudo ./output/bin/embedded-cluster-original version list-images > expected.txt
```
laverya, line 323: ...why do 'version' commands require sudo again?

cmd/embedded-cluster/install.go
```diff
@@ -81,18 +83,27 @@ func updateLocalArtifactMirrorService(port int) error {
...
 const (
-	localArtifactMirrorSystemdConfFile         = "/etc/systemd/system/local-artifact-mirror.service.d/embedded-cluster.conf"
-	localArtifactMirrorEnvironmentFileContents = `[Service]
-Environment="LOCAL_ARTIFACT_MIRROR_PORT=%d"`
+	localArtifactMirrorSystemdConfFile    = "/etc/systemd/system/local-artifact-mirror.service.d/embedded-cluster.conf"
+	localArtifactMirrorDropInFileContents = `[Service]
+Environment="LOCAL_ARTIFACT_MIRROR_PORT=%d"
+Environment="LOCAL_ARTIFACT_MIRROR_DATA_DIR=%s"
+ExecStart=
+ExecStart=%s serve
```
sgalsaleh, line 91: why is ExecStart in here twice?

cmd/embedded-cluster/join.go
```diff
@@ -347,6 +333,8 @@ var joinCommand = &cli.Command{
 			}
 		}
 
+		tryRemoveTmpDirContents(provider)
```
sgalsaleh, line 336: ditto

cmd/embedded-cluster/restore.go
```diff
@@ -886,370 +882,399 @@ func installAndWaitForRestoredK0sNode(c *cli.Context, applier *addons.Applier) (
...
+					}
+				}
 			}
 
-			logrus.Infof("You'll be guided through the process of restoring %s from a backup.\n", binName)
-			logrus.Info("Enter information to configure access to your backup storage location.\n")
-			s3Store := newS3BackupStore()
+			// If the installation is available, we can further augment the runtime config from the
+			// installation.
+			if state != ecRestoreStateNew &&
```
sgalsaleh, line 982: should this just try to read the installation without checking the state and not fail if it couldn't since it's not critical?

sgalsaleh: feels brittle to have to check states like this as they could change

.github/workflows/ci.yaml
laverya: If instantiating the provider requires sudo, that implies some bad things to me

cmd/embedded-cluster/list_images.go
```diff
@@ -2,27 +2,47 @@ package main
...
+		Usage:  "List images embedded in the cluster",
+		Hidden: true,
+		Flags: []cli.Flag{
+			getDataDirFlag(runtimeConfig),
+		},
+		Before: func(c *cli.Context) error {
+			if os.Getuid() != 0 {
+				return fmt.Errorf("version metadata command must be run as root")
+			}
+			return nil
```
laverya, line 27: I still don't really understand why this takes sudo now

cmd/embedded-cluster/metadata.go
```diff
@@ -65,7 +83,8 @@ func gatherVersionMetadata(k0sCfg *k0sconfig.ClusterConfig) (*types.ReleaseMetad
 		versionsMap[defaults.BinaryName()] = channelRelease.VersionLabel
 	}
 
-	sha, err := goods.K0sBinarySHA256()
+	materializer := goods.NewMaterializer(provider)
+	sha, err := materializer.K0sBinarySHA256()
```
laverya, line 87: this should not require using the provider
there shouldn't be any need to write out the k0s binary, no?

cmd/embedded-cluster/restore.go
```diff
@@ -886,370 +882,399 @@ func installAndWaitForRestoredK0sNode(c *cli.Context, applier *addons.Applier) (
...
+				var err error
+				backupToRestore, err = getBackupFromRestoreState(c.Context, provider, c.String("airgap-bundle") != "")
+				if err != nil {
+					return fmt.Errorf("unable to resume: %w", err)
+				}
+				if backupToRestore != nil {
+					completionTimestamp := backupToRestore.Status.CompletionTimestamp.Time.Format("2006-01-02 15:04:05 UTC")
+					logrus.Infof("Resuming restore from backup %q (%s)\n", backupToRestore.Name, completionTimestamp)
+
+					runtimeConfig, err = overrideRuntimeConfigFromBackup(c, runtimeConfig, backupToRestore)
```
sgalsaleh, line 973: should this update the `provider` as well?

cmd/local-artifact-mirror/serve.go
```diff
@@ -12,22 +12,32 @@ import (
...
 	"github.com/replicatedhq/embedded-cluster/pkg/defaults"
 	"github.com/urfave/cli/v2"
 	k8snet "k8s.io/utils/net"
 )
 
-// serveCommand starts a http server that serves files from the /var/lib/embedded-cluster
-// directory. This server listen only on localhost and is used to serve files needed by
-// the autopilot during an upgrade.
+var (
+	whilelistServeDirs = []string{"bin", "charts", "images"}
```
sgalsaleh, line 22:
```suggestion
	whitelistServeDirs = []string{"bin", "charts", "images"}
```

cmd/embedded-cluster/metadata.go
laverya: yeah, the current code doesn't require writing out the binary at all - so we should just move the function not to depend on a materializer object

operator/pkg/migrations/registry.go
```diff
@@ -110,6 +110,9 @@ func RegistryData(ctx context.Context) error {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      registry.RegistryDataMigrationCompleteSecretName,
 			Namespace: registry.RegistryNamespace(),
+			Labels: map[string]string{
+				"replicated.com/disaster-recovery": "ec-install",
```
sgalsaleh, line 114: why? the `ec-install` label was specific to the `installation` CR/CRD. can we use a different label for this?

sgalsaleh: also, why is it needed now but not before?

operator/pkg/upgrade/upgrade.go
```diff
@@ -31,6 +32,11 @@ func Upgrade(ctx context.Context, cli client.Client, in *clusterv1beta1.Installa
 		return fmt.Errorf("k0s upgrade: %w", err)
 	}
 
+	err = registryMigrationStatus(ctx, cli, in)
```
sgalsaleh, line 35: why would there be a registry migration during an upgrade?

pkg/addons/velero/static/values.tpl.yaml
```diff
@@ -26,5 +26,5 @@ configMaps:
       image: '{{ ImageString (index .Images "velero-restore-helper") }}'
 {{- end }}
 nodeAgent:
-  podVolumePath: /var/lib/k0s/kubelet/pods
+  podVolumePath: /var/lib/embedded-cluster/k0s/kubelet/pods
```
sgalsaleh, line 29: this doesn't seem configurable, no?

sgalsaleh: nevermind.

e2e/version_test.go
```diff
@@ -44,7 +44,7 @@ func TestVersion(t *testing.T) {
 
 	t.Logf("%s: validating 'embedded-cluster version metadata' in node 0", time.Now().Format(time.RFC3339))
 	line2 := []string{"embedded-cluster", "version", "metadata"}
-	stdout, stderr, err = RunRegularUserCommandOnNode(t, tc, 0, line2)
```
laverya, line 47: we should change this back when we fix the metadata command sudo requirement

operator/Makefile
```diff
@@ -62,12 +62,8 @@ manifests: kustomize controller-gen ## Generate WebhookConfiguration, ClusterRol
 fmt: ## Run go fmt against code.
 	go fmt ./...
 
-.PHONY: vet
-vet: ## Run go vet against code.
-	go vet ./...
-
```
laverya, line 68: does vet not work anymore? just wondering why we're getting rid of it

pkg/addons/registry/registry.go
```diff
@@ -120,21 +121,21 @@ func (o *Registry) GenerateHelmConfig(k0sCfg *k0sv1beta1.ClusterConfig, onlyDefa
...
-
-	registryServiceIP, err := helpers.GetLowerBandIP(serviceCIDR, registryLowerBandIPIndex)
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to get cluster IP for registry service: %w", err)
-	}
-	values["service"] = map[string]interface{}{
-		"clusterIP": registryServiceIP.String(),
-	}
-
 	if !onlyDefaults {
```
laverya, line 124: why remove the IPs from the default metadata? It just feels to me like this introduces the possibility of upgrade issues

.github/workflows/ci.yaml
emosbaugh: version itself doesnt but anything that requires us to materialize does because we have a new tmp dir if i remember correctly

cmd/embedded-cluster/install.go
emosbaugh: https://stackoverflow.com/a/68818218

> When you add entries to an override file, they are by default appended to any existing entries. That is, if your service example.service has:

> By specifying an empty ExecStart, you are "clearing out" all previous entries. So if your example.service has:

cmd/embedded-cluster/install.go
```diff
@@ -596,167 +615,168 @@ func validateAdminConsolePassword(password, passwordCheck string) bool {
...
+			cfg, err := installAndWaitForK0s(c, provider, applier, proxy)
+			if err != nil {
+				return err
+			}
+			logrus.Debugf("running outro")
+			if err := runOutro(c, provider, applier, cfg); err != nil {
+				metrics.ReportApplyFinished(c, err)
+				return err
+			}
+			tryRemoveTmpDirContents(provider)
```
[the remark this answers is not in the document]
emosbaugh: yes it should

cmd/embedded-cluster/list_images.go
emosbaugh: its cause we have to create the tmp dir in case anything uses it when we instantiate the provider. could i fast follow with trying to remove the sudo dependency?

cmd/embedded-cluster/metadata.go
emosbaugh: ok, fine...

operator/pkg/upgrade/upgrade.go
laverya: new installation objects don't get the registry migration status copied over

cmd/embedded-cluster/metadata.go
laverya: to clarify, the implementation of `materializer.K0sBinarySHA256()` does not use the `materializer` member functions/variables/etc at all, and so it could be a standalone function and thus not require us to initialize a provider etc

laverya: https://github.com/replicatedhq/embedded-cluster/blob/80952f3da30129f4fa5c29711303a3e0842918b4/pkg/goods/materializer.go#L193

operator/Makefile
emosbaugh: why vet in both operator and root? its the same project

operator/pkg/migrations/registry.go
emosbaugh: this fixes a bug that doesnt seem to have any impact but the registry gets removed entirely from the clusterconfig after an upgrade of an ha cluster. this is because the registry migration status condition gets set to false in the new installation so the ha registry is never added back. This would have impact if we were ever to try to change a registry value as k0s is no longer controlling this object.

see related changes:

https://github.com/replicatedhq/embedded-cluster/pull/1238/files#diff-ec385ffb8fa0ab6903f92cdf907ca63ed0df571f73ca6c1edaabc7395fc1771eR95-R101

https://github.com/replicatedhq/embedded-cluster/pull/1238/files#diff-1595f9a51283933adb1a95f0d4a42f0576e783a1932c0d0da27fa1e5ca5628b3R35

cmd/embedded-cluster/metadata.go
emosbaugh: done

cmd/embedded-cluster/restore.go
```diff
@@ -886,370 +882,399 @@ func installAndWaitForRestoredK0sNode(c *cli.Context, applier *addons.Applier) (
...
-		return nil
+
+	if in.Spec.RuntimeConfig == nil {
+		in.Spec.RuntimeConfig = &ecv1beta1.RuntimeConfigSpec{}
 	}
-	logrus.Debugf("updating local artifact mirror port from flag to %d on installation %q", port, in.Name)
-	in.Spec.LocalArtifactMirror.Port = port
-	if err := kcli.Update(c.Context, in); err != nil {
+
+	// We do not support changing these values so we must choose the values from the installation
```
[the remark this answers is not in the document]
emosbaugh: i will remove this. this is from before i added that annotation and check

pkg/addons/registry/registry.go
emosbaugh: i guess i do not understand what onlyDefaults means then. what is not default about the tlsSecretName?

cmd/embedded-cluster/restore.go
emosbaugh: it does on line 995

https://github.com/replicatedhq/embedded-cluster/pull/1238/files#diff-86e2f8fce801979bbdfe9a3e2f20ace3982b4f7e25189c832a39381e7c3a902dR995

pkg/addons/registry/registry.go
laverya: "onlyDefaults" is what we pass when rendering the chart metadata for a release 

Apparently we don't include the tlsSecretName in that, but I don't understand why we also want to stop including the default service IP config

laverya: If it's intentional that's fine but it didn't seem related to the major change to me

operator/pkg/upgrade/upgrade.go
sgalsaleh: and they shouldn't, no?

operator/pkg/migrations/registry.go
sgalsaleh: i see. any reason not to have it use the registry DR label instead of the ec-install one?

operator/pkg/upgrade/upgrade.go
emosbaugh: it needs it else there is a bug explained here https://github.com/replicatedhq/embedded-cluster/pull/1238#discussion_r1793917391

operator/pkg/migrations/registry.go
emosbaugh: i believe if i use the registry component it is too late and the operator will revert the registry back to non-ha in the clusterconfig. i would be wrong

laverya (dismissed):

operator/pkg/upgrade/upgrade.go
sgalsaleh: i see. seems like a bandaid covering a bigger issue in the reconciliation process.

sgalsaleh: i wonder if it should be preserved by whatever creates the new installation object though like everything else

emosbaugh: ya i agree its a bandaid. i did create this. i think we could improve the logic here

https://app.shortcut.com/replicated/story/113699/it-is-not-possible-to-migrate-to-high-availability-after-a-restore

laverya (approved):

sgalsaleh (approved):

emosbaugh:
E2E test is failing with the following error which is unrelated to this change

```
✗  1 host preflight failed
        
         •  The filesystem at /var/lib/embedded-cluster is more than 80% full. Ensure sufficient space is available, or use the --data-dir flag to specify an 
            alternative data directory.                                                                                                                       
        
        Please address this issue and try again.
```

PR #1241, opened 2024-09-27
fix(ci): buildtools handle helm chart app version change

emosbaugh:

[Seaweedfs](https://artifacthub.io/packages/helm/seaweedfs/seaweedfs) changed their app version in the helm chart without changing the helm chart version. This does not play well with our shrinkwrapped releases and now we have cves in the seaweed image.

https://github.com/replicatedhq/embedded-cluster/actions/runs/11060070387/job/30733862137

Right now im just warning but i think we need to handle this somehow.

```
WARN[0004] dst tag exists but app versions do not match (src: 3.73, dst: 3.71) 
```

```release-note

```

### DIFF
--- a/cmd/buildtools/utils.go
+++ b/cmd/buildtools/utils.go
@@ -23,6 +23,7 @@ import (
 	"github.com/replicatedhq/embedded-cluster/pkg/helm"
 	"github.com/replicatedhq/embedded-cluster/pkg/release"
 	"github.com/sirupsen/logrus"
+	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/repo"
 )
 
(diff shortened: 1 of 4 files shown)


cmd/buildtools/utils.go
```diff
@@ -475,13 +491,17 @@ func MirrorChart(repo *repo.Entry, name, ver string) error {
...
 	if err != nil && !strings.HasSuffix(err.Error(), "not found") {
 		return fmt.Errorf("verify tag exists: %w", err)
 	} else if err == nil {
-		os.Remove(tmpf)
-		logrus.Warnf("cowardly refusing to override dst (tag %s already exist)", ver)
+		if srcMeta.AppVersion == dstMeta.AppVersion {
+			logrus.Infof("cowardly refusing to override dst (tag %s already exist)", ver)
+			return nil
+		}
+		logrus.Warnf("dst tag exists but app versions do not match (src: %s, dst: %s)", srcMeta.AppVersion, dstMeta.AppVersion)
```
sgalsaleh, line 504: shouldn't the idea be to push in this case? why is there a return statement?

emosbaugh: doesnt that defeat the purpose of shrinkwrapping?

sgalsaleh (approved):

PR #1262, opened 2024-10-01
fix: validate admin console password length

JGAntunes:
Make sure we validate the admin console password length requiring it to be at least 6 characters, similar to what already [happens in KOTS](https://github.com/replicatedhq/kots/blob/dab9434bb24c13ffb5f37b528c515121b31db089/pkg/util/password_prompt.go#L22-L25)

https://app.shortcut.com/replicated/story/104936/admin-console-minimum-password-length-inconsistent

I've added a test to the password validation function and was trying to see if there's a way that we could test the `maybeAskAdminConsolePassword` method, but since it relies on input prompts I'm unsure what would be the best mechanism to do so 🤔 open to suggestions.

Meanwhile I've also run a couple of manual tests in the local dev env:
```
root@node0:/replicatedhq/embedded-cluster# output/bin/embedded-cluster install --license local-dev/Joao.yaml
? Set the Admin Console password: ***
? Confirm the Admin Console password: ***
Passwords must have more than 6 characters. Please try again.
? Set the Admin Console password: ******
? Confirm the Admin Console password: ******
Passwords don't match. Please try again.
? Set the Admin Console password: ******
? Confirm the Admin Console password: ******
✔  Host files materialized!
(...)
```

```
root@node0:/replicatedhq/embedded-cluster# output/bin/embedded-cluster install --license local-dev/Joao.yaml --no-prompt
The Admin Console password is set to password
(...)
```

```
root@node0:/replicatedhq/embedded-cluster# output/bin/embedded-cluster install --license local-dev/Joao.yaml --admin-console-password 123
Passwords must have more than 6 characters. Please try again.
unable to set the Admin Console password
root@node0:/replicatedhq/embedded-cluster# output/bin/embedded-cluster install --license local-dev/Joao.yaml --admin-console-password 123456
✔  Host files materialized!
(...)
```

```
root@node0:/replicatedhq/embedded-cluster# output/bin/embedded-cluster install help
NAME:
   embedded-cluster install - Install embedded-cluster

USAGE:
   embedded-cluster install command [command options]

COMMANDS:
   help, h  Shows a list of commands or help for one command

OPTIONS:
   --admin-console-password value             Password for the Admin Console (minimum 6 characters)
   (...)
```

```
root@node0:/replicatedhq/embedded-cluster# output/bin/embedded-cluster install --license local-dev/Joao.yaml
? Set the Admin Console password (minimum 6 characters):
unable to ask for input: interrupt
```

```release-note
NONE
```

NONE

### DIFF
--- a/cmd/embedded-cluster/install.go
+++ b/cmd/embedded-cluster/install.go
@@ -567,14 +567,14 @@ func maybeAskAdminConsolePassword(c *cli.Context) (string, error) {
 		promptA := prompts.New().Password("Set the Admin Console password:")
 		promptB := prompts.New().Password("Confirm the Admin Console password:")
 
-		if validatePassword(promptA, promptB) {
+		if validateAdminConsolePassword(promptA, promptB) {
 			return promptA, nil
 		}
 	}
 	return "", fmt.Errorf("unable to set the Admin Console password after %d tries", maxTries)
 }
 
-func validatePassword(password, passwordCheck string) bool {
+func validateAdminConsolePassword(password, passwordCheck string) bool {
 	if len(password) < 6 {
 		logrus.Info("Passwords must have more than 6 characters. Please try again.")
 		return false
--- a/cmd/embedded-cluster/install.go
+++ b/cmd/embedded-cluster/install.go
@@ -551,36 +551,37 @@ func runOutro(c *cli.Context, applier *addons.Applier, cfg *k0sconfig.ClusterCon
 func maybeAskAdminConsolePassword(c *cli.Context) (string, error) {
 	defaultPassword := "password"
 	userProvidedPassword := c.String("admin-console-password")
-	if c.Bool("no-prompt") {
-		if userProvidedPassword != "" {
-			return userProvidedPassword, nil
-		} else {
-			logrus.Infof("The Admin Console password is set to %s", defaultPassword)
-			return defaultPassword, nil
-		}
-	}
+	// If there's a user provided password we'll try that first
 	if userProvidedPassword != "" {
+		if !validateAdminConsolePassword(userProvidedPassword, "", false) {
+			return "", fmt.Errorf("unable to set the Admin Console password")
+		}
 		return userProvidedPassword, nil
+	}
+	// No user provided password but prompt is disabled so we set our default password
+	if c.Bool("no-prompt") {
+		logrus.Infof("The Admin Console password is set to %s", defaultPassword)
+		return defaultPassword, nil
 	}
 	maxTries := 3
 	for i := 0; i < maxTries; i++ {
 		promptA := prompts.New().Password("Set the Admin Console password:")
 		promptB := prompts.New().Password("Confirm the Admin Console password:")
 
-		if validateAdminConsolePassword(promptA, promptB) {
+		if validateAdminConsolePassword(promptA, promptB, true) {
 			return promptA, nil
 		}
 	}
 	return "", fmt.Errorf("unable to set the Admin Console password after %d tries", maxTries)
 }
 
-func validateAdminConsolePassword(password, passwordCheck string) bool {
-	if len(password) < 6 {
-		logrus.Info("Passwords must have more than 6 characters. Please try again.")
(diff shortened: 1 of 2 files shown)


cmd/embedded-cluster/install.go
```diff
@@ -551,31 +554,42 @@ func runOutro(c *cli.Context, applier *addons.Applier, cfg *k0sconfig.ClusterCon
 func maybeAskAdminConsolePassword(c *cli.Context) (string, error) {
 	defaultPassword := "password"
 	userProvidedPassword := c.String("admin-console-password")
-	if c.Bool("no-prompt") {
-		if userProvidedPassword != "" {
-			return userProvidedPassword, nil
-		} else {
```
JGAntunes, line 557: We always prioritised the user provided password whether we have `no-prompt` or not.

cmd/embedded-cluster/install.go
```diff
@@ -551,31 +554,42 @@ func runOutro(c *cli.Context, applier *addons.Applier, cfg *k0sconfig.ClusterCon
...
-			logrus.Infof("The Admin Console password is set to %s", defaultPassword)
-			return defaultPassword, nil
-		}
-	}
+	// If there's a user provided password we'll try that first
 	if userProvidedPassword != "" {
+		if !validateAdminConsolePassword(userProvidedPassword, "", false) {
+			return "", fmt.Errorf("unable to set the Admin Console password")
+		}
```
JGAntunes, line 561: I'm unsure if we only wanna validate the prompts on the case where the user provides a password through a CLI param. I assumed the latter.

cmd/embedded-cluster/install.go
```diff
@@ -551,31 +554,42 @@ func runOutro(c *cli.Context, applier *addons.Applier, cfg *k0sconfig.ClusterCon
...
-			// TODO: Should we add extra password validation here? e.g length, complexity etc
+		if validateAdminConsolePassword(promptA, promptB, true) {
 			return promptA, nil
 		}
-		logrus.Info("Passwords don't match. Please try again.")
 	}
 	return "", fmt.Errorf("unable to set the Admin Console password after %d tries", maxTries)
 }
 
+func validateAdminConsolePassword(password, passwordCheck string, mustMatch bool) bool {
```
laverya, line 581: The only time this is called with `mustMatch: false` is when it's provided via the CLI. Would it make more sense for that function call to just supply the password twice to the function?

cmd/embedded-cluster/install_test.go
```diff
@@ -200,3 +200,70 @@ versionLabel: testversion
 		})
 	}
 }
+
+func Test_validateAdminConsolePassword(t *testing.T) {
+
+	tests := []struct {
```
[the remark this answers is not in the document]
laverya: we have the ability to run integration tests with `expect`, but that is probably a bit farther than we want to go here

cmd/embedded-cluster/install.go
JGAntunes: Personally, I prefer being explicit about the functionality VS having it be dependent on behaviour which is non-explicit. But happy to change it if y'all think differently.

cmd/embedded-cluster/install_test.go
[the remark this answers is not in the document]
JGAntunes: I've added a simple set of tests to `maybeAskAdminConsolePassword` without exercising the prompts.

cmd/embedded-cluster/install.go
laverya: my feeling is that 'mustMatch' isn't particularly explicit - if we're going to do something like that, we should have two functions one for 'passwordLength' and another for 'passwordsMatch' IMO

@sgalsaleh ?

JGAntunes: Went with your original suggestion via - https://github.com/replicatedhq/embedded-cluster/pull/1262/commits/9ab78499e3eefeed87cf8bb11998303c503ef779 - let me know what you think @laverya

laverya (approved):

PR #1265, opened 2024-10-02
Enable goldpinger collector and analyzer

hedge-sparrow:

Enables the goldpinger collector and analyzer now that the collector is capable of temporarily provisioning goldpinger. 

```release-note
Enable goldpinger collector and analyzer
```

NONE

### DIFF
--- a/operator/charts/embedded-cluster-operator/troubleshoot/cluster-support-bundle.yaml
+++ b/operator/charts/embedded-cluster-operator/troubleshoot/cluster-support-bundle.yaml
@@ -7,6 +7,7 @@ metadata:
 spec:
   uri: https://raw.githubusercontent.com/replicatedhq/embedded-cluster-operator/main/charts/embedded-cluster-operator/troubleshoot/cluster-support-bundle.yaml
   collectors:
+  - goldpinger: {}
   - logs:
       name: podlogs/embedded-cluster-operator
       namespace: embedded-cluster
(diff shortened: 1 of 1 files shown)


laverya (dismissed):

laverya:
What namespace does the goldpinger collector run in?

Does it work in airgap? How?

hedge-sparrow:
> What namespace does the goldpinger collector run in?
default but can be overridden in the spec if necessary

> Does it work in airgap? How?
not yet but it'll fail open if the images can't be pulled, we still need some thought on how to include the images in the EC airgap builder

laverya:
Could we change the spec to run goldpinger in the `goldpinger` namespace? That's more likely to match what the vendor would have setup IMO

laverya (dismissed):

hedge-sparrow:
I've updated this so that it'll only enable the collector/analyzer in online installations as a temporary workaround to not having the images in airgap yet

operator/charts/embedded-cluster-operator/templates/embedded-cluster-troubleshoot-goldpinger.yaml
```diff
@@ -0,0 +1,28 @@
...
+    kind: SupportBundle
+    metadata:
+      name: embedded-cluster-troubleshoot-goldpinger
+      labels:
+        troubleshoot.sh/kind: support-bundle
+    spec:
+      collectors:
+      - goldpinger:
+          namespace: goldpinger
+          image: bloomberg/goldpinger:3.10.1
```
sgalsaleh, line 22: shouldn't this go through `proxy.replicated.com/anonymous/` like all the other images? because the customer env might not allow traffic to dockerhub. i would recommend using a digest as well so the image gets cached and still works if there is a dockerhub outage.

ricardomaraschini (approved):

PR #1302, opened 2024-10-09
feat: check if data-dir is a symlink

ricardomaraschini:

data directory can't be a symlink, we have decided to fail if this is the case.


ricardomaraschini:
Replaced by https://github.com/replicatedhq/embedded-cluster/pull/1308

PR #1308, opened 2024-10-10
feat: check if data-dir is a symlink

ricardomaraschini:

data directory can't be a symlink, we have decided to fail if this is the case. this pr adds a preflight check for that.

### DIFF
--- a/pkg/preflights/host-preflight.yaml
+++ b/pkg/preflights/host-preflight.yaml
@@ -116,6 +116,13 @@ spec:
     - udpPortStatus:
         collectorName: Calico Communication Port
         port: 4789
+    - run:
+        collectorName: check-data-dir-symlink
+        command: sh
+        args:
+          - -c
+          - |
+            [ -d "{{ .DataDir }}" ] && [ -L "{{ .DataDir }}" ] && echo "{{ .DataDir }} is a symlink" || echo "{{ .DataDir }} is not a symlink"
   analyzers:
     - cpu:
         checkName: CPU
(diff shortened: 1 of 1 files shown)


pkg/preflights/host-preflight.yaml
```diff
@@ -116,6 +116,13 @@ spec:
     - udpPortStatus:
         collectorName: Calico Communication Port
         port: 4789
+    - run:
+        collectorName: check-data-dir-symlink
+        command: sh
+        args:
+          - -c
+          - |
+            [ -d "{{ .DataDir }}" ] && [ -L "{{ .DataDir }}" ] && echo "{{ .DataDir }} is a symlink" || echo "{{ .DataDir }} is not a symlink"
```
ajp-io, line 125: So if datadir isn't a directory, we print "not a symlink." That's probably fine if Ethan's changes validate that datadir is a directory. Does it?

ricardomaraschini: That is a good point. It may be that someone just created a file there with the `embedded-cluster` name.

ricardomaraschini: If the data dir is a file then the installation bails out as it can't turn the file into a directory in order to materialize:

```
root@red-water:/home/developer# echo xxx >/var/lib/embedded-cluster
root@red-water:/home/developer# ./embedded-cluster install
unable to create embedded-cluster tmp dir: mkdir /var/lib/embedded-cluster: not a directory
root@red-water:/home/developer#
```

emosbaugh (approved):

PR #1318, opened 2024-10-14
feat: detect if a mitm proxy is being used in host preflights

adamancini:

```release-note

```


kinds/apis/v1beta1/runtimeconfig_types.go
```diff
@@ -26,6 +26,12 @@ type RuntimeConfigSpec struct {
 	AdminConsole AdminConsoleSpec `json:"adminConsole,omitempty"`
 	// LocalArtifactMirrorPort holds the Local Artifact Mirror configuration.
 	LocalArtifactMirror LocalArtifactMirrorSpec `json:"localArtifactMirror,omitempty"`
+
+	// Proxy holds the proxy configuration.
+	Proxy *ProxySpec `json:"proxy,omitempty"`
```
emosbaugh, line 31: Ultimately this does belong here but I think a refactor must first be done by our team

kinds/apis/v1beta1/runtimeconfig_types.go
```diff
@@ -26,6 +26,12 @@ type RuntimeConfigSpec struct {
...
+
+	// Private CA holds a path to a trusted private CA certificate PEM file.
+	PrivateCA string `json:"privateCA,omitempty"`
```
emosbaugh, line 34: this will not be compatible with prior versions of the product but i think that is ok since it is a new feature.

you will need to set this via the install command. this will also require you to update kots with the latest kinds which means a kinds release is required. i can help with that when ready.

pkg/defaults/provider.go
```diff
@@ -187,3 +187,38 @@ func (d *Provider) AdminConsolePort() int {
...
+	}
+	return ""
+}
+
+func (d *Provider) NoProxy() string {
+	if d.runtimeConfig != nil && d.runtimeConfig.Proxy != nil {
+		return d.runtimeConfig.Proxy.NoProxy
+	}
+	return ""
+}
```
emosbaugh, line 217: remove this

cmd/embedded-cluster/preflights.go
```diff
@@ -150,13 +150,13 @@ var joinRunPreflightsCommand = &cli.Command{
 			return fmt.Errorf("embedded cluster version mismatch - this binary is version %q, but the cluster is running version %q", versions.Version, jcmd.EmbeddedClusterVersion)
 		}
 
-		setProxyEnv(jcmd.InstallationSpec.Proxy)
-		proxyOK, localIP, err := checkProxyConfigForLocalIP(jcmd.InstallationSpec.Proxy, c.String("network-interface"))
+		setProxyEnv(jcmd.InstallationSpec.RuntimeConfig.Proxy)
```
emosbaugh, line 153: revert all changes in this file back to using `jcmd.InstallationSpec.Proxy`

cmd/embedded-cluster/join.go
```diff
@@ -202,13 +202,13 @@ var joinCommand = &cli.Command{
 			return fmt.Errorf("embedded cluster version mismatch - this binary is version %q, but the cluster is running version %q", versions.Version, jcmd.EmbeddedClusterVersion)
 		}
 
-		setProxyEnv(jcmd.InstallationSpec.Proxy)
-		proxyOK, localIP, err := checkProxyConfigForLocalIP(jcmd.InstallationSpec.Proxy, c.String("network-interface"))
+		setProxyEnv(jcmd.InstallationSpec.RuntimeConfig.Proxy)
```
emosbaugh, line 205: revert all changes in this file back to using `jcmd.InstallationSpec.Proxy`

cmd/embedded-cluster/install.go
```diff
@@ -161,6 +161,11 @@ func RunHostPreflights(c *cli.Context, provider *defaults.Provider, applier *add
 		K0sDataDir:              provider.EmbeddedClusterK0sSubDir(),
 		OpenEBSDataDir:          provider.EmbeddedClusterOpenEBSLocalSubDir(),
 		SystemArchitecture:      runtime.GOARCH,
+		PrivateCA:               provider.PrivateCA(),
+		HTTPProxy:               provider.HttpProxy(),
+		HTTPSProxy:              provider.HttpsProxy(),
+		ProvidedNoProxy:         provider.ProvidedNoProxy(),
+		NoProxy:                 provider.NoProxy(),
```
emosbaugh, line 168: are you able to use the `proxy *ecv1beta1.ProxySpec` arg instead rather than the provider?

emosbaugh (changes requested):

cmd/embedded-cluster/preflights.go
adamancini: done

cmd/embedded-cluster/join.go
adamancini: done

kinds/apis/v1beta1/runtimeconfig_types.go
adamancini: ack

pkg/defaults/provider.go
adamancini: done

cmd/embedded-cluster/install.go
adamancini: yeah, that seems to work

kinds/apis/v1beta1/runtimeconfig_types.go
adamancini: @emosbaugh is it expected that a user might pass multiple CA arguments to the installer?  I see that in https://github.com/replicatedhq/embedded-cluster/blob/mitm-proxy-preflight/cmd/embedded-cluster/install.go#L817 we parse the install command and build a map that might hold multiple CA files - I might need to go back to the collector code to make sure that the collector can import more than one

adamancini: thinking something like https://github.com/replicatedhq/embedded-cluster/pull/1318/files#diff-998cd2ce861bc777f6423c98894319d4fa0137fdefef5a3e0bf6987c4ed6d4b2R621 I can use to get just the paths and then pass the paths as a list

emosbaugh: it looks like they can pass multiple. i think that makes sense.

kinds/apis/v1beta1/runtimeconfig_types.go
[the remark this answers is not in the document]
adamancini: yep, thats been removed in my latest

adamancini:
@emosbaugh at https://github.com/replicatedhq/embedded-cluster/blob/mitm-proxy-preflight/cmd/embedded-cluster/install.go#L165 I can try to access the proxy variables from `proxy` that gets passed in, but if those are blank then I get segfaults when trying to put this struct together - is there a better way to do that than simply checking each `proxy.HTTPProxy`, `proxy.HTTPSProxy`... 

I tried doing this as a test

```
	data := preflights.TemplateData{
		ReplicatedAPIURL:        replicatedAPIURL,
		ProxyRegistryURL:        proxyRegistryURL,
		IsAirgap:                isAirgap,
		AdminConsolePort:        provider.AdminConsolePort(),
		LocalArtifactMirrorPort: provider.LocalArtifactMirrorPort(),
		DataDir:                 provider.EmbeddedClusterHomeDirectory(),
		K0sDataDir:              provider.EmbeddedClusterK0sSubDir(),
		OpenEBSDataDir:          provider.EmbeddedClusterOpenEBSLocalSubDir(),
		SystemArchitecture:      runtime.GOARCH,
		PrivateCA:               "",
		HTTPProxy:               "",
		HTTPSProxy:              "",
		ProvidedNoProxy:         "",
		NoProxy:                 "",
	}
```

and that passes the point where I would have segfaulted

emosbaugh:
you need to run the following to pass ci

```
make -C kinds generate
make -C operator manifests
```

cmd/embedded-cluster/install.go
```diff
@@ -638,6 +649,15 @@ func validateAdminConsolePassword(password, passwordCheck string) bool {
...
 }
 
+// return only the first private CA path for now - troubleshoot needs a refactor to support multiple CAs in the future
+func getPrivateCAPath(c *cli.Context) string {
+	privateCA := ""
+	if len(c.StringSlice("private-ca")) > 0 {
+		privateCA = c.StringSlice("private-ca")[0]
+	}
+	return privateCA
+}
```
laverya, line 659: This won't work on joins - I think you'll need to add the CAs to the preflight inputs

operator/charts/embedded-cluster-operator/Chart.yaml
```diff
@@ -1,9 +1,8 @@
-# DEVELOPMENT USE ONLY! THIE FILE IS NOT USED FOR DEPLOYS!
 apiVersion: v2
 name: embedded-cluster-operator
 description: Embedded Cluster Operator
 type: application
-version: 0.0.0
+version: 1.16.0+k8s-1.30-24-g21dbb2dc
```
laverya, line 5: this should be reverted

operator/charts/embedded-cluster-operator/values.yaml
```diff
@@ -9,8 +8,8 @@ embeddedBinaryName: v0.0.0
 embeddedClusterID: 123456789
 
 image:
-  repository: replicated/embedded-cluster-operator-image-staging
-  tag: 0.0.8
+  repository: ttl.sh/ada/embedded-cluster-operator-image
+  tag: 1.16.0-k8s-1.30-24-g21dbb2dc
```
laverya, line 12: this should be reverted too

pkg/addons/embeddedclusteroperator/static/metadata.yaml
```diff
@@ -5,16 +5,16 @@
 # $ make buildtools
 # $ output/bin/buildtools update addon <addon name>
 #
-version: 1.12.1+k8s-1.29
-location: oci://proxy.replicated.com/anonymous/registry.replicated.com/library/embedded-cluster-operator
+version: 1.16.0+k8s-1.30-24-g21dbb2dc
+location: oci://proxy.replicated.com/anonymous/ttl.sh/ada/embedded-cluster-operator
```
laverya, line 9: this should be reverted three

pkg/preflights/template.go
```diff
@@ -82,5 +87,7 @@ func renderTemplate(spec string, data TemplateData) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("execute template: %w", err)
 	}
+
+	fmt.Printf("%+v\n", buf.String())
```
laverya, line 91: this looks like it was included for debugging and should not make it to prod

laverya (dismissed):

laverya (approved):

laverya (approved):

emosbaugh (approved):

PR #1321, opened 2024-10-14
merge 'install' and 'upgrade' helm spec generation code

laverya:

Currently, the initial install and the operator generate helm charts in independent processes. The installer actually owns the charts + values, which are rendered to a 'metadata' file for an installation. The operator then consumes the metadata file, and overrides certain values on upgrades.

If the overrides are incomplete, then features can be rolled back upon upgrade, which isn't great - and at best it means maintaining code in two places.

This PR has the operator call the same code to generate helm charts as the initial install, with only two pieces of dependence on the metadata.json file - the operator images and chart upstream location. (the images can't be compiled into the operator image itself as they contain the content sha of the image itself, and while the operator chart upstream _could_ be compiled in properly I found it far easier just to rely on the metadata there too)

```release-note

```


operator/pkg/charts/detection.go
```diff
@@ -102,15 +102,18 @@ func DetectChartCompletion(existingHelm *v1beta1.HelmExtensions, installedCharts
...
 				}
 
 				// if the spec's HashValues does not match the status's ValuesHash, the chart is currently being applied with the new values
 				if installedChart.Spec.HashValues() != installedChart.Status.ValuesHash {
+					fmt.Printf("Chart %s is currently being applied with new values, waiting for hash %q\n", chart.Name, installedChart.Spec.HashValues())
 					diffDetected = true
 				}
 
 				if installedChart.Status.Version != chart.Version {
+					fmt.Printf("Chart %s version is different from the current version\n", chart.Name)
```
laverya, line 116: these need to get removed....

pkg/addons/embeddedclusteroperator/embeddedclusteroperator.go
```diff
@@ -263,43 +263,58 @@ func (e *EmbeddedClusterOperator) Outro(ctx context.Context, provider *defaults.
...
-				IsDisasterRecoverySupported: licenseDisasterRecoverySupported(license),
+				IsDisasterRecoverySupported: licenseDisasterRecoverySupported(e.license),
 			},
 		},
 	}
 	if err := cli.Create(ctx, &installation); err != nil {
 		return fmt.Errorf("unable to create installation: %w", err)
 	}
+
+	gotInstallation, err := waitForInstallationToExist(ctx, cli, installation.Name)
```
emosbaugh, line 283: i cant understand at all why you have this right after creating it. you will need a comment here to explain it.

pkg/addons/embeddedclusteroperator/embeddedclusteroperator.go
```diff
@@ -7,6 +7,8 @@ import (
 	_ "embed"
 	"encoding/json"
 	"fmt"
+	"k8s.io/apimachinery/pkg/api/errors"
+	types2 "k8s.io/apimachinery/pkg/types"
```
emosbaugh, line 11: how about apitypes?

operator/pkg/cli/version.go
```diff
@@ -0,0 +1,60 @@
+package cli
+
+import (
+	"fmt"
+	"github.com/replicatedhq/embedded-cluster/pkg/addons/embeddedclusteroperator"
```
emosbaugh, line 5: this should probably be moved down to the other block

operator/pkg/charts/charts.go
```diff
@@ -77,51 +85,88 @@ func mergeHelmConfigs(ctx context.Context, meta *ectypes.ReleaseMetadata, in *cl
...
+			},
+		},
+		"utils": {
+			Repo: strings.Split(ecUtilsImage, ":")[0],
+			Tag: map[string]string{
+				"amd64": strings.Join(strings.Split(ecUtilsImage, ":")[1:], ":"),
+				"arm64": strings.Join(strings.Split(ecUtilsImage, ":")[1:], ":"),
+			},
+		},
+	}
```
emosbaugh, line 128: could you split this out into its own function operatorImages?

emosbaugh (changes requested):

pkg/addons/embeddedclusteroperator/embeddedclusteroperator.go
laverya: Because otherwise it's a race condition with the apiserver

pkg/addons/embeddedclusteroperator/embeddedclusteroperator.go
```diff
@@ -326,3 +341,18 @@ func k0sConfigToNetworkSpec(k0sCfg *k0sv1beta1.ClusterConfig) *ecv1beta1.Network
 
 	return network
 }
+
+func waitForInstallationToExist(ctx context.Context, cli client.Client, name string) (*ecv1beta1.Installation, error) {
+	var installation ecv1beta1.Installation
+	for i := 0; i < 20; i++ {
+		if err := cli.Get(ctx, types2.NamespacedName{Name: name}, &installation); err != nil {
```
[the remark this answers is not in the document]
laverya: that's part of why it's a local function, but yeah I'll use the more generally applicable kubeutils instead

operator/pkg/charts/charts.go
laverya: yep yep, that would make sense

pkg/addons/embeddedclusteroperator/embeddedclusteroperator.go
laverya: comment added

emosbaugh (approved):

PR #1324, opened 2024-10-15
feat: introduces a new `--cidr` command line flag

ricardomaraschini:

This commit introduces the `--cidr` flag. This flag is meant to replace the `--pod-cidr` and `--service-cidr` flags (these are now hidden). When provided `--cidr` must be equal or larger than a `/16` and half of the IPs go to the pod network while the other half goes to the service network.

The default value for `--cidr` is defined as `10.244.0.0/16` resulting in pods getting the `10.244.0.0/17` and services getting the `10.244.128.0/17` network addresses.

```release-note
Introduced the new --cidr flag, this flag defines both Pod and Service network addresses (half each).
```


pkg/netutils/nets_test.go
```diff
@@ -0,0 +1,168 @@
...
+			a:    "192.168.0.0/23",
+			b:    "192.168.2.0/24",
+			want: false,
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			got, _, err := NetworksAreAdjacentAndSameSize(tt.a, tt.b)
+			assert.NoError(t, err)
+			assert.Equal(t, tt.want, got)
+		})
```
laverya, line 135: can we test the returned value here too? Writing it out to a string might make this easier

e2e/install_test.go
```diff
@@ -951,7 +950,15 @@ func TestSingleNodeAirgapUpgradeCustomCIDR(t *testing.T) {
...
 	// ensure that the cluster is using the right IP ranges.
 	t.Logf("%s: checking service and pod IP addresses", time.Now().Format(time.RFC3339))
-	if stdout, stderr, err := tc.RunCommandOnNode(0, []string{"check-cidr-ranges.sh", "^10.128.[0-9]*.[0-9]", "^10.129.[0-9]*.[0-9]"}); err != nil {
+
+	// we have used --cidr 192.168.0.0/16 during install time so pods are
+	// expected to be in the 192.168.0.0/17 range while services are in the
+	// 192.168.128.0/17 range. i.e. pods are in 192.168.0-127.* while
+	// services in 192.168.128-254.*.
+	podregex := `192\.168\.\([0-9]\|[1-9][0-9]\|12[0-7]\)\.`
+	svcregex := `192\.168\.\(12[8-9]\|1[3-9][0-9]\|[2-5][0-9][0-9]\)\.`
```
laverya, line 959: these regexes sound like a great argument for using the same ranges as the original test, ooof

ricardomaraschini: Those are not supported as we require at least a /16.

pkg/netutils/nets_test.go
ricardomaraschini: I will be opening a PR with this.

PR #1326, opened 2024-10-15
Unhide air gap and alphabetize flags

ajp-io:

I realized the air gap flags are still hidden, so I unhid those. I also alphabetized the CLI flags in the help menus so they're easier to read. The CIDR and proxy flags are from functions, so those didn't get alphabetized. Ultimately I'd like them to be alphabetized, but I didn't bother to rework those functions.

NONE

```release-note
Unhides the `--airgap-bundle` flag for the `install`, `join`, and `restore` commands.
```

NONE


ajp-io:
For example:
Screenshot 2024-10-15 at 11 35 55 AM

laverya (approved):

PR #1341, opened 2024-10-16
fix: upgrades from versions prior to 1.15.0 have incorrect data directories

emosbaugh:

The data directories are not preserved when upgrading from a version prior to the introduction of the new data directories. Additionally, airgap upgrades fail when there is a worker node because LAM crashes since there is no kubeconfig which it is using to discover the data directories.

```release-note
- Fixes an issue that prevents upgrading airgap installations with at least one worker node.
- Fixes an issue that causes Velero pods to crash on upgrades from versions prior to 1.15.0.
```


operator/pkg/charts/charts.go
```diff
@@ -282,6 +282,12 @@ func updateInfraChartsFromInstall(in *v1beta1.Installation, clusterConfig *k0sv1
 					return nil, fmt.Errorf("set helm values velero.configuration: %w", err)
 				}
 
+				podVolumePath := filepath.Join(provider.EmbeddedClusterK0sSubDir(), "kubelet/pods")
```
emosbaugh, line 285: this is not directly related but is a bug i found where this is being reverted on upgrade

operator/pkg/charts/charts.go
```diff
@@ -282,6 +282,12 @@ func updateInfraChartsFromInstall(in *v1beta1.Installation, clusterConfig *k0sv1
...
+				newVals, err = helm.SetValue(newVals, "nodeAgent.podVolumePath", podVolumePath)
```
emosbaugh, line 286: this is not directly related but is a bug i found where this is being reverted on upgrade

.github/actions/e2e/action.yml
```diff
@@ -108,7 +108,7 @@ runs:
       export EXPECT_K0S_VERSION_PREVIOUS=${{ inputs.k0s-version-previous }}
       make e2e-test TEST_NAME=${{ inputs.test-name }}
   - name: Troubleshoot
-    if: ${{ failure() }}
+    if: ${{ !cancelled() }}
```
emosbaugh, line 111: TODO: revert this

e2e/cluster/docker/cluster.go
```diff
@@ -82,10 +82,10 @@ func (c *Cluster) WaitForReady() {
 }
 
 func (c *Cluster) Cleanup(envs ...map[string]string) {
-	if c.t.Failed() {
-		c.generateSupportBundle(envs...)
-		c.copyPlaywrightReport()
-	}
+	// if c.t.Failed() {
```
emosbaugh, line 85: TODO: revert this

e2e/cluster/lxd/cluster.go
```diff
@@ -989,10 +989,10 @@ func (c *Cluster) InstallTestDependenciesDebian(t *testing.T, node int, withProx
 }
 
 func (c *Cluster) Cleanup(envs ...map[string]string) {
-	if c.T.Failed() {
-		c.generateSupportBundle(envs...)
-		c.copyPlaywrightReport()
-	}
+	// if c.T.Failed() {
```
emosbaugh, line 992: TODO: revert this

laverya (approved):

laverya (approved):

PR #1346, opened 2024-10-19
bug(seaweedfs): fix data path, add validation in e2e tests

emosbaugh:

The seaweedfs data paths are incorrect because the values are set wrong (global.data.hostPathPrefix vs master.data.hostPathPrefix).

Additionally, fixes an issue that prevents data paths from being set correctly for upgrades from the pre-minio version.

Adds validation to e2e tests for data paths.

```release-note
Fixes an issue that causes the data paths to be set incorrectly for seaweedfs when cluster is converted to HA.
```

### DIFF
--- a/e2e/cluster/docker/cluster.go
+++ b/e2e/cluster/docker/cluster.go
@@ -82,10 +82,9 @@ func (c *Cluster) WaitForReady() {
 }
 
 func (c *Cluster) Cleanup(envs ...map[string]string) {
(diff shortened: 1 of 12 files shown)


emosbaugh:
TestFiveNodesAirgapUpgrade failure is a flake and unrelated to the changes. Will merge through this one.

laverya (approved):

PR #1350, opened 2024-10-21
chore: unit test returned supernet

ricardomaraschini:

as per pr comment this makes sure we are also validating the returned supernet.

ref: https://github.com/replicatedhq/embedded-cluster/pull/1324#discussion_r1806661271

### DIFF
--- a/e2e/install_test.go
+++ b/e2e/install_test.go
@@ -996,7 +996,7 @@ func TestSingleNodeAirgapUpgradeCustomCIDR(t *testing.T) {
 	// expected to be in the 192.168.0.0/17 range while services are in the
 	// 192.168.128.0/17 range. i.e. pods are in 192.168.0-127.* while
 	// services in 192.168.128-254.*.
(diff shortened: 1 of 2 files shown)


e2e/install_test.go
```diff
@@ -996,7 +996,7 @@ func TestSingleNodeAirgapUpgradeCustomCIDR(t *testing.T) {
 	// expected to be in the 192.168.0.0/17 range while services are in the
 	// 192.168.128.0/17 range. i.e. pods are in 192.168.0-127.* while
 	// services in 192.168.128-254.*.
-	podregex := `192\.168\.\([0-9]\|[1-9][0-9]\|12[0-7]\)\.`
+	podregex := `192\.168\.\([0-9]\|[1-9][0-9]\|12[0-7]\|1[0-1][0-9]\)\.`
```
ricardomaraschini, line 999: Can we somehow simplify this ?

laverya: my suggestion for simplifying this is "use a /15 so that each sub range is a /16, and you only have to match the first two numbers in the quad"

laverya (approved):

PR #1352, opened 2024-10-21
chore: Collect LAM service config in support bundle

banjoh:

Collect Local Artefact Mirror systemd service config in support bundle

NONE

```release-note
Collect local artefact mirror service config in support bundle
```

NONE

### DIFF
--- a/operator/charts/embedded-cluster-operator/templates/embedded-cluster-lam-service-config.yaml
+++ b/operator/charts/embedded-cluster-operator/templates/embedded-cluster-lam-service-config.yaml
@@ -1,0 +1,48 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: embedded-cluster-lam-service-config
+  labels:
+    troubleshoot.sh/kind: support-bundle
+    {{- with (include "embedded-cluster-operator.labels" $ | fromYaml) }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+data:
+  support-bundle-spec: |
+    apiVersion: troubleshoot.sh/v1beta2
+    kind: SupportBundle
+    metadata:
+      name: embedded-cluster-lam-service-config
+      labels:
+        troubleshoot.sh/kind: support-bundle
+    spec:
+      collectors:
+        - runDaemonSet:
+            name: "local-artifact-mirror-service-config"
+            namespace: embedded-cluster
+            podSpec:
+              containers:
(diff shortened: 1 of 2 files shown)


pkg/goods/support/host-support-bundle.tmpl.yaml
```diff
@@ -153,6 +153,9 @@ spec:
       collectorName: 'check-umount'
       command: 'sh'
       args: ['-c', 'command -v umount']
+  - copy:
+      collectorName: installer/lam-service-config
+      path: /etc/systemd/system/local-artifact-mirror.service.d/embedded-cluster.conf
```
emosbaugh, line 158: does this support globs? is this possible?

```suggestion
      path: /etc/systemd/system/local-artifact-mirror.service.d/*
```

operator/charts/embedded-cluster-operator/templates/embedded-cluster-lam-service-config.yaml
```diff
@@ -0,0 +1,48 @@
...
+    spec:
+      collectors:
+        - runDaemonSet:
+            name: "local-artifact-mirror-service-config"
+            namespace: embedded-cluster
+            podSpec:
+              containers:
+              - image: {{ .Values.utilsImage }}
+                imagePullPolicy: Always
+                args: ["chroot","/host","cat","/etc/systemd/system/local-artifact-mirror.service.d/embedded-cluster.conf"]
```
emosbaugh, line 27: would it be possible to collect all files in this directory?

pkg/goods/support/host-support-bundle.tmpl.yaml
laverya: I think just a `/` is enough, no need for the `/*`

laverya: https://troubleshoot.sh/docs/host-collect-analyze/copy/ `*` is in the logs though, so yeah let's go with that

operator/charts/embedded-cluster-operator/templates/embedded-cluster-lam-service-config.yaml
banjoh: Something like `"/host","sh", "-c", "cat /etc/systemd/system/local-artifact-mirror.service.d/*"` will work. All files would be concatenated into one file. That might be confusing. I think its better to collect one file at a time

emosbaugh (approved):

PR #1356, opened 2024-10-21
fix: reset fails on worker nodes after an upgrade

emosbaugh:

Reset is failing on worker nodes after an upgrade because the cli is preferring the default data directory over the one discovered from the filesystem.

```release-note

```

### DIFF
--- a/cmd/embedded-cluster/flags.go
+++ b/cmd/embedded-cluster/flags.go
@@ -58,11 +58,27 @@ func getLocalArtifactMirrorPortFlag(runtimeConfig *ecv1beta1.RuntimeConfigSpec) 
 	}
 }
 
-func getDataDirFlag(runtimeConfig *ecv1beta1.RuntimeConfigSpec) cli.Flag {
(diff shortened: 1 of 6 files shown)


cmd/embedded-cluster/flags.go
```diff
@@ -58,7 +58,7 @@ func getLocalArtifactMirrorPortFlag(runtimeConfig *ecv1beta1.RuntimeConfigSpec)
 	}
 }
 
-func getDataDirFlag(runtimeConfig *ecv1beta1.RuntimeConfigSpec) cli.Flag {
+func getInstallDataDirFlag(runtimeConfig *ecv1beta1.RuntimeConfigSpec) *cli.StringFlag {
```
sgalsaleh, line 61: how about:
```suggestion
func getDataDirFlagWithDefault(runtimeConfig *ecv1beta1.RuntimeConfigSpec) *cli.StringFlag {
```
since it's used for other commands and not just the install command?

sgalsaleh (approved):

laverya (approved):

PR #1358, opened 2024-10-22
store runtime config in /etc/embedded-cluster/ec.yaml

laverya:

This allows the use of `shell`, `reset` and `support-bundle` on instances with custom EC data directories without having to pass in a `--data-dir` argument.

```release-note

```


pkg/configutils/runtime.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+func WriteRuntimeConfig(spec *v1beta1.RuntimeConfigSpec) error {
+	if spec == nil {
+		return nil
+	}
+
+	location := defaults.PathToECConfig()
+
+	err := os.MkdirAll(filepath.Dir(location), 0700)
```
emosbaugh, line 21: why so strict? how about 755?

pkg/configutils/runtime.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to remove existing runtime config: %w", err)
+	}
+
+	yml, err := yaml.Marshal(spec)
+	if err != nil {
+		return fmt.Errorf("unable to marshal runtime config: %w", err)
+	}
+
+	err = os.WriteFile(location, yml, 0600)
```
emosbaugh, line 37: why so strict? how about 644?

```
ethan@ethanm-ec-1:~$ ls -ld /etc/k0s/
drwxr-xr-x 3 root root 4096 Oct 22 20:09 /etc/k0s/
ethan@ethanm-ec-1:~$ ls -l /etc/k0s/
total 8
drwxr-xr-x 2 root root 4096 Oct 22 20:09 containerd.d
-rw-r--r-- 1 root root  362 Oct 22 20:09 containerd.toml
```

emosbaugh (changes requested):

emosbaugh (approved):

PR #1370, opened 2024-10-23
chore: bump troubleshoot to allow optional cluster specs

ricardomaraschini:

latest version of troubleshoot now supports bypassing errors when reading support bundle specs from the cluster (https://github.com/replicatedhq/troubleshoot/pull/1660). in the previous version it used to fail.

this pr also makes the `embedded-cluster support-bundle` command use the kubeconfig as an environment variable.

```release-note

```

### DIFF
--- a/cmd/embedded-cluster/support_bundle.go
+++ b/cmd/embedded-cluster/support_bundle.go
@@ -34,6 +34,9 @@ func supportBundleCommand() *cli.Command {
 
 			kubeConfig := provider.PathToKubeConfig()
 			hostSupportBundle := provider.PathToEmbeddedClusterSupportFile("host-support-bundle.yaml")
+			if _, err := os.Stat(hostSupportBundle); err != nil {
+				return fmt.Errorf("unable to find host support bundle: %w", err)
+			}
 
 			spin := spinner.Start()
 			spin.Infof("Collecting support bundle (this may take a while)")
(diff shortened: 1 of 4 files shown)


hedge-sparrow (approved):

cmd/embedded-cluster/support_bundle.go
```diff
@@ -45,10 +48,10 @@ func supportBundleCommand() *cli.Command {
 					Writer:       stdout,
 					ErrWriter:    stderr,
 					LogOnSuccess: true,
+					Env:          map[string]string{"KUBECONFIG": kubeConfig},
```
marccampbell, line 51: What's the reason for changing this from the `--kubeconfig` flag to passing it as an env var? Its likely that both are valid options, but I'm unclear on the reason we are changing this.

ricardomaraschini: It seems like `--kubeconfig` flag denotes intention so it fails if the file does not exist.

```
root@divine-sea:/home/developer# ./support-bundle --kubeconfig /does/not/exist ./spec.yaml
Error: failed to convert kube flags to rest config: stat /does/not/exist: no such file or directory
root@divine-sea:/home/developer# KUBECONFIG=/does/not/exist ./support-bundle ./spec.yaml


A support bundle was generated and saved at support-bundle-2024-10-24T13_29_49.tar.gz. Please send this file to your software vendor for support.
root@divine-sea:/home/developer#
```

PR #1371, opened 2024-10-23
feat: configure firewalld if it is running

ricardomaraschini:

starts to configure firewalld if it is running. this pr makes the installer to create a zone with 'accept' target for packets related to the pod cidr and the service cidr.

```release-note
Added firewalld auto configuration.
```


ricardomaraschini:
We have decided that this is not the path we want to follow. We want to preflight this and then ask the administrator to do the changes themselves.

PR #1372, opened 2024-10-23
instead of hardcoding values hashes in unit testing build them from the expected values

laverya:

This allows changing chart values without also having to manually recalculate the desired values hash.

```release-note

```


laverya:
merging without waiting for integration tests because this was only a change to a unit test file, and unit tests passed

PR #1377, opened 2024-10-23
fix mismatch between admin console and embedded cluster versions

laverya:

```release-note

```


laverya:
This is to fix an error message
```
{"level":"info","ts":"2024-10-23T16:44:28Z","msg":"deployment has embedded cluster version (v1.16.0+k8s-1.29) which does not match current embedded cluster version (1.16.0+k8s-1.29). will not process..."}
```
or at least I think it will

emosbaugh (dismissed):

emosbaugh (approved):

PR #1398, opened 2024-10-24
add more collectors and analyzers

adamancini:

Add more collectors for system resources, logs, and network

https://app.shortcut.com/replicated/story/111997/update-host-support-bundle-spec-with-our-go-to-info-for-troubleshooting

Not sure, probably not

```release-note
NONE

```

NONE

### DIFF
--- a/pkg/goods/support/host-support-bundle.tmpl.yaml
+++ b/pkg/goods/support/host-support-bundle.tmpl.yaml
@@ -74,6 +74,23 @@ spec:
       command: top
       args: ['-b', '-n', '1']
   - run:
+      collectorName: uptime
+      command: uptime
+      args: []
+  - run:
+      collectorName: uname
+      command: uname
+      args: ['-a']
+  - run:
+      collectorName: "hostnames"
+      command: "sh"
+      args:
+        - -c
+        - |
+          echo "hostname = $(hostname)"
+          echo "/proc/sys/kernel/hostname = $(cat /proc/sys/kernel/hostname)"
+          echo "uname -n = $(uname -n)"
+  - run:
       collectorName: df
       command: df
       args: ['-h']
@@ -88,6 +105,82 @@ spec:
   - run:
       collectorName: uptime
       command: uptime
+  - run:
+      collectorName: sestatus
+      command: sestatus
+      args: []
+  - run:
+      collectorName: apparmor-status
+      command: apparmor_status
+      args: []
+  - run:
+      collectorName: iptables
+      command: iptables
+      args: ["-L", "-v"]
+  - run:
+      collectorName: iptables-version
+      command: iptables
+      args: ["--version"]
+  - run:
+      collectorName: nftables-list
+      command: nft
+      args: ["list", "table", "filter"]
+  - run:
+        collectorName: "ipvsadm"
+        command: "ipvsadm"
+        args: ["-l", "-n"]
+  - run:
+      collectorName: "lsblk"
+      command: "lsblk"
+      args: ["--fs"]
+  - run:
+      collectorName: lvm
+      command: pvdisplay
+      args: []
+  - run:
+      collectorName: lvm
+      command: vgdisplay
+      args: []
+  - run:
+      collectorName: lvm
+      command: lvdisplay
+      args: []
+  - run:
+      collectorName: "netstat-ports"
+      command: "netstat"
+      args: ["-t", "-u", "-l", "-p", "-n"]
+  - run:
+      collectorName: "netstat-route-table"
+      command: "netstat"
+      args: ["-r", "-n"]
+  - run:
+      collectorName: "resolvectl-status"
+      command: "resolvectl"
+      args: ["status"]
+  - run:
+      collectorName: "resolv-conf"
+      command: "cat"
+      args: ["/etc/resolv.conf"]
+  - run:
+      collectorName: "systemd-resolved-conf"
+      command: "cat"
+      args: ["/etc/systemd/resolved.conf"]
+  - run:
+      collectorName: "nsswitch-conf"
+      command: "cat"
+      args: ["/etc/nsswitch.conf"]
+  - run:
+      collectorName: "hosts"
+      command: "cat"
+      args: ["/etc/hosts"]
+  - run:
+      collectorName: "ip-route-table"
+      command: "ip"
+      args: ["route"]
+  - run:
+      collectorName: "sysctl"
+      command: "sysctl"
+      args: ["-a"]
   - run:
       collectorName: k0s-version
       command: /usr/local/bin/k0s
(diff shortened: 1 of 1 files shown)


pkg/goods/support/host-support-bundle.tmpl.yaml
```diff
@@ -448,3 +606,38 @@ spec:
...
+      fileName: host-collectors/run-host/sestatus.txt
+      regex: '(?m)^Current mode:\s+enforcing'
+      ignoreIfNoFiles: true
+      outcomes:
+        - fail:
+            when: "true"
+            message: "SELinux is enabled when it should be disabled for kubernetes to work properly"
+        - pass:
+            when: "false"
+            message: "SELinux is disabled as expected"
```
chris-sanders, line 631: Doesn't k0s ship selinux config? Are we sure this one applies to EC?

pkg/goods/support/host-support-bundle.tmpl.yaml
```diff
@@ -448,3 +606,38 @@ spec:
...
             when: "false"
             message: "No signs of hostname changes found"
+  - textAnalyze:
+      checkName: Check if localhost resolves to 127.0.0.1
+      fileName: host-collectors/run-host/localhost-ips.txt
+      regex: 'localhost has address 127.0.0.1'
+      outcomes:
+        - fail:
+            when: "false"
+            message: "'localhost' does not resolve to 127.0.0.1 ip address"
```
chris-sanders, line 616: I think we need to make a pass to make these match the new EC style of more informative action oriented messages.

Something like: localhost doesn't resovle to 127.0.0.1 add <hostname> to /etc/hosts to resolve.

pkg/goods/support/host-support-bundle.tmpl.yaml
```diff
@@ -156,9 +284,39 @@ spec:
...
+      fileSize: 22Mi
+      operationSizeBytes: 2300
+      datasync: true
+      enableBackgroundIOPS: true
+      backgroundIOPSWarmupSeconds: 10
+      backgroundWriteIOPS: 300
+      backgroundWriteIOPSJobs: 6
+      backgroundReadIOPS: 50
+      backgroundReadIOPSJobs: 1
+      exclude: true
```
chris-sanders, line 315: This isn't the right folder for sure, we should also check if this is the most recent test. I think EC already tests filesystem performance now. Is this the same as the Preflight?

At the very least it's not a two-minute-benchamark so I'm pretty sure you've lifted the old one.

pkg/goods/support/host-support-bundle.tmpl.yaml
```diff
@@ -73,6 +73,23 @@ spec:
       collectorName: top
       command: top
       args: ['-b', '-n', '1']
+  - run:
+      collectorName: uptime
+      command: uptime
+      args: []
```
chris-sanders, line 79: I think this is a duplicate it's on line 93 as well.

pkg/goods/support/host-support-bundle.tmpl.yaml
```diff
@@ -448,3 +606,38 @@ spec:
...
+            message: "SELinux is disabled as expected"
+  - textAnalyze:
+      checkName: "Detect Threat Management and Network Security Tools"
+      fileName: host-collectors/run-host/ps-detect-antivirus-and-security-tools.txt
+      regex: '\b(clamav|sophos|esets_daemon|fsav|symantec|mfend|ds_agent|kav|bdagent|s1agent|falcon|illumio|xagt)\b'
+      ignoreIfNoFiles: true
+      outcomes:
+        - fail:
+            when: "true"
+            message: "Antivirus or Network Security tools detected. These tools can interfere with kubernetes operation."
```
chris-sanders, line 640: I think we need to think about the messaging on this one too. Do we want to always tell them to turn these off? If so we should instead say to disable the tool (and reboot?) to resolve this issue.

If however, we're not _sure_ it results in a failure I'm not sure we should have an analyzer for it. Although I think I'm ok with just telling people to turn it off and if they ignore it they hopefully know what they are doing.

chris-sanders (changes requested):
Be sure @ajp-io takes a look at the pass/fail messages as well.
A few things I noticed for updating.

pkg/goods/support/host-support-bundle.tmpl.yaml
adamancini: I don't know, but maybe alex can chime in - I would still like to know the current state if it got changed after-the-fact.  thinking along the same lines we should always check ip_forwarding is enabled even if we expect k0s to turn it on

pkg/goods/support/host-support-bundle.tmpl.yaml
adamancini: yeah this is a lift from the old one - will update the data directory, good catch

chris-sanders: I believe this is the new one: https://github.com/replicatedhq/embedded-cluster/blob/cf9f38c56e978ee704e6d9b5e21c1dd9b564d88a/pkg/preflights/host-preflight.yaml#L429

pkg/goods/support/host-support-bundle.tmpl.yaml
chris-sanders: Oh nevermind: https://docs.k0sproject.io/v1.31.1+k0s.1/selinux/#set-selinux-labels-for-k0s-installation-files
We don't do those things so it's not configured go ahead and leave this check.

Just need to update the message. "SELinux is not supported, disable it with <insert command here> to proceed."

adamancini: done

pkg/goods/support/host-support-bundle.tmpl.yaml
adamancini: done

pkg/goods/support/host-support-bundle.tmpl.yaml
adamancini: update with new language

pkg/goods/support/host-support-bundle.tmpl.yaml
adamancini: done

pkg/goods/support/host-support-bundle.tmpl.yaml
adamancini: good catch

pkg/goods/support/host-support-bundle.tmpl.yaml
```diff
@@ -448,3 +596,47 @@ spec:
...
+      fileName: host-collectors/run-host/sestatus.txt
+      regex: '(?m)^Current mode:\s+enforcing'
+      ignoreIfNoFiles: true
+      outcomes:
+        - fail:
+            when: "true"
+            message: "SELinux enforcement is not supported, disable it or set to permissive mode to proceed"
+        - pass:
+            when: "false"
+            message: "SELinux is disabled as expected"
```
ajp-io, line 621: I think we tested and an install worked with SELinux enabled. I'd double check with the team though.

adamancini: interesting; TIL

ajp-io: From standup: let's continue to collect this, but let's not analyze it since we think it works with SELinux at this point.

laverya (approved):
This looks good to me if we remove the SELinux failure
If there was a 'display this' level below 'warn' I would love it there but I don't think that exists

laverya (approved):

PR #1399, opened 2024-10-24
feat: pass --config-values to kots during embedded cluster install

laverya:

```release-note
adds `--config-values` to the install command, allowing automated installations.
```


cmd/embedded-cluster/install.go
```diff
@@ -703,6 +703,10 @@ func installCommand() *cli.Command {
 					Usage: "Skip host preflight checks. This is not recommended.",
 					Value: false,
 				},
+				&cli.StringFlag{
+					Name:  "config-values",
+					Usage: "path to a manifest containing config values (must be apiVersion: kots.io/v1beta1, kind: ConfigValues)path to a manifest containing config values (must be apiVersion: kots.io/v1beta1, kind: ConfigValues)",
```
marccampbell, line 708: this looks like a copy/paste error

laverya: it is, I'll fix it

pkg/configutils/kots.go
```diff
@@ -0,0 +1,35 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+type gvk struct {
+	ApiVersion string `yaml:"apiVersion"`
+	Kind       string `yaml:"kind"`
+}
+
+// ValidateKotsConfigValues checks if the file exists and has the 'kots.io/v1beta1 ConfigValues' GVK
+func ValidateKotsConfigValues(filename string) error {
```
marccampbell, line 15: Is there a function in KOTS that does this?

e2e/install_test.go
```diff
@@ -2255,3 +2257,97 @@ func TestInstallWithPrivateCAs(t *testing.T) {
 
 	t.Logf("%s: test complete", time.Now().Format(time.RFC3339))
 }
+
+func TestInstallWithConfigValues(t *testing.T) {
```
marccampbell, line 2261: Is there a way to test this without e2e?

Makefile
```diff
@@ -7,7 +7,7 @@ ARCH ?= $(shell go env GOARCH)
 
 APP_NAME = embedded-cluster
 ADMIN_CONSOLE_CHART_REPO_OVERRIDE =
-ADMIN_CONSOLE_IMAGE_OVERRIDE =
+ADMIN_CONSOLE_IMAGE_OVERRIDE = ttl.sh/automated-11614003868/kotsadm:24h
```
marccampbell, line 10: This probably shouldn't be commited

operator/pkg/artifacts/registry.go
```diff
@@ -57,7 +57,7 @@ func EnsureRegistrySecretInECNamespace(ctx context.Context, cli client.Client, i
 			}
 		}
 
-		obj.ObjectMeta.Labels = applyECOperatorLabels(obj.ObjectMeta.Labels, "upgrader")
+		obj.ObjectMeta.Labels = applyECOperatorLabels(obj.ObjectMeta.Labels, "upgrader-registry-secret")
```
marccampbell, line 60: why is this part of this PR?

pkg/cmd/install.go
```diff
@@ -894,6 +898,16 @@ func getAddonsApplier(c *cli.Context, runtimeConfig *ecv1beta1.RuntimeConfigSpec
 	if adminConsolePwd != "" {
 		opts = append(opts, addons.WithAdminConsolePassword(adminConsolePwd))
 	}
+
+	if c.String("config-values") != "" {
+		err := configutils.ValidateKotsConfigValues(c.String("config-values"))
+		if err != nil {
+			return nil, fmt.Errorf("unable to validate config values file: %w", err)
```
marccampbell, line 905: does this turn into the CLI output someone sees if they pass malformed config?

e2e/install_test.go
laverya: not that I'm aware of, because the thing being tested is "does kots work in this scenario" (and considering that it panics without code changes that is a very relevant thing to test)

Makefile
laverya: I need to test with the updated kots image before I merge the kots PR, though
and I would prefer to make sure it works in CI as well as in local dev

laverya: it won't get _merged_, certainly!

pkg/cmd/install.go
laverya: yes

operator/pkg/artifacts/registry.go
laverya: it probably shouldn't be
I was looking through this while trying to figure out the labels for the apply-upgrade job and discovered that we were using the same 'name' label for two different jobs

pkg/configutils/kots.go
sgalsaleh: we can probably make it a hidden kots cli command

e2e/install_test.go
sgalsaleh: hmm, i think you can use:
- the new dryrun integration testing framework to verify that the config values get passed down to the kots cli correctly
- write a unit or integration test in kots for determining the next app status
?

it is a bit more effort

laverya: It's a bit more effort and doesn't actually cover the code flow in question
I can add an integration test to make sure the values are getting passed properly, but adding a test in kots that covers this flow is a lot harder - the function in question is not unit tested there and we do not have a integration test that uses embedded cluster

tests/dryrun/install_test.go
```diff
@@ -242,3 +245,53 @@ func TestCustomPortsInstallation(t *testing.T) {
 
 	t.Logf("%s: test complete", time.Now().Format(time.RFC3339))
 }
+
+func TestConfigValuesInstallation(t *testing.T) {
+	tmpdir, err := os.MkdirTemp("", "embedded-cluster-test")
```
sgalsaleh, line 250: i would suggest using `t.TempDir()`

tests/dryrun/install_test.go
```diff
@@ -242,3 +245,53 @@ func TestCustomPortsInstallation(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("failed to create temp dir: %v", err)
+	}
+	defer os.RemoveAll(tmpdir)
+	tempfile := filepath.Join(tmpdir, "values.yaml")
+	os.WriteFile(tempfile, []byte(`
```
sgalsaleh, line 256: let's move this to a yaml file under the `assets` directory and use go embed similar to the other tests

laverya: done

e2e/install_test.go
laverya: I've added this integration test, but it doesn't cover much (just that we're actually passing the flag to kots)

sgalsaleh (approved):

PR #1405, opened 2024-10-25
chore: add private-ca from flag to RuntimeConfigSpec

emosbaugh:

Required for https://github.com/replicatedhq/embedded-cluster/pull/1318

```release-note

```


cmd/embedded-cluster/restore.go
```diff
@@ -933,6 +933,7 @@ func restoreCommand() *cli.Command {
 					Usage: "Disable interactive prompts.",
 					Value: false,
 				},
+				getPrivateCAsFlag(runtimeConfig),
```
emosbaugh, line 936: is this needed for the initial install? it will get restored from the backup.

laverya: Can we access the backup without the private CAs? Not always

emosbaugh: That is what i assume so it will be needed to contact replicated app etc...

kinds/apis/v1beta1/runtimeconfig_types.go
```diff
@@ -27,6 +27,9 @@ type RuntimeConfigSpec struct {
 	AdminConsole AdminConsoleSpec `json:"adminConsole,omitempty"`
 	// LocalArtifactMirrorPort holds the Local Artifact Mirror configuration.
 	LocalArtifactMirror LocalArtifactMirrorSpec `json:"localArtifactMirror,omitempty"`
+
+	// Private CAs holds a list of paths to trusted private CA certificate PEM files.
+	PrivateCAs []string `json:"privateCAs,omitempty"`
```
laverya, line 32: this can't be a list of paths, it has to be the files themselves

emosbaugh: fixed. these are no longer used on join

PR #1435, opened 2024-10-31
Add additional namespaces for collection

chris-sanders:
The support bundles doesn't include some namespaces that are necessary to debug some failure condistion.


laverya (approved):
Should we add kotsadm here too?

chris-sanders:
The only reason I didn't add `kotsadm` is I felt like kots _should_ be adding that? So I think we should collect it but shouldn't KOTS be adding that to it's bundle already?

laverya (approved):

PR #1449, opened 2024-11-05
feat: add preflight for ip forwarding

ricardomaraschini:

we require users to manually enable ip forwarding on their nodes before installing. this preflight checks if ip forwarding is enabled and fails if it is not.

```release-note
Added preflight for IP forwarding.
```


ricardomaraschini:
@JGAntunes I guess there is overlap between what you are doing and what I am doing. Would you mind reviewing this ?

ajp-io (approved):

PR #1453, opened 2024-11-05
fix: unable to upgrade EC when introducing a required config item w/o…

sgalsaleh:

Fixes an issue where updating the app failed if the new app version contained a required config item without a default or value.

[SC-111215](https://app.shortcut.com/replicated/story/111215)

Yes

```release-note
* Fixes an issue where updating the app failed if the new app version contained a required config item without a default or value.
```

NONE

### DIFF
--- a/scripts/ci-upload-binaries.sh
+++ b/scripts/ci-upload-binaries.sh
@@ -96,8 +96,11 @@ function kotsbin() {
     local kots_version=
     kots_version=$(make print-KOTS_VERSION)
 
-    local kots_override=
-    kots_override=$(make print-KOTS_BINARY_URL_OVERRIDE)
+    local kots_url_override=
+    kots_url_override=$(make print-KOTS_BINARY_URL_OVERRIDE)
+
+    local kots_file_override=
+    kots_file_override=$(make print-KOTS_BINARY_FILE_OVERRIDE)
 
     # check if the binary already exists in the bucket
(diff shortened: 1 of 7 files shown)


laverya (approved):

scripts/ci-upload-binaries.sh
```diff
@@ -96,8 +96,11 @@ function kotsbin() {
...
 
-    local kots_override=
-    kots_override=$(make print-KOTS_BINARY_URL_OVERRIDE)
+    local kots_url_override=
+    kots_url_override=$(make print-KOTS_BINARY_URL_OVERRIDE)
+
+    local kots_file_override=
+    kots_file_override=$(make print-KOTS_BINARY_FILE_OVERRIDE)
 
     # check if the binary already exists in the bucket
```
emosbaugh, line 105: shouldnt you skip this check if there is a binary override url?

emosbaugh (dismissed):

scripts/ci-upload-binaries.sh
sgalsaleh: i didn't wanna touch the url override functionality as that can be used in CI. i only added support for file override from the filesystem and that should still check if the binary exists because the kots version will be the sha of the binary file and should not re-upload if already exists

emosbaugh (approved):

PR #1458, opened 2024-11-06
fix(ci): retry wait for new app pods to be running

emosbaugh:

Retry evaluates the expression before the pipe

```release-note

```


laverya (approved):

laverya:
there's probably more than one of these that needs to be fixed

PR #1461, opened 2024-11-06
fix: fail to discover azure standard sku public ip address

emosbaugh:

Cant retrieve standard sku address from instance metadata

https://learn.microsoft.com/en-us/answers/questions/7932/public-ip-not-available-via-metadata

```
azureuser@ethan-1:~$ sudo ./embedded-cluster-smoke-test-staging-app install --license license.yaml --skip-host-preflights
? Set the Admin Console password (minimum 6 characters): ********
? Confirm the Admin Console password: ********
✔  Host files materialized!
✔  Host preflights skipped
✔  Node installation finished!
✔  Storage is ready!
✔  Embedded Cluster Operator is ready!
✔  Admin Console is ready!
✔  Additional components are ready!
Visit the Admin Console to configure and install embedded-cluster-smoke-test-staging-app: http://52.249.223.56:30000
```

```release-note
Fixes an issue when installing in Azure that caused the install command to output the instance private address rather than the public one in the link to the Admin Console.
```

### DIFF
--- a/pkg/defaults/cloudprovider.go
+++ b/pkg/defaults/cloudprovider.go
@@ -24,31 +24,37 @@ func init() {
 // string is returned.
 func TryDiscoverPublicIP() string {
 	if !shouldUseMetadataService() {
+		logrus.Debug("No cloud provider metadata service found, skipping public IP discovery")
 		return ""
 	}
 
 	publicIP := tryDiscoverPublicIPAWSIMDSv2()
 	if publicIP != "" {
+		logrus.Debugf("Found public IP %s using AWS IMDSv2", publicIP)
 		return publicIP
 	}
 
 	publicIP = tryDiscoverPublicIPAWSIMDSv1()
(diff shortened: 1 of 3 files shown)


pkg/defaults/cloudprovider.go
```diff
@@ -41,12 +42,16 @@ func TryDiscoverPublicIP() string {
...
+	if publicIP != "" {
+		return publicIP
+	}
+
 	publicIP = tryDiscoverPublicIPAzure()
 	if publicIP != "" {
 		return publicIP
 	}
 
-	// If we reach this point, we failed to discover the public IP.
```
laverya, line 49: seems to me like this comment is still accurate?

emosbaugh: it felt like a pointless comment that ai wrote

sgalsaleh (approved):

PR #1484, opened 2024-11-11
feat: enable ip_forward, disable arp_ignore arp_filter

ricardomaraschini:

we are now enabling ip_forward on the node, this is required for the embedded-cluster to work properly. we are also disabling arp_ignore and arp_filter to make sure the system is prepared for calico.

```release-note
Auto configure system parameters for ip_forward, arp_filter and arp_ignore
```

We need to document that the installer now requires the `sysctl` command to be present on the target system. This command is found in the `procps` (Debian/Ubuntu) or `procps-ng` (Rocky Linux / RHEL) package.

### DIFF
--- a/pkg/configutils/runtime.go
+++ b/pkg/configutils/runtime.go
@@ -3,12 +3,14 @@ package configutils
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 
 	"github.com/replicatedhq/embedded-cluster/kinds/apis/v1beta1"
 	"github.com/replicatedhq/embedded-cluster/pkg/defaults"
 	"github.com/replicatedhq/embedded-cluster/pkg/goods"
 	"github.com/replicatedhq/embedded-cluster/pkg/helpers"
+	"github.com/sirupsen/logrus"
 	"sigs.k8s.io/yaml"
 )
 
@@ -60,15 +62,22 @@ func ReadRuntimeConfig() (*v1beta1.RuntimeConfigSpec, error) {
 }
 
 // ConfigureSysctl writes the sysctl config file for the embedded cluster and
-// reloads the sysctl configuration.
+// reloads the sysctl configuration. This function has a distinct behavior: if
+// the sysctl binary does not exist it returns an error but if it fails to lay
+// down the sysctl config on disk it simply returns nil.
 func ConfigureSysctl(provider *defaults.Provider) error {
+	if _, err := exec.LookPath("sysctl"); err != nil {
+		return fmt.Errorf("unable to find sysctl binary: %w", err)
+	}
+
 	materializer := goods.NewMaterializer(provider)
 	if err := materializer.SysctlConfig(); err != nil {
-		return fmt.Errorf("unable to materialize sysctl config: %w", err)
+		logrus.Debugf("unable to materialize sysctl config: %v", err)
+		return nil
 	}
 
 	if _, err := helpers.RunCommand("sysctl", "--system"); err != nil {
-		return fmt.Errorf("unable to configure sysctl: %w", err)
+		logrus.Debugf("unable to configure sysctl: %v", err)
 	}
 	return nil
 }
--- a/pkg/configutils/runtime.go
+++ b/pkg/configutils/runtime.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/replicatedhq/embedded-cluster/kinds/apis/v1beta1"
 	"github.com/replicatedhq/embedded-cluster/pkg/defaults"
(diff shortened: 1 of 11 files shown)


JGAntunes:
> We need to document that the installer now requires the sysctl command to be present on the target system. This command is found in the procps (Debian/Ubuntu) or procps-ng (Rocky Linux / RHEL) package.

Unsure if it's a better alternative but I looked over at - https://github.com/lorenzosaino/go-sysctl - while working on the troubleshoot implementation for `sysctl` checks. We ended up not going with it because of support for Darwin, but I wonder if it would make sense for embedded cluster since we're guaranteed(?) that the system will always have `procfs` mounted (the only requirement to use the lib).

pkg/goods/static/99-embedded-cluster.conf
```diff
@@ -0,0 +1,9 @@
+# this entry enables ip forwarding. this feature is necessary as embedded
+# cluster creates virtual network interfaces and need the traffic among them to
+# be forwarded.
+net.ipv4.ip_forward = 1
+
+# arp filter and ignore need to be disabled otherwise we can't have arp
+# resolving across the calico network interfaces.
+net.ipv4.conf.default.arp_filter = 0
+net.ipv4.conf.default.arp_ignore = 0
```
JGAntunes, line 9: We need to do the same for `net.ipv4.conf.all.*` too:
- https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html

> The max value from conf/{all,interface}/arp_ignore is used when ARP request is received on the {interface}

> arp_filter for the interface will be enabled if at least one of conf/{all,interface}/arp_filter is set to TRUE, it will be disabled otherwise

JGAntunes (commented):
Left a comment on the `arp` settings I think we need to address.

Wondering if we should add a test/assertion for install/join/restore to validate the file is materialized and that is deleted on reset?

ricardomaraschini:
> > We need to document that the installer now requires the sysctl command to be present on the target system. This command is found in the procps (Debian/Ubuntu) or procps-ng (Rocky Linux / RHEL) package.
> 
> Unsure if it's a better alternative but I looked over at - https://github.com/lorenzosaino/go-sysctl - while working on the troubleshoot implementation for `sysctl` checks. We ended up not going with it because of support for Darwin, but I wonder if it would make sense for embedded cluster since we're guaranteed(?) that the system will always have `procfs` mounted (the only requirement to use the lib).

How do we persist the config (across reboots) using the library ?

pkg/goods/static/99-embedded-cluster.conf
ricardomaraschini: Makes sense to me. Addressed.

JGAntunes:
> How do we persist the config (across reboots) using the library ?

Yeah I don't think it's suited for this, there's - https://github.com/lorenzosaino/go-sysctl/blob/f1dde0a7a4859d95ae07ecd6dbe40561e6929c12/client.go#L131-L145 - but you would still need to run it on every reboot somehow

pkg/configutils/runtime.go
```diff
@@ -57,3 +60,24 @@ func ReadRuntimeConfig() (*v1beta1.RuntimeConfigSpec, error) {
 
 	return &spec, nil
 }
+
+// ConfigureSysctl writes the sysctl config file for the embedded cluster and
+// reloads the sysctl configuration. This function has a distinct behavior: if
+// the sysctl binary does not exist it returns an error but if it fails to lay
+// down the sysctl config on disk it simply returns nil.
+func ConfigureSysctl(provider *defaults.Provider) error {
```
JGAntunes, line 68: Would it be worth adding a unit test for this method?

pkg/goods/materializer.go
```diff
@@ -76,6 +76,19 @@ func (m *Materializer) CalicoNetworkManagerConfig() error {
 	return nil
 }
 
+// SysctlConfig writes the embedded sysctl config to the /etc/sysctl.d directory.
+func (m *Materializer) SysctlConfig() error {
```
JGAntunes, line 80: Would it be worth adding a unit test for this method?

JGAntunes (commented):
Overall lgtm, I wonder if we would like to add a couple of unit tests to the newly added logic?

pkg/configutils/runtime.go
ricardomaraschini: Added.

pkg/goods/materializer.go
ricardomaraschini: Added.

ricardomaraschini:
@JGAntunes Please review it once more.

JGAntunes (approved):

PR #1490, opened 2024-11-13
feat: prompt in the install command if newer app release version is available

emosbaugh:

When installing, the user could be unaware that there is a newer version since downloading the release archive. This will prompt to confirm that this is intended.

Screenshot 2024-11-13 at 9 26 34 AM

```release-note
A prompt has been added to the install command to confirm whether or not to continue when a newer version is available of the application.
```

### DIFF
--- a/pkg/cmd/install.go
+++ b/pkg/cmd/install.go
@@ -288,7 +288,12 @@ func isAlreadyInstalled() (bool, error) {
 	}
 }
 
-func getLicenseFromFilepath(licenseFile string, rel *release.ChannelRelease) (*kotsv1beta1.License, error) {
+func getLicenseFromFilepath(licenseFile string) (*kotsv1beta1.License, error) {
+	rel, err := release.GetChannelRelease()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get release from binary: %w", err) // this should only be if the release is malformed
+	}
+
 	// handle the three cases that do not require parsing the license file
 	// 1. no release and no license, which is OK
 	// 2. no license and a release, which is not OK
@@ -338,7 +343,11 @@ func getLicenseFromFilepath(licenseFile string, rel *release.ChannelRelease) (*k
 	return license, nil
 }
 
-func checkAirgapMatches(c *cli.Context, rel *release.ChannelRelease) error {
+func checkAirgapMatches(c *cli.Context) error {
+	rel, err := release.GetChannelRelease()
+	if err != nil {
+		return fmt.Errorf("failed to get release from binary: %w", err) // this should only be if the release is malformed
+	}
 	if rel == nil {
 		return fmt.Errorf("airgap bundle provided but no release was found in binary, please rerun without the airgap-bundle flag")
 	}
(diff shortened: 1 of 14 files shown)


pkg/cmd/install.go
```diff
@@ -799,10 +800,28 @@ func installCommand() *cli.Command {
 					return err // we want the user to see the error message without a prefix
 				}
 			}
+
+			if !isAirgap && license != nil {
```
sgalsaleh, line 804: nit: it would cleaner to have this entire block be part of the `maybePromptForAppUpdate` function

sgalsaleh (approved):

sgalsaleh (approved):

pkg/cmd/install.go
```diff
@@ -799,10 +800,22 @@ func installCommand() *cli.Command {
...
 			}
+
+			if err := maybePromptForAppUpdate(c, prompts.New(), license); err != nil {
+				if errors.Is(err, ErrNothingElseToAdd) {
+					metrics.ReportApplyFinished(c, err)
+					return err
+				}
+				// If we get an error other than ErrNothingElseToAdd, we warn and continue as
+				// this check is not critical.
+				logrus.Warnf("Failed to check for newer app versions: %v", err)
```
ajp-io, line 811: Does this get printed out? IMO we should just log it internally if anything and move on. There's nothing a user can do about this, so it's more likely to confuse them than anything.

emosbaugh: updated

pkg/cmd/install.go
```diff
@@ -799,10 +800,22 @@ func installCommand() *cli.Command {
...
 			}
+
+			if err := maybePromptForAppUpdate(c, prompts.New(), license); err != nil {
+				if errors.Is(err, ErrNothingElseToAdd) {
+					metrics.ReportApplyFinished(c, err)
+					return err
+				}
+				// If we get an error other than ErrNothingElseToAdd, we warn and continue as
+				// this check is not critical.
+				logrus.Infof("WARNING: Failed to check for newer app versions: %v", err)
```
ajp-io, line 811: My comment might not have been clear. I think `logrus.Infof` still prints to the terminal. I'm saying that if we fail to perform this check, that should be silent to the user, because they can't really do anything about that, and it won't affect their installation. I think printing a warning will be construed as "something went wrong" and cause people to reach out to us.

emosbaugh: Sorry it was clear. I didnt realize info printed to the terminal. I will change to debug

emosbaugh: updated

pkg/cmd/install.go
emosbaugh: updated

pkg/cmd/vendorapp.go
```diff
@@ -0,0 +1,136 @@
...
+		releaseURL,
+		license.Spec.LicenseID,
+		channelRelease.AppSlug,
+		channelRelease.ChannelSlug,
+	)
+
+	// if no-prompt is true, we don't prompt the user and continue by default.
+	// SKIP_APP_UPDATE_PROMPT is an escape hatch used by the CI to skip the prompt in case this
+	// release becomes out of date.
+	if !c.Bool("no-prompt") && os.Getenv("SKIP_APP_UPDATE_PROMPT") != "true" {
```
sgalsaleh, line 72: doesn't the CI already use --no-prompt though? isn't `if !c.Bool("no-prompt") {` sufficient?

pkg/prompts/decorative/decorative.go
```diff
@@ -2,18 +2,53 @@
...
+
 	"github.com/AlecAivazis/survey/v2"
+	"github.com/AlecAivazis/survey/v2/terminal"
 	"github.com/sirupsen/logrus"
 )
 
 // Decorative is a decorative prompt.
-type Decorative struct{}
+type Decorative struct {
+	in  terminal.FileReader
```
sgalsaleh, line 14: why not just use the plain prompt?

emosbaugh: updated

sgalsaleh (approved):

PR #1507, opened 2024-11-15
chore(preflights): replace run collector with host sysctl for ip_forwarding

JGAntunes:

Follow up to https://github.com/replicatedhq/embedded-cluster/pull/1500 we should use the already existing `sysctl` collector for the `ip_forwarding` check

NONE

I run a manual test locally, however it's hard to reproduce with the recent changes introduced in - https://github.com/replicatedhq/embedded-cluster/pull/1484. Instead I cherry picked the analyzer and collector and run them separately using the preflight binary:

image

After `sysctl net.ipv4.ip_forward=0`

image

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/preflights/host-preflight.yaml
+++ b/pkg/preflights/host-preflight.yaml
@@ -148,10 +148,6 @@ spec:
         exclude: '{{ eq .GlobalCIDR.CIDR "" }}'
         CIDRRangeAlloc: '{{ .GlobalCIDR.CIDR }}'
         desiredCIDR: {{.GlobalCIDR.Size}}
-    - run:
-        collectorName: "kernel-parameters"
-        command: "sysctl"
-        args: ["-a"]
     - sysctl: {}
   analyzers:
     - cpu:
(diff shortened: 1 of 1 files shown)


pkg/preflights/host-preflight.yaml
```diff
@@ -886,3 +871,12 @@ spec:
...
               when: 'net.ipv4.conf.all.arp_ignore == 0'
               message: "ARP ignore is disabled for all interfaces on the host."
+    - sysctl:
+        checkName: "IP forwarding"
+        outcomes:
+          - fail:
+              when: 'net.ipv4.ip_forward == 0'
+              message: "IP forwarding must be enabled. To enable it, edit /etc/sysctl.conf, add or uncomment the line 'net.ipv4.ip_forward=1', and run 'sudo sysctl -p'."
+          - pass:
+              when: 'net.ipv4.ip_forward > 0'
```
JGAntunes, line 881: According to the docs - https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html:
> ip_forward - BOOLEAN
>
>        0 - disabled (default)
>
>        not 0 - enabled

ajp-io (approved):

PR #1519, opened 2024-11-18
fix: new app version prompt default to no, do not prompt if no terminal

emosbaugh:

Changes the default for the prompt to no. No longer respects the --no-prompt flag and instead will not prompt if there is no terminal.

```release-note

```


laverya (approved):
Makes sense to me, I haven't run it yet though

PR #1528, opened 2024-11-20
fix(metrics): redact specific flags which might hold secrets

JGAntunes:

Redact secrets sent through the metrics package through the flags parameter for installations. We do this by filtering the provided flags and, upon detecting predefined flags, redact the provided values and replace them by `"*****"`.

https://app.shortcut.com/replicated/story/116082/redact-secrets-from-embedded-cluster-bot-alerts-slack-channel

Automated tests added.

```release-note
Redact potential secret information sent to our metrics backend through the provided flag parameters
```

NONE

### DIFF
--- a/pkg/metrics/util.go
+++ b/pkg/metrics/util.go
@@ -46,7 +46,7 @@ func redactFlags(flags []string) []string {
 		if isFlagKey && strings.Contains(flag, "=") {
 
 			// Split the flag into key and value
-			flagParts := strings.Split(flag, "=")
+			flagParts := strings.SplitN(flag, "=", 2)
 			key := flagParts[0]
 			value := flagParts[1]
 
--- a/pkg/metrics/util.go
+++ b/pkg/metrics/util.go
@@ -3,6 +3,7 @@ package metrics
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/replicatedhq/embedded-cluster/pkg/metrics/types"
 	"github.com/replicatedhq/embedded-cluster/pkg/versions"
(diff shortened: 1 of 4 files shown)


pkg/metrics/util.go
```diff
@@ -22,3 +23,33 @@ func EventPayload(ev types.Event) ([]byte, error) {
 	payload := map[string]interface{}{"event": ev, "versions": vmap}
 	return json.Marshal(payload)
 }
+
+var secretKeywords = []string{"password", "secret", "token", "key"}
+
+// redactFlags redacts presumed secret values from a string slice based on preset keywords.
+// For example, given ["--license", "./path", "--password", "123"], it will return ["--license", "./path", "--password", "*****"]
+func redactFlags(flags []string) []string {
```
emosbaugh, line 31: i dont think you handle the case where the flag is used with an equals sign

```
--admin-console-password=*****
```

emosbaugh (changes requested):

pkg/metrics/util.go
```diff
@@ -22,3 +23,33 @@ func EventPayload(ev types.Event) ([]byte, error) {
...
+	result := make([]string, len(flags))
+
+	valueHasSecret := false
+	for i, flag := range flags {
+		// Check if the is the flag name or the value
+		isFlagKey := strings.HasPrefix(flag, "-")
```
jtuchscherer, line 37: If the string we want to redact (like the password or the key) starts with a dash, we would miss redacting it. This is probably a very unlikely edge case, but it could still occur.

pkg/metrics/util.go
JGAntunes: Good point, let me rework this.

pkg/metrics/util.go
JGAntunes: Good call, let me check for that.

pkg/metrics/util.go
JGAntunes: Fixed via - https://github.com/replicatedhq/embedded-cluster/pull/1528/commits/9a65e095c4760737bd0e47ca847cc2440b0a7a4d

pkg/metrics/util.go
```diff
@@ -22,3 +23,57 @@ func EventPayload(ev types.Event) ([]byte, error) {
...
+		// The previous iteration detected one of the secret keywords on flag key, let's redact this value
+		if valueHasSecret {
+			result[i] = "*****"
+			valueHasSecret = false
+			continue
+		}
+		// Check if this is the flag name or the value
+		isFlagKey := strings.HasPrefix(flag, "-")
+		// Flag is of the form --key=value
+		if isFlagKey && strings.Contains(flag, "=") {
```
JGAntunes, line 46: I was hopping to avoid this "complex" parsing logic by reusing something but it so happens that most of what I could find is built around parsing flags for a set of predefined keys, not an arbitrary set like we're trying to do here. (i.e. the standard flag package - https://pkg.go.dev/flag)

JGAntunes:
@emosbaugh after our chat yesterday I opted to simplify the approach where instead of using a preset of keywords we use specific flags we want to redact.

The main issue with trying to pursue a more hardened approach here that could work for future changes (e.g. we add a new `--token` flag to the install command) is the type associated with each flag and the potential for edge cases. For example we can have a boolean flag and we have no idea if there should a value afterwards or not. And we could also technically have a password start with `-` or `--`, which would lead us to believe that could in fact be a flag.

Something we can consider pursuing in the future if we end up having to mutate this variable a lot:
- https://github.com/replicatedhq/embedded-cluster/blob/b408461439047526e6d2071460103867271e009b/pkg/metrics/util.go#L27
Would be finding a way to export the flag set we define here, so that we can reuse across these 2 places and find a way to decorate each flag definition so that we could say "this should always be redacted":
- https://github.com/replicatedhq/embedded-cluster/blob/b408461439047526e6d2071460103867271e009b/cmd/installer/cli/install.go#L274-L298
The lift to do it VS what we would get out of it right now though feels too much for me.

emosbaugh (approved):

emosbaugh (approved):

PR #1531, opened 2024-11-21
Move provider init to the root cmd

marccampbell:


cmd/installer/cli/install.go
```diff
@@ -120,9 +119,7 @@ func InstallCmd(ctx context.Context, name string) *cobra.Command {
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			provider := defaults.NewProviderFromRuntimeConfig(runtimeConfig)
 			os.Setenv("TMPDIR", provider.EmbeddedClusterTmpSubDir())
```
sgalsaleh, line 122: why isn't this also moved?

PR #1537, opened 2024-11-22
Add ability to attach a replicated release to an unreleased EC binary

marccampbell:


Makefile
```diff
@@ -359,3 +359,49 @@ test-lam-e2e: pkg/goods/bins/local-artifact-mirror
 bin/installer:
 	@mkdir -p bin
 	go build -o bin/installer ./cmd/installer
+
+# make test-embed channel=Unstable app=slackernews
+.PHONY: test-embed
+test-emded: export OS=linux
+test-embed: export ARCH=amd64
```
sgalsaleh, line 366: This means we can't use this in the dev env no?

Makefile
```diff
@@ -359,3 +359,49 @@ test-lam-e2e: pkg/goods/bins/local-artifact-mirror
...
+test-embed: static embedded-cluster
+	@echo "Cleaning up previous release directory..."
```
sgalsaleh, line 368: Feels like we should move this to dagger or a script. Seems overloaded to have all of this logic in the makefile

Makefile
```diff
@@ -359,3 +359,49 @@ test-lam-e2e: pkg/goods/bins/local-artifact-mirror
...
+	rm -rf ./hack/release
```
sgalsaleh, line 369: We use the "build" directory in other places for gitignored artifacts

Makefile
```diff
@@ -359,3 +359,49 @@ test-lam-e2e: pkg/goods/bins/local-artifact-mirror
 bin/installer:
 	@mkdir -p bin
 	go build -o bin/installer ./cmd/installer
+
+# make test-embed channel=Unstable app=slackernews
+.PHONY: test-embed
```
sgalsaleh, line 364: Why not use the embedded-cluster-release-builder binary we already have? We could replace it with this one I don't mind.

Makefile
marccampbell: it won't affect, this exports them only in this 1 new make target

Makefile
marccampbell: ok, but this is not a build artifact, the build artifact is in the build dir   this is added to gitignore

Makefile
marccampbell: this is a little more specific to the workflow i am trying to crfeate

sgalsaleh (approved):

Makefile
sgalsaleh: I understand. I'm saying we won't be able to benefit from it. Which might be fine

PR #1551, opened 2024-11-26
Add comment about nil runtime config in old installation objects

sgalsaleh:

Adds a comment about nil runtime config in old installation objects

NONE

```release-note
NONE
```

NONE


sgalsaleh:
Merging as this is only a comment and does not affect tests.

PR #1553, opened 2024-11-26
send preflight results event when preflights fail

laverya:

When a preflight fails/warns, send a copy of the preflight results to our metrics endpoint.

```release-note

```


cmd/installer/cli/install_runpreflights.go
```diff
@@ -378,10 +379,10 @@ func RunHostPreflights(cmd *cobra.Command, applier *addons.Applier, replicatedAP
 		return nil
 	}
 
-	return runHostPreflights(cmd, hpf, proxy, assumeYes)
+	return runHostPreflights(cmd, hpf, proxy, assumeYes, replicatedAPIURL)
 }
 
-func runHostPreflights(cmd *cobra.Command, hpf *v1beta2.HostPreflightSpec, proxy *ecv1beta1.ProxySpec, assumeYes bool) error {
+func runHostPreflights(cmd *cobra.Command, hpf *v1beta2.HostPreflightSpec, proxy *ecv1beta1.ProxySpec, assumeYes bool, replicatedAPIURL string) error {
```
emosbaugh, line 385: since this function is called from many different command wouldnt it be helpful to report which command was run? also, is it intention to report for the `run-preflights` commands? and not related to this change, but why prompt at all for those commands?

laverya: I have updated the reporting to include the command that was called (install, run-preflights, join etc)

laverya: > and not related to this change, but why prompt at all for those commands?

Because it's the same codepath, but we hide the 'skip-host-preflights' and 'ignore-host-preflight' flags on those commands, so there shouldn't be prompts (the code we're using expects the flags to exist, which is why they're there at all)

emosbaugh (approved):

PR #1571, opened 2024-12-02
installer2

marccampbell:
adds the new `install2` hidden command which does not actually install anything yet.

some related refactoring:

IsInstalled should not be responsible for printing the error output, it's a simple function. The caller should determine if and when to print a function. 

my local env test-embed: set the version since the git history is not the right place for this.

moved some of the proxy parsing out and into a parseProxyArgs function. it's not yet called everywhere, but will be .  

added an env var to not report metrics, using the closest name to an industry standard env var name i could find.

### DIFF
--- a/cmd/installer/cli/cidr.go
+++ b/cmd/installer/cli/cidr.go
@@ -18,21 +18,37 @@ func addCIDRFlags(cmd *cobra.Command) {
 	cmd.Flags().String("cidr", ecv1beta1.DefaultNetworkCIDR, "CIDR block of available private IP addresses (/16 or larger)")
 }
 
-// DeterminePodAndServiceCIDRS determines, based on the command line flags,
+func parseCIDRFlags(cmd *cobra.Command) error {
+	if cmd.Flags().Changed("cidr") && (cmd.Flags().Changed("pod-cidr") || cmd.Flags().Changed("service-cidr")) {
+		return fmt.Errorf("--cidr flag can't be used with --pod-cidr or --service-cidr")
+	}
+
+	cidr, err := cmd.Flags().GetString("cidr")
+	if err != nil {
+		return fmt.Errorf("unable to get cidr flag: %w", err)
+	}
+
+	if err := netutils.ValidateCIDR(cidr, 16, true); err != nil {
+		return fmt.Errorf("invalid cidr %q: %w", cidr, err)
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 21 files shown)


cmd/installer/cli/cidr.go
```diff
@@ -18,21 +18,37 @@ func addCIDRFlags(cmd *cobra.Command) {
...
 
-// DeterminePodAndServiceCIDRS determines, based on the command line flags,
-// what are the pod and service CIDRs to be used for the cluster. If both
-// --pod-cidr and --service-cidr have been set, they are used. Otherwise,
-// the cidr flag is split into pod and service CIDRs.
-func determinePodAndServiceCIDRs(cmd *cobra.Command) (string, string, error) {
-	podCIDRFlag, err := cmd.Flags().GetString("pod-cidr")
-	if err != nil {
-		return "", "", fmt.Errorf("unable to get pod-cidr flag: %w", err)
+func parseCIDRFlags(cmd *cobra.Command) error {
```
sgalsaleh, line 21: `validateCIDRFlags` might be a better name?

pkg/addons2/openebs/install.go
```diff
@@ -0,0 +1,5 @@
+package openebs
+
+func (o *OpenEBS) Install() error {
```
sgalsaleh, line 3: we might want to always just use `helm upgrade -i` if we ever want the installation process to be idempotent.

pkg/websocket/server.go
```diff
@@ -17,6 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+// wsDialer is the default dialer but with a shorter timeout.
```
sgalsaleh, line 20: i removed this comment in this pr because it's not accurate :). as this dialer doesn't use a proxy either.

pkg/metrics/reporter.go
```diff
@@ -21,6 +21,10 @@ import (
 var clusterIDMut sync.Mutex
 var clusterID *uuid.UUID
 
+func metricsEnabled() bool {
```
sgalsaleh, line 24: why not (re-)use the `DisableMetrics` function in `pkg/metrics/airgap.go` that we already use for airgap? can have it respect this env var for example.

pkg/addons2/openebs/install.go
marccampbell: not implemented here, but sure

sgalsaleh (approved):

PR #1574, opened 2024-12-03
Update image-scan job to exclude kube-proxy

jtuchscherer:
Currently, every image scan workflow fails because of an image with an accepted vulnerability. This PR will exclude the image from the scan

no

no

no


marccampbell (approved):

emosbaugh:
Why not use a trivyignore file? https://trivy.dev/latest/docs/configuration/filtering/#by-finding-ids

PR #1575, opened 2024-12-04
Make openebs addon install successfully

sgalsaleh:


pkg/addons2/openebs/static/metadata.yaml
sgalsaleh, line 1: i copied this file from the `addons` package instead of referencing the existing one. it does mean that this won't get updated automatically, but we can later merge these files when we replace the `addons` package.

pkg/addons2/openebs/static/values.tpl.yaml
sgalsaleh, line 1: ditto

PR #1576, opened 2024-12-04
Install Admin Console as part of install2

sgalsaleh:


pkg/addons2/adminconsole/install.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return errors.Wrap(err, "create helm client")
+	}
+
+	_, err = helm.Install(ctx, releaseName, Metadata.Location, Metadata.Version, helmValues, namespace)
+	if err != nil {
+		return errors.Wrap(err, "install admin console")
+	}
+
+	// install the application
```
marccampbell, line 58: do we need to do anything here to make sure it's ready / running before the next step?

pkg/addons2/adminconsole/install.go
```diff
@@ -0,0 +1,194 @@
...
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func (a *AdminConsole) Install(ctx context.Context, kcli client.Client, writer *spinner.MessageWriter) error {
+	// some resources are not part of the helm chart and need to be created before the chart is installed
+
+	if err := createNamespace(ctx, kcli, namespace); err != nil {
```
marccampbell, line 25: helm has a --create-namespace flag, can we just set that var in our helm installer to make this unecessary?

pkg/addons2/adminconsole/install.go
```diff
@@ -0,0 +1,194 @@
...
+		return errors.Wrap(err, "create namespace")
+	}
+
+	if err := createKotsPasswordSecret(ctx, kcli, namespace, a.Password); err != nil {
+		return errors.Wrap(err, "create kots password secret")
+	}
+
+	if err := createKotsCAConfigmap(ctx, kcli, namespace, a.PrivateCAs); err != nil {
```
marccampbell, line 33: why aren't these helm values?

pkg/addons2/adminconsole/install.go
sgalsaleh: we need to create some resources before we install the chart, and we need the namespace to exist in order to do that. i've captured part of that in the comment above

pkg/addons2/adminconsole/install.go
sgalsaleh: the `--wait` flag is set to true in the helm install, which waits for things to be ready before moving on

pkg/addons2/adminconsole/install.go
marccampbell: ok, lets put these (and the namespace) into a function / block that we identify as needing to be moved to kots, and then schedule that work on the helm chart so that we can remove it from here?  no change to this code, just organize it for easy removal as a single block?

marccampbell (approved):

PR #1579, opened 2024-12-05
Auto select network iface when not provided

marccampbell:


cmd/installer/cli/install2.go
```diff
@@ -88,6 +91,17 @@ func Install2Cmd(ctx context.Context, name string) *cobra.Command {
 			flags.podCIDR = pod
 			flags.serviceCIDR = svc
 
+			// if a network interface flag was not provided, attempt to discover it
+			if flags.networkInterface == "" {
+				autoInterface, err := determineBestNetworkInterface()
+				if err != nil {
+					flags.autoSelectNetworkInterfaceErr = err
```
sgalsaleh, line 98: i feel like this should return an error

marccampbell: i don't think this should return an error.  there are times when we can't auto determine using this function, and i want to fall back to the current behavior for now.  the goal is to add a prompt when we can't auto determine, but working with @ajp-io on the UX for that prompt, will be adding in a "soon" PR.

if we return an error here, we will break the current product experience

sgalsaleh (approved):

PR #1580, opened 2024-12-05
run host preflights as part of the install2 2

sgalsaleh:


sgalsaleh:
Merging into a branch

PR #1582, opened 2024-12-05
Configure sysctl before running preflights in install2

sgalsaleh:


marccampbell (approved):

sgalsaleh:
Merging as this does not affect tests yet.

PR #1585, opened 2024-12-07
Install Velero as part of install2 command

sgalsaleh:


pkg/addons2/addons.go
```diff
@@ -44,6 +46,16 @@ func InstallAddons(ctx context.Context, opts InstallOptions) error {
 		},
 	}
 
+	drEnabled, err := helpers.DisasterRecoveryEnabled(opts.License)
```
marccampbell, line 49: i think it would be cleaner to have a way for opts.IsDisasterRecoveryEnabled as a variable instead of this lookup here. Let's aim to make it more self contained.

marccampbell (changes requested):

pkg/addons2/addons.go
sgalsaleh: hmm, this is just using `license.Spec.IsDisasterRecoverySupported`, and the license is an input to this function. i can just get rid of this helper function

sgalsaleh: updated. let me know if that works

pkg/addons2/addons.go
```diff
@@ -44,6 +45,12 @@ func InstallAddons(ctx context.Context, opts InstallOptions) error {
 		},
 	}
 
+	if opts.License != nil && opts.License.Spec.IsDisasterRecoverySupported {
```
marccampbell, line 48: I still want to make this simpler. This function should not depend on that specific field and implementation of the license spec struct.

Have the caller (install2.go) make the determination if velero needs to be installed, and pass it in as a bool (IsVeleroInstalled, something like that). Right now, this jkust means that you'll pass that from opts.License.Spec into that bool field, but we remove the knowledge of license in this function.

The end result is that we have something that's more integration testable since we don't need to provide a license, we can provide the inputs to the function, the direction of what to do.

marccampbell (changes requested):

pkg/addons2/velero/velero.go
```diff
@@ -0,0 +1,47 @@
+package velero
+
+import (
+	_ "embed"
+
+	"github.com/pkg/errors"
+	ecv1beta1 "github.com/replicatedhq/embedded-cluster/kinds/apis/v1beta1"
+	"github.com/replicatedhq/embedded-cluster/pkg/release"
+	"github.com/replicatedhq/embedded-cluster/pkg/runtimeconfig"
+	"gopkg.in/yaml.v2"
```
emosbaugh, line 10: why not v3?

marccampbell (approved):

marccampbell (approved):

PR #1596, opened 2024-12-09
Free up more disk space on the github runner

sgalsaleh:


marccampbell (approved):

laverya (approved):

laverya:
It looks like this only takes about 30 seconds to run along with the other disk space cleanup processes, and saves an extra 10GB - easily worth it

PR #1601, opened 2024-12-10
feat: Collect license used to install cluster in support bundles

banjoh:
- Store license file used to install cluster in `<data-dir>`
- Add host collector spec to collect the license in support bundle

Demo: https://asciinema.org/a/uhpIRKKc5C9k7MtP1ovI69Kte

I missed showing the file permissions of `license.yaml` in the demo
```sh
[evans] $ make ssh-node0
root@node0:/replicatedhq/embedded-cluster# ls -al /root/ec-data-dir/license.yaml
-r-------- 1 root root 5514 Dec 23 16:24 /root/ec-data-dir/license.yaml
root@node0:/replicatedhq/embedded-cluster#
```

[sc-116758](https://app.shortcut.com/replicated/story/116758/include-license-in-host-support-bundles-generated-by-embedded-cluster)

NONE

```release-note
Store installer license file in data-dir and add collect when generating a support bundle
```

NONE

### DIFF
--- a/cmd/installer/cli/install.go
+++ b/cmd/installer/cli/install.go
@@ -196,6 +196,12 @@ func InstallCmd(ctx context.Context, name string) *cobra.Command {
 				return err
 			}
 
+			logrus.Debugf("copy license file to %s", dataDir)
+			if err := copyLicenseFileToDataDir(licenseFile, dataDir); err != nil {
+				// TODO: Do we need to report this error?
+				logrus.Warnf("unable to copy license file to %s: %v", dataDir, err)
+			}
+
 			opts := addonsApplierOpts{
 				assumeYes:    assumeYes,
 				license:      licenseFile,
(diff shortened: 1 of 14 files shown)


JGAntunes (commented):
Change LGTM overall, wondering what you think about adding a test for this changes?

My suggestion would be to piggy back on the dryrun tests we have for the `install` command (which should be relatively quick to do) here:
- https://github.com/replicatedhq/embedded-cluster/blob/2c21588e99729804c8050bb8f20516309ed539ae/tests/dryrun/install_test.go#L19-L40

JGAntunes (dismissed):

laverya (approved):

laverya:
well, that solution.... did not work
(two reasons: one is that the new location for the check also encountered the 'previous-stable' tests, which don't have this code yet, and another is that it also ran after _restores_ which don't have the file either)

jtuchscherer (approved):
Went through the change with Andrew in person. Passing the short sha variable is not great, but it's a good solution for now

PR #1633, opened 2024-12-16
Handle webscoket upgrade command

marccampbell:


pkg/upgrade/upgrade.go
```diff
@@ -0,0 +1,230 @@
...
+
+	fmt.Printf("Upgrading to version %s\n", in.Spec.Config.Version)
+	err = k0sUpgrade(ctx, cli, in)
+	if err != nil {
+		return fmt.Errorf("k0s upgrade: %w", err)
+	}
+
+	// Augment the installation with data dirs that may not be present in the previous version.
+	// This is important to do ahead of updating the cluster config.
+	// We still cannot update the installation object as the CRDs are not updated yet.
```
sgalsaleh, line 52: this is no longer true as we're using configmaps not CRDs. we should probably just remove `reApplyInstallation` and update the installation here directly as part of this or right after it

pkg/upgrade/installation.go
```diff
@@ -0,0 +1,78 @@
...
+	clusterv1beta1 "github.com/replicatedhq/embedded-cluster/kinds/apis/v1beta1"
+	"github.com/replicatedhq/embedded-cluster/pkg/kubeutils"
+	"k8s.io/client-go/util/retry"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// reApplyInstallation updates the installation spec to match what's in the configmap used by the upgrade job.
+// This is required because the installation CRD may have been updated as part of this upgrade, and additional fields may be present now.
+func reApplyInstallation(ctx context.Context, cli client.Client, in *clusterv1beta1.Installation) error {
```
sgalsaleh, line 16: not used :)

sgalsaleh (approved):

sgalsaleh (approved):

PR #1706, opened 2025-01-15
fix(dr): improved dr should not restore the component app backup

emosbaugh:

Improved dr should not restore the component app backup as well as the custom restore cr.

```release-note

```


laverya (approved):

laverya:
I'm trusting the e2e tests to have caught any possible regressions here

PR #1718, opened 2025-01-17
feat(ec): migration from v1 to v2 - fixes

emosbaugh:

Continued progress for v1 to v2 migrations.

This changes:
- using a job to a pod for the manager installation
- kots will now run a pod for the migration rather than run it in process as the eco namespace gets removed and there are permissions errors
- fixes a bug where the operator would reconcile the configmap rather than CR
- disable reconciliation in operator rather than removing operator to prevent failed migrations from leaving cluster in a bad state

```release-note

```


operator/pkg/cli/migrate_v2_pod.go
```diff
@@ -0,0 +1,339 @@
+package cli
```
laverya, line 1: why did this file end up moving out of migratev2?

operator/pkg/cli/migrate_v2_pod.go
```diff
@@ -0,0 +1,339 @@
...
+	logf("Successfully ensured v2 migration pod")
+
+	logf("Waiting for v2 migration pod to finish")
+	err = waitForMigrateV2Pod(ctx, cli)
+	if err != nil {
+		return fmt.Errorf("wait for pod: %w", err)
+	}
+	logf("Successfully waited for v2 migration pod to finish")
+
+	// NOTE: the installation config map cannot be deleted because the service account gets deleted
```
sgalsaleh, line 113: why would you even want to delete the configmap?

sgalsaleh (approved):

laverya (approved):

operator/pkg/cli/migrate_v2_pod.go
emosbaugh: just to clean things up

operator/pkg/cli/migrate_v2_pod.go
emosbaugh: it didnt. i changed it from "job" to "pod"

laverya: That just says "it was unintentional" to me, though?

laverya: It used to be in operator/pkg/cli/migratev2/managerjob.go

emosbaugh: This is a new pod that got added. managerjob.go changed to managerpod.go.

This file is covered in the bullet `kots will now run a pod for the migration rather than run it in process as the eco namespace gets removed and there are permissions errors`

PR #1727, opened 2025-01-21
feat: v2 migration e2e tests

emosbaugh:

```release-note

```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -473,6 +473,77 @@ jobs:
           echo "K0S_VERSION=\"$K0S_VERSION\""
           echo "k0s_version=$K0S_VERSION" >> "$GITHUB_OUTPUT"
 
+  build-upgrade-v2:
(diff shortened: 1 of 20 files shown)


.github/workflows/ci.yaml
```diff
@@ -473,6 +473,77 @@ jobs:
           echo "K0S_VERSION=\"$K0S_VERSION\""
           echo "k0s_version=$K0S_VERSION" >> "$GITHUB_OUTPUT"
 
+  build-upgrade-v2:
+    name: Build upgrade v2
+    runs-on: embedded-cluster-2
```
laverya, line 478: why does this need to be an additional embedded cluster release? Is it being built differently than the normal upgrade version?

emosbaugh: this tests a v2 to v2 upgrade after a v1 to v2 migration with the build-upgrade version. if it did not create a different ec release then it wouldnt be able to test a cluster upgrade.

PR #1729, opened 2025-01-23
fix: [sc-118372] /var/lib/embedded-cluster is not exec by other non-root user

nvanthao:
Story details: https://app.shortcut.com/replicated/story/118372

## Problem
When creating directories using `os.MkdirAll`, the final permissions are affected by the system's `umask`. This behavior is undesirable because it can lead to permission issues when other processes attempt to access or modify files in our data directory (`/var/lib/embedded-cluster`). Specifically, directories may end up with restrictive permissions that prevent necessary access by other services or processes.

## Solution
To ensure consistent and correct directory permissions, this PR sets the `umask` to `022` before beginning execution of the install, restore, or join commands.

## Demo
Before: https://asciinema.org/a/6BIISglNrpFdvONP03d683q15

```
ls -lah /var/lib/embedded-cluster/                                                                                  
total 24K                                                                                                                                            
drwx------  6 root root 4.0K Jan 22 23:54 .                                                                                                          
drwxr-xr-x 39 root root 4.0K Jan 22 23:54 ..                                                                                                         
drwx------  2 root root 4.0K Jan 22 23:54 bin                                                                                                        
drwx------  3 root root 4.0K Jan 22 23:54 k0s                                                                                                        
drwx------  2 root root 4.0K Jan 22 23:54 support                                                                                                    
drwx------  2 root root 4.0K Jan 22 23:54 tmp
```

After: https://asciinema.org/a/GZCkeoOtPuW0uOPhCDP6EuXdc

```
ls -lah /var/lib/embedded-cluster/
total 36K
drwxr-xr-x 6 root root 4.0K Jan 23 02:46 .
drwxr-xr-x 1 root root 4.0K Jan 23 02:20 ..
drwxr-xr-x 2 root root 4.0K Jan 23 02:46 bin
drwxr-xr-x 3 root root 4.0K Jan 23 02:46 k0s
drwx------ 2 root root 4.0K Jan 23 02:46 support
drwxr-xr-x 2 root root 4.0K Jan 23 02:46 tmp
```


pkg/helpers/fs.go
```diff
@@ -96,3 +96,13 @@ func RemoveAll(path string) error {
 	}
 	return me.ErrorOrNil()
 }
+
+// InitDir creates a directory if it does not exist and make sure it has the right permissions
+func InitDir(dir string, perm os.FileMode) error {
```
sgalsaleh, line 101: nit: why not call it `MkdirAll`?

sgalsaleh (commented):
There's many other places in the code that use `os.MkdirAll` that I think we should change as well

pkg/helpers/fs.go
```diff
@@ -96,3 +96,13 @@ func RemoveAll(path string) error {
...
+	if err := os.MkdirAll(dir, perm); err != nil {
+		return fmt.Errorf("unable to create directory %s: %w", dir, err)
```
emosbaugh, line 103: i see no reason to wrap this. it is redundant

```suggestion
		return err
```

nvanthao:
I've added more dir but only for dirs under `EmbeddedClusterHomeDirectory` as it could be use by other users created by k0s.

banjoh:
I think we should modify these as well since they are code paths in the install workflow
- https://github.com/replicatedhq/embedded-cluster/blob/381a48ee27991225b815a51e67229ed5d785dc4e/pkg/runtimeconfig/defaults.go#L36
- https://github.com/replicatedhq/embedded-cluster/blob/381a48ee27991225b815a51e67229ed5d785dc4e/pkg/airgap/materialize.go#L63

@emosbaugh / @sgalsaleh is there any harm in applying this change to all `os.MkdirAll` in the codebase?

sgalsaleh:
@banjoh no, that should be fine I think.

banjoh (changes requested):
I ran an install using a binary built from this PR but the install failed. Here is my install attempt

```sh
root@evans-vm1:/home/evans# umask 0077
root@evans-vm1:/home/evans# ./embedded-cluster install
? Set the Admin Console password (minimum 6 characters): ******
? Confirm the Admin Console password: ******
✔  Host files materialized!
✔  Host preflights succeeded!
✔  Waiting for embedded-cluster node to be ready
unable to wait for node: unable to get status: exit status 1: Error: status: can't get "status" via "/run/k0s/status.sock": Get "http://localhost/status": dial unix /run/k0s/status.sock: connect: connection refused
```

I picked a number of paths where the permissions were not right such as
- `drwx------ root /var/lib/embedded-cluster/`
- All files in `/var/lib/embedded-cluster/k0s/bin` have `-r-xr-x--- root`. They do not have global execute permission

Since the install failed very early on, there might have been some other paths not set right later on.

### Changes needed
- I addition to the changes here, I think we should consider setting `0022` umask before the install starts using `unix.Umask(0022)`. This will be a good safeguard to ensure we do not miss created files as well. This PR just addresses directories.
- We need to test `join`ing a new node.
- We need cover install and joining in e2e tests where the umask used is `0077`. I choose this one since its what recommendations such as STIG/CIS document.

sgalsaleh:
@banjoh did you resolve conflicts first before you built the binary?

banjoh:
> @banjoh did you resolve conflicts first before you built the binary?

Hmm. No. Good catch. I'll do that then run another test.

banjoh:
```go
            umask := 0022
            oldUmask := unix.Umask(umask)
            if oldUmask != umask {
                logrus.Debugf("seting umask to %04o from %04o", umask, oldUmask)
            }
```

@nvanthao I added this change to `func RootCmd(ctx context.Context, name string) *cobra.Command` to test working how the install works. It succeeded with your last commit (before the merge done above - `git checkout 1d31f87d1a35d7415fcfe913315fc91e8b23d7fa`)

laverya:
```
    install_test.go:1071: set umask stdout: 'umask 0077' >> /etc/profile
    install_test.go:1072: stderr: 
    install_test.go:1078: check umask stdout: 0022
    install_test.go:1079: stderr: 
```
evidently my test did not work

banjoh (approved):

cmd/installer/cli/root.go
```diff
@@ -68,6 +69,9 @@ func RootCmd(ctx context.Context, name string) *cobra.Command {
 				metrics.DisableMetrics()
 			}
 
+			// set the umask to 022 so that we can create files/directories with 755 permissions
+			_ = syscall.Umask(0o022)
```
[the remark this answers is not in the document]
laverya: I'll add it to the various 'main' functions, and try to come up with a better comment for it

cmd/installer/cli/install.go
```diff
@@ -171,6 +172,9 @@ func preRunInstall(cmd *cobra.Command, flags *InstallCmdFlags) error {
 		return fmt.Errorf("install command must be run as root")
 	}
 
+	// set the umask to 022 so that we can create files/directories with 755 permissions
+	_ = syscall.Umask(0o022)
```
emosbaugh, line 176: can you at the very least logrus.Debug this error?

cmd/installer/cli/join.go
```diff
@@ -96,6 +97,9 @@ func preRunJoin(flags *JoinCmdFlags) error {
 
 	flags.isAirgap = flags.airgapBundle != ""
 
+	// set the umask to 022 so that we can create files/directories with 755 permissions
+	_ = syscall.Umask(0o022)
```
emosbaugh, line 101: logrus.Debug the error

cmd/installer/cli/root.go
emosbaugh, line 73: logrus.Debug the error

emosbaugh (changes requested):

cmd/installer/cli/install.go
laverya: it doesn't return an error, it returns the previous umask value

tests/dryrun/install_test.go
```diff
@@ -391,3 +396,35 @@ func TestConfigValuesInstallation(t *testing.T) {
...
+func TestRestrictiveUmask(t *testing.T) {
+	oldUmask := syscall.Umask(0o077)
+	defer syscall.Umask(oldUmask)
+
+	testDefaultInstallationImpl(t)
+
+	// check that folders created in this test have the right permissions
+	folderList := []string{
+		runtimeconfig.EmbeddedClusterHomeDirectory(),
+		runtimeconfig.EmbeddedClusterBinsSubDir(),
```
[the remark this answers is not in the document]
laverya: kubectl-preflight has been added

cmd/installer/cli/root.go
banjoh: The return value is not an error. Its the [previous umask](https://pkg.go.dev/syscall#Umask).

emosbaugh (approved):

PR #1733, opened 2025-01-23
helm chart install should work when airgapped

laverya:

This adds a number of install2 tests, and fixes a few problems discovered while doing that - primarily airgap helm chart installation support and marking the installation as 'installed' after installing helm charts.

```release-note

```


pkg/addons2/seaweedfs/install.go
```diff
@@ -27,9 +27,15 @@ func (s *SeaweedFS) Install(ctx context.Context, kcli client.Client, writer *spi
...
 
+	airgapChartsPath := ""
+	if s.IsAirgap {
+		airgapChartsPath = runtimeconfig.EmbeddedClusterChartsSubDir()
+	}
+
 	hcli, err := helm.NewHelm(helm.HelmOptions{
 		KubeConfig: runtimeconfig.PathToKubeConfig(),
 		K0sVersion: versions.K0sVersion,
+		AirgapPath: airgapChartsPath,
```
sgalsaleh, line 38: why not override the `ChartPath` field below instead?

laverya: because that has a format like `oci://proxy.replicated.com/anonymous/...` and then I'd have to special case behavior based on the format of the string

Pretty sure the 'pull OCI' code would not appreciate being passed a filepath

pkg/helm/client.go
```diff
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
```
emosbaugh, line 9: use filepath

pkg/addons2/velero/velero.go
```diff
@@ -11,7 +11,8 @@ import (
 )
 
 type Velero struct {
-	Proxy *ecv1beta1.ProxySpec
+	Proxy    *ecv1beta1.ProxySpec
+	IsAirgap bool
```
emosbaugh, line 15: why is this unused?

cmd/installer/cli/install2.go
```diff
@@ -483,9 +491,21 @@ func recordInstallation(ctx context.Context, flags Install2CmdFlags, k0sCfg *k0s
...
+}
+
+func updateInstallation(ctx context.Context, install *ecv1beta1.Installation) error {
+	kcli, err := kubeutils.KubeClient()
+	if err != nil {
+		return fmt.Errorf("create kube client: %w", err)
+	}
+
+	if err := kubeutils.UpdateInstallation(ctx, kcli, install); err != nil {
+		return fmt.Errorf("Failed to update installation")
```
emosbaugh, line 507: i know the messaging will be weird to the end user but lets just keep this consistent

```suggestion
		return fmt.Errorf("update installation")
```

pkg/addons2/seaweedfs/install.go
emosbaugh: i agree with above. how about using url scheme `file://`?

e2e/install2_test.go
```diff
@@ -0,0 +1,288 @@
...
+
+	// TODO: check installation state
+
+	appUpgradeVersion := fmt.Sprintf("appver-%s-upgrade", os.Getenv("SHORT_SHA"))
+	testArgs := []string{appUpgradeVersion}
+
+	t.Logf("%s: upgrading cluster", time.Now().Format(time.RFC3339))
+	if stdout, stderr, err := tc.RunPlaywrightTest("deploy-upgrade", testArgs...); err != nil {
+		t.Fatalf("fail to run playwright test deploy-app: %v: %s: %s", err, stdout, stderr)
+	}
```
emosbaugh, line 51: i wouldnt call this done until youve implemented `check-postupgrade-state.sh`. Otherwise you are regressing functionality and its hidden.

emosbaugh (changes requested):

pkg/addons2/velero/velero.go
laverya: well that is certainly an argument in favor of centralizing the 'should this be airgap mode' flag

e2e/install2_test.go
laverya: I agree that it's not done yet, but it's still better than nothing
this just closes out the 'make airgap work' ticket while I continue on with the 'test install2' ticket

pkg/addons2/velero/velero.go
laverya: I've changed install2 to pass the helm client as a param, so airgap logic gets removed from the addon code

e2e/install2_test.go
laverya: added a basic check now

pkg/addons2/seaweedfs/install.go
laverya: I've refactored the code to only have to specify the AirgapPath once and then pass the resulting hcli object into each addon - that way the airgap logic doesn't have to be duplicated

Does that make sense to both of you?

sgalsaleh (approved):

emosbaugh (approved):

PR #1756, opened 2025-01-28
install operator with no custom values in install2

laverya:

This includes the operator chart in install2, and uses the CRD to store the installation kind in the cluster.

```release-note

```


cmd/installer/cli/install2.go
```diff
@@ -5,12 +5,15 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"sigs.k8s.io/yaml"
```
emosbaugh, line 8: Move this down

sgalsaleh (approved):

PR #1760, opened 2025-01-29
feat(v2): restore command

emosbaugh:

```release-note

```

### DIFF
--- a/cmd/installer/cli/restore.go
+++ b/cmd/installer/cli/restore.go
@@ -81,20 +81,9 @@ const (
 )
 
 func RestoreCmd(ctx context.Context, name string) *cobra.Command {
-	var (
-		airgapBundle            string
-		dataDir                 string
-		localArtifactMirrorPort int
-		networkInterface        string
-		assumeYes               bool
-		skipHostPreflights      bool
-		ignoreHostPreflights    bool
-		skipStoreValidation     bool
+	var flags Install2CmdFlags
 
-		s3Store = &s3BackupStore{}
-
-		proxy *ecv1beta1.ProxySpec
-	)
+	var s3Store = &s3BackupStore{}
 
 	cmd := &cobra.Command{
 		Use:           "restore",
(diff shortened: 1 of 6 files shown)


cmd/installer/cli/restore.go
```diff
@@ -81,446 +81,389 @@ const (
...
-				logrus.Debugf("configuring network manager")
-				if err := configureNetworkManager(cmd.Context()); err != nil {
-					return fmt.Errorf("unable to configure network manager: %w", err)
-				}
+	opts := addonsApplierOpts{
+		assumeYes:    flags.assumeYes,
+		license:      "",
+		airgapBundle: flags.airgapBundle,
+		overrides:    "",               // TODO: why not set this?
+		privateCAs:   flags.privateCAs, // TODO: this was changed, are we sure...
```
emosbaugh, line 192: i cant see why we wouldnt want this. preflights will fail

cmd/installer/cli/restore.go
```diff
@@ -81,446 +81,389 @@ const (
...
 
-				logrus.Debugf("configuring network manager")
-				if err := configureNetworkManager(cmd.Context()); err != nil {
-					return fmt.Errorf("unable to configure network manager: %w", err)
-				}
+	opts := addonsApplierOpts{
+		assumeYes:    flags.assumeYes,
+		license:      "",
+		airgapBundle: flags.airgapBundle,
+		overrides:    "",               // TODO: why not set this?
```
emosbaugh, line 191: should we add this to the restore command? im not sure the restore will apply the overrides in all cases.

cmd/installer/cli/restore.go
```diff
@@ -81,446 +81,389 @@ const (
...
-		assumeYes               bool
-		skipHostPreflights      bool
-		ignoreHostPreflights    bool
-		skipStoreValidation     bool
-
-		s3Store = &s3BackupStore{}
-
-		proxy *ecv1beta1.ProxySpec
-	)
+	var flags Install2CmdFlags
```
sgalsaleh, line 84: this feels wrong

cmd/installer/cli/restore.go
laverya: I think we should in case there are overrides for openebs or velero, no?

cmd/installer/cli/restore.go
```diff
@@ -81,446 +81,389 @@ const (
...
+		fallthrough
 
-			return nil
-		},
-	}
+	case ecRestoreStateRestoreECO:
+		logrus.Debugf("setting restore state to %q", ecRestoreStateRestoreECO)
+		if err := setECRestoreState(ctx, ecRestoreStateRestoreECO, backupToRestore.GetName()); err != nil {
+			return fmt.Errorf("unable to set restore state: %w", err)
+		}
```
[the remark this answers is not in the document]
laverya: (The whole point of EC extensions is to be present before the app is, so we can't expect vendors to restore them with the app IMO)

sgalsaleh (approved):

PR #1763, opened 2025-01-29
feat: preflight for at least 1000 fs.inotify.max_user_instances

laverya:

```release-note

```

### DIFF
--- a/pkg/preflights/host-preflight.yaml
+++ b/pkg/preflights/host-preflight.yaml
@@ -901,6 +901,15 @@ spec:
           - pass:
               when: 'net.ipv4.ip_forward > 0'
               message: "IP forwarding is enabled."
+    - sysctl:
+        checkName: "Adequate inotify max_user_instances"
+        outcomes:
+          - fail:
+              when: 'fs.inotify.max_user_instances < 1000'
+              message: "Inadequate inotify max_user_instances. To enable it, edit /etc/sysctl.conf, add or edit 'fs.inotify.max_user_instances' to a value of at least 1000, and run 'sudo sysctl -p'."
+          - pass:
+              when: 'fs.inotify.max_user_instances > 999'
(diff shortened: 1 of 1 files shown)


adamancini (commented):
lgtm

adamancini (approved):

ajp-io:
Why does the limit need to be this high? Can EC necessarily not work unless the limit is this high? When I get a GCP instance, the limit is 128. I'm hesitant to add preflights that will fail unless they're absolutely necessary.

ajp-io:
If it is absolutely necessary, we could also consider setting this sysctl like we do for ip forwarding, and then the preflight only fails if we can't set it ourselves at the beginning. That way we don't introduce a lot of friction with failing preflights.

adamancini:
@ajp-io I do think we should probably put it into a drop-in - there are some cases where sysctl dropins get overridden by end user configs, like via Puppet or Ansible - so I support a preflight, supportbundle analyzer, and we should install a sysctl drop-in config.

jtuchscherer:
Let's hold off on merging and talk through the implications of adding this pre-flight check

laverya:
> Why does the limit need to be this high? Can EC necessarily not work unless the limit is this high? When I get a GCP instance, the limit is 128. I'm hesitant to add preflights that will fail unless they're absolutely necessary.

I was just on a support case where 128 was preventing k0s from functioning properly - that's what prompted this

ajp-io:
>a more precise estimate probably depends heavily on the # of pods that you deploy and what they're doing

If this is customer-specific, would it be better to support vendor-supplied host preflight checks and have ITRS and anyone else who needs it set this themselves? ITRS seems like they have more charts (which I image translates to more pods) than many/most vendors. That makes me think many people won't be affected by this issue but will potentially face unneeded preflight failures anyway.

PR #1774, opened 2025-01-30
fix(v2): report preflight failures

emosbaugh:

Adds back preflight reporting.
Fixes empty ErrNothingElseToAdd errors.

```release-note

```

### DIFF
--- a/cmd/installer/cli/adminconsole_resetpassword.go
+++ b/cmd/installer/cli/adminconsole_resetpassword.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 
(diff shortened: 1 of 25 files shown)


cmd/installer/cli/adminconsole_resetpassword.go
```diff
@@ -13,10 +14,8 @@ import (
 
 func AdminConsoleResetPasswordCmd(ctx context.Context, name string) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:           "reset-password",
-		Short:         fmt.Sprintf("Reset the %s Admin Console password", name),
-		SilenceErrors: true,
-		SilenceUsage:  true,
```
emosbaugh, line 19: not needed because its set on the root command

pkg/metrics/reporter.go
```diff
@@ -23,6 +24,21 @@ import (
 var clusterIDMut sync.Mutex
 var clusterID *uuid.UUID
 
+// ErrorNoFail is an error that is excluded from metrics failures.
+type ErrorNoFail struct {
+	Err error
+}
```
laverya, line 30: why is this not in the types package?

laverya (approved):

PR #1782, opened 2025-01-31
fix TestLocalArtifactMirror and TestCommandsRequireSudo

laverya:

There are a few tests that do not pass a license to the install command. They all need to be either reworked or removed, as that is now required.

this fixes `TestLocalArtifactMirror` and `TestCommandsRequireSudo`.

`TestHostPreflightCustomSpec` and `TestUnsupportedOverrides` are still WIP.

```release-note

```


cmd/installer/cli/install2.go
```diff
@@ -211,6 +211,8 @@ func preRunInstall2(cmd *cobra.Command, flags *Install2CmdFlags) error {
 			return fmt.Errorf("unable to parse license file: %w", err)
 		}
 		flags.license = l
+	} else if requireLicense {
```
emosbaugh, line 214: Can you add this as well to addInstallFlags and MarkFlagRequired?

cmd/installer/cli/restore2.go
```diff
@@ -92,7 +92,7 @@ func Restore2Cmd(ctx context.Context, name string) *cobra.Command {
 		Use:   "restore",
 		Short: fmt.Sprintf("Restore a %s cluster", name),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			if err := preRunInstall2(cmd, &flags); err != nil {
+			if err := preRunInstall2(cmd, &flags, false); err != nil {
```
emosbaugh, line 95: is the license not required to restore?

laverya: I don't think so? the license is stored inside the admin console after all

emosbaugh: sure but what about custom domains? why not just make it required?

cmd/installer/cli/install2.go
laverya: There was a 'uncomment when we get tests passing' for requiring the license flag
I've done that now instead of this 'requireLicense' stuff

cmd/installer/cli/restore2.go
laverya: because the only thing we'd need it for is metrics and it feels wrong to require it for just metrics
easy to change later of course

emosbaugh (approved):

PR #1783, opened 2025-01-31
chore(v2): e2e test multiple upgrades

emosbaugh:

Tests a second upgrade after a v1 to v2 migration.

```release-note

```


emosbaugh:
[sc-119572]

e2e/install_test.go
```diff
@@ -522,15 +522,44 @@ func TestSingleNodeUpgradePreviousStable(t *testing.T) {
...
 	if stdout, stderr, err := tc.RunPlaywrightTest("deploy-upgrade", testArgs...); err != nil {
 		t.Fatalf("fail to run playwright test deploy-app: %v: %s: %s", err, stdout, stderr)
 	}
 
-	t.Logf("%s: checking installation state after upgrade", time.Now().Format(time.RFC3339))
+	// use upgraded binaries to run the commands
+	// TODO: this is a temporary workaround and should eventually be a feature of EC
+
+	t.Logf("%s: downloading embedded-cluster %s on node 0", time.Now().Format(time.RFC3339), appUpgradeVersion)
+	line = []string{"vandoor-prepare.sh", appUpgradeVersion, os.Getenv("LICENSE_ID"), "false"}
```
laverya, line 537: Why do you need to call vandoor-prepare here? Do we actually call the on-host embedded-cluster binary in our installation state checks?

For that matter, do we not upgrade that binary in the v2 release process anymore?

e2e/install_test.go
```diff
@@ -668,15 +697,44 @@ func TestUpgradeEC18FromReplicatedApp(t *testing.T) {
...
+	if stdout, stderr, err := tc.RunCommandOnNode(0, line); err != nil {
+		t.Fatalf("fail to download embedded-cluster version %s on node 0: %v: %s: %s", appUpgradeVersion, err, stdout, stderr)
+	}
```
laverya, line 715: same question here - what is this binay being used for?

e2e/install_test.go
laverya: oh we don't do it in online do we 🤦 

still raises the question of "what is it used for"

e2e/install_test.go
emosbaugh: We check the kots version with the kots cli and we need to update it or it throws an error that the cli version doesn't match the api version

e2e/install_test.go
emosbaugh: We check the kots version with the kots cli and we need to update it or it throws an error that the cli version doesn't match the api version

emosbaugh: Without this i get error

```
          • KOTS CLI version 1.124.0 does not match API version 1.124.1. To update, run:
```

e2e/install_test.go
emosbaugh: Without this i get error

```
          • KOTS CLI version 1.124.0 does not match API version 1.124.1. To update, run:
```

emosbaugh: i was wrong. removed

e2e/install_test.go
emosbaugh: i was wrong. removed

laverya (approved):

PR #1798, opened 2025-02-03
check chart contents directly instead of using k0s chart object

laverya:

```release-note

```

### DIFF
--- a/e2e/scripts/check-postupgrade-state.sh
+++ b/e2e/scripts/check-postupgrade-state.sh
@@ -54,17 +54,6 @@ main() {
         exit 1
     fi
 
-    # ensure that the nginx-ingress chart has been updated
-    local latest_nginx_chart_secret=
-    latest_nginx_chart_secret=$(kubectl get secrets -n ingress-nginx | grep helm | awk '{ print $1 }' | sort -r | head -n 1)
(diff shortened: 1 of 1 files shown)


e2e/scripts/check-postupgrade-state.sh
```diff
@@ -47,25 +54,29 @@ main() {
...
-    # fi
-    # # ensure new version is present
-    # if ! kubectl describe chart -n kube-system k0s-addon-chart-ingress-nginx | grep -q "4.12.0-beta.0"; then
-    #     echo "4.12.0-beta.0 not found in ingress-nginx chart"
-    #     exit 1
-    # fi
+    # ensure that the nginx-ingress chart has been updated
+    local latest_nginx_chart_secret=
+    latest_nginx_chart_secret=$(kubectl get secrets -n ingress-nginx | grep helm | awk '{ print $1 }' | sort -r | head -n 1)
+    if ! kubectl get secret -n ingress-nginx "$latest_nginx_chart_secret" -ojsonpath="{.data.release}" | base64 -d | base64 -d | gzip -d | grep -q "test-upgrade-value" ; then
```
sgalsaleh, line 60: do we really need this if we're already checking the applied service below?

laverya: I could go either way honestly

sgalsaleh (approved):

PR #1799, opened 2025-02-03
feat(v2): fix extensions apply

emosbaugh:

- Respects ForceUpgrade
- Respects order

```release-note

```

### DIFF
--- a/pkg/extensions/upgrade.go
+++ b/pkg/extensions/upgrade.go
@@ -11,32 +11,19 @@ import (
 	"github.com/replicatedhq/embedded-cluster/operator/pkg/k8sutil"
 	"github.com/replicatedhq/embedded-cluster/pkg/helm"
 	"github.com/replicatedhq/embedded-cluster/pkg/helpers"
-	"github.com/replicatedhq/embedded-cluster/pkg/runtimeconfig"
-	"github.com/replicatedhq/embedded-cluster/pkg/versions"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
-	actionInstall   = "Install"
-	actionUpgrade   = "Upgrade"
-	actionUninstall = "Uninstall"
+	actionInstall   = helmAction("Install")
+	actionUpgrade   = helmAction("Upgrade")
+	actionUninstall = helmAction("Uninstall")
 )
 
-func Upgrade(ctx context.Context, kcli client.Client, prev *ecv1beta1.Installation, in *ecv1beta1.Installation) error {
-	airgapChartsPath := ""
-	if in.Spec.AirGap {
-		airgapChartsPath = runtimeconfig.EmbeddedClusterChartsSubDir()
-	}
+type helmAction string
 
-	hcli, err := helm.NewClient(helm.HelmOptions{
-		K0sVersion: versions.K0sVersion,
-		AirgapPath: airgapChartsPath,
-	})
-	if err != nil {
-		return errors.Wrap(err, "create helm client")
-	}
-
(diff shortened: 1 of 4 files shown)


pkg/extensions/upgrade.go
```diff
@@ -47,6 +47,8 @@ func Upgrade(ctx context.Context, kcli client.Client, prev *ecv1beta1.Installati
 	// diff the extensions
 	diffResult := diffExtensions(prev.Spec.Config.Extensions, in.Spec.Config.Extensions)
 
+	// NOTE: this does not respect order and may add extensions prior to upgrading ones with a lower order.
```
sgalsaleh, line 50: order is respected in the `diffExtensions` function per operation though. How should it be respected otherwise?

emosbaugh: im not sure how it should be respected. but what if an extension needed to be upgraded first because it contained a new crd for a cr that was applied from another extension that was added?

pkg/extensions/upgrade.go
```diff
@@ -45,113 +32,170 @@ func Upgrade(ctx context.Context, kcli client.Client, prev *ecv1beta1.Installati
...
+	if err != nil {
+		return errors.Wrap(err, "check if extension is already processed")
+	} else if processed {
+		slog.Info(fmt.Sprintf("Extension %s already processed!", ext.Name))
+		return nil
+	}
+
+	actionIng, _ := formatAction(action)
+
+	slog.Info(fmt.Sprintf("%s extension %s", actionIng, ext.Name))
```
sgalsaleh, line 132: why not keep logging the version like before?

pkg/extensions/util.go
```diff
@@ -116,44 +120,40 @@ func diffExtensions(oldExts, newExts ecv1beta1.Extensions) DiffResult {
...
+		r := diffResult{
+			Ext: newChart,
+		}
+		oldChart, ok := oldCharts[name]
+		if !ok {
+			// chart was added.
+			r.Action = actionInstall
+		} else {
+			r.Action = actionUpgrade
+			r.NeedsUpgrade = !reflect.DeepEqual(oldChart, newChart)
```
sgalsaleh, line 135: any reason to not just `continue` here and skip this extension? seems weird to have the action be `upgrade` if the extensions doesn't need an upgrade

pkg/extensions/upgrade.go
emosbaugh: I am just trying to make it consistent. Should we go with all structured logging?

Other instances where it is not structured:

```
			slog.Info(fmt.Sprintf("%s already installed", ext.Name))
```

```
		slog.Info(fmt.Sprintf("%s is ready!", ext.Name))
```

pkg/extensions/util.go
emosbaugh: because otherwise the condition will be missing from the CR

pkg/extensions/upgrade.go
sgalsaleh: we should go with structured logging, but slog doesn't provide wrapping out of the box. this is how i had it before:
```golang
slog.Info(fmt.Sprintf("%s extension", actionIng), "name", ext.Name, "version", ext.Version)
```

pkg/extensions/util.go
sgalsaleh: is that a problem though? i think it's more confusing to had the CR show the extension was upgraded when it didn't need to be upgraded.

sgalsaleh: I guess if we want to be verbose, we can at least have a 4th condition to indicate it was a no-op, or maybe just a log.

sgalsaleh: nevermind, i guess this is fine

sgalsaleh (approved):

pkg/extensions/util.go
sgalsaleh: That's how addons are processed. We indicate they were upgraded even if there was no diff. So this is consistent.

PR #1821, opened 2025-02-05
fix(v2): retry enabling ha fails

emosbaugh:

fails with

```
unable to enable high availability: install seaweedfs: create prerequisites: create s3 service: create s3 service: Service "ec-seaweedfs-s3" is invalid: spec.clusterIPs: Invalid value: []string{"10.244.128.12"}: failed to allocate IP 10.244.128.12: provided IP is already allocated
```

```release-note

```

### DIFF
--- a/pkg/addons/seaweedfs/install.go
+++ b/pkg/addons/seaweedfs/install.go
@@ -101,18 +101,13 @@ func ensureService(ctx context.Context, kcli client.Client, serviceCIDR string) 
 	obj.ObjectMeta.Labels = ApplyLabels(obj.ObjectMeta.Labels, "s3")
 
 	var existingObj corev1.Service
-	if err := kcli.Get(ctx, client.ObjectKey{Name: obj.Name, Namespace: obj.Namespace}, &existingObj); err != nil {
-		if !k8serrors.IsNotFound(err) {
-			return errors.Wrap(err, "get s3 service")
+	if err := kcli.Get(ctx, client.ObjectKey{Name: obj.Name, Namespace: obj.Namespace}, &existingObj); err != nil && !k8serrors.IsNotFound(err) {
+		return errors.Wrap(err, "get s3 service")
+	} else if err == nil {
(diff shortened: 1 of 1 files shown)


pkg/addons/seaweedfs/install.go
```diff
@@ -100,6 +100,21 @@ func createService(ctx context.Context, kcli client.Client, serviceCIDR string)
...
+		}
+	} else {
+		if existingObj.Spec.ClusterIP != clusterIP {
+			existingObj.Spec.ClusterIP = clusterIP
+			if err := kcli.Update(ctx, &existingObj); err != nil {
+				return errors.Wrap(err, "update s3 service")
+			}
+		}
+		return nil
+	}
```
laverya, line 116: I think you need to check if the current clusterIP matches the existing, and if it does not delete + recreate
IIRC changing the specified IP isn't a thing that can be done

pkg/addons/seaweedfs/install.go
```diff
@@ -100,6 +100,16 @@ func createService(ctx context.Context, kcli client.Client, serviceCIDR string)
 
 	obj.ObjectMeta.Labels = ApplyLabels(obj.ObjectMeta.Labels, "s3")
 
+	var existingObj corev1.Service
+	if err := kcli.Get(ctx, client.ObjectKey{Name: obj.Name, Namespace: obj.Namespace}, &existingObj); err != nil && !k8serrors.IsNotFound(err) {
+		return errors.Wrap(err, "get s3 service")
+	} else if err == nil {
+		err := kcli.Delete(ctx, &existingObj)
```
laverya, line 107:
```suggestion
	} else if err == nil {
		if existingObj.Spec.ClusterIP == clusterIP {
			return nil
		fi
		err := kcli.Delete(ctx, &existingObj)
```

We shouldn't delete+recreate if it's already fine

laverya (approved):

PR #1832, opened 2025-02-06
do not build a separate binary for the legacy-dr test suite

laverya:

Instead of building a legacy-dr embedded cluster binary, use the normal binary and rely on replicated.app to embed the relevant app yaml files into it.

```release-note

```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -213,75 +213,6 @@ jobs:
           echo "K0S_VERSION=\"$K0S_VERSION\""
           echo "k0s_version=$K0S_VERSION" >> "$GITHUB_OUTPUT"
 
-  build-legacydr:
-    name: Build legacy DR
-    runs-on: embedded-cluster-2
-    needs:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -636,12 +566,6 @@ jobs:
           export RELEASE_YAML_DIR=e2e/kots-release-install
           ./scripts/ci-release-app.sh
 
-          # promote a release with improved dr support
-          export EC_VERSION="$(git describe --tags --abbrev=4 --match='[0-9]*.[0-9]*.[0-9]*')-legacydr"
-          export APP_VERSION="appver-${SHORT_SHA}-legacydr"
```
emosbaugh, line 641: do we not test this in airgap?

emosbaugh (approved):

.github/workflows/ci.yaml
laverya: we don't test this in airgap apparently

PR #1834, opened 2025-02-06
Fix cleaning up helm values before install / upgrade

sgalsaleh:

Fixes an issue when cleaning up helm values before install / upgrade where `[]map[string]interface{}` types weren't handled, which broke admin console upgrades in some scenarios.

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/helm/client.go
+++ b/pkg/helm/client.go
@@ -545,6 +545,8 @@ func cleanUpMapValue(v interface{}) interface{} {
 	switch v := v.(type) {
 	case []interface{}:
 		return cleanUpInterfaceArray(v)
+	case []map[string]interface{}:
+		return cleanUpMapInterfaceArray(v)
 	case map[string]interface{}:
 		return cleanUpGenericMap(v)
 	case map[interface{}]interface{}:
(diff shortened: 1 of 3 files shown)


pkg/helm/client.go
```diff
@@ -545,6 +545,8 @@ func cleanUpMapValue(v interface{}) interface{} {
 	switch v := v.(type) {
 	case []interface{}:
 		return cleanUpInterfaceArray(v)
+	case []map[string]interface{}:
+		return cleanUpMapInterfaceArray(v)
```
laverya, line 549: should we also add
```suggestion
	case []map[string]interface{}:
		return cleanUpMapInterfaceArray(v)
	case []map[interface{}]interface{}:
		return cleanUpMapInterfaceArray(v)
```

pkg/helm/client.go
```diff
@@ -571,6 +573,15 @@ func cleanUpInterfaceArray(in []interface{}) []interface{} {
 	return result
 }
 
+// Cleans up a slice of map to interface into slice of actual values
+func cleanUpMapInterfaceArray(in []map[string]interface{}) []map[string]interface{} {
```
laverya, line 577:
```suggestion
func cleanUpMapInterfaceArray(in []map[interface{}]interface{}) []map[string]interface{} {
```

pkg/helm/client.go
```diff
@@ -571,6 +573,15 @@ func cleanUpInterfaceArray(in []interface{}) []interface{} {
...
+	result := make([]map[string]interface{}, len(in))
+	for i, v := range in {
+		result[i] = cleanUpGenericMap(v)
+	}
+	return result
+}
```
laverya, line 583:
```suggestion
func cleanUpMapInterfaceArray(in []map[interface{}]interface{}) []map[string]interface{} {
	result := make([]map[string]interface{}, len(in))
	for i, v := range in {
		result[i] = cleanUpInterfaceMap(v)
	}
	return result
}
```

laverya (approved):

PR #1838, opened 2025-02-06
chore: fix generic map conversion

emosbaugh:

Makes map conversion more resilient. No longer have to add new case statements for each type.

```release-note

```

### DIFF
--- a/pkg/helm/client.go
+++ b/pkg/helm/client.go
@@ -28,6 +28,7 @@ import (
 	"helm.sh/helm/v3/pkg/storage/driver"
 	"helm.sh/helm/v3/pkg/uploader"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	k8syaml "sigs.k8s.io/yaml"
 )
 
 type RESTClientGetterFactory func(namespace string) genericclioptions.RESTClientGetter
@@ -356,7 +357,10 @@ func (h *HelmClient) Install(ctx context.Context, opts InstallOptions) (*release
 		}
 	}
 
-	cleanVals := cleanUpGenericMap(opts.Values)
+	cleanVals, err := cleanUpGenericMap(opts.Values)
+	if err != nil {
+		return nil, fmt.Errorf("clean up generic map: %w", err)
+	}
 
 	release, err := client.RunWithContext(ctx, chartRequested, cleanVals)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/helm/client.go
```diff
@@ -526,80 +536,17 @@ func (h *HelmClient) getRESTClientGetter(namespace string) genericclioptions.RES
...
-	return result
-}
-
-// Cleans up a slice of map to interface into slice of actual values
-func cleanUpInterfaceMapArray(in []map[interface{}]interface{}) []map[string]interface{} {
-	result := make([]map[string]interface{}, len(in))
-	for i, v := range in {
-		result[i] = cleanUpInterfaceMap(v)
+func cleanUpGenericMap(m map[string]interface{}) (map[string]interface{}, error) {
+	b, err := yaml.Marshal(m)
```
sgalsaleh, line 540: why not k8syaml?

pkg/helm/client_test.go
```diff
@@ -102,11 +102,11 @@ func Test_cleanUpGenericMap(t *testing.T) {
 			want: map[string]interface{}{
 				"nest": map[string]interface{}{
 					"abc":    "xyz",
-					"number": 5,
+					"number": float64(5),
 					"float":  1.5,
 					"bool":   true,
-					"array": []map[string]interface{}{
-						{
+					"array": []interface{}{
```
sgalsaleh, line 108: think helm will accept an `[]interface{}`?

pkg/helm/client.go
emosbaugh: i think because k8syaml cannot marshal a map[interface{}]interface{}. i will test and add a comment

pkg/helm/client_test.go
emosbaugh: i assume since its passing tests...

sgalsaleh: our e2e tests never caught this bug, which is why these unit tests were introduced. so i wouldn't be so confident...

sgalsaleh: if `want` did not change at all in these unit tests, i would be more confident in the change.

pkg/helm/client.go
emosbaugh: i added comments to clarify

pkg/helm/client_test.go
emosbaugh: i added comments to clarify

pkg/helm/client_test.go
```diff
@@ -149,7 +150,14 @@ func Test_cleanUpGenericMap(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			req := require.New(t)
-			req.Equal(tt.want, cleanUpGenericMap(tt.in))
+			out, err := cleanUpGenericMap(tt.in)
+			req.NoError(err, "cleanUpGenericMap failed")
+			req.Equal(tt.want, out)
+
+			// ultimately helm calls k8syaml.Marshal so we must make sure that the output is compatible
```
sgalsaleh, line 157: did you verify (even if manually) that this would've caught the issue fixed by https://github.com/replicatedhq/embedded-cluster/pull/1834?

emosbaugh: How can I verify? Can you provide reproduction steps? They are not in the original fix.

sgalsaleh (approved):

PR #1849, opened 2025-02-07
configure firewalld automatically

ajp-io:

```release-note

```


cmd/installer/goods/systemd/firewalld-config.tpl.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<zone target="ACCEPT">
```
ajp-io, line 2: k0s shows `default` instead of `ACCEPT`, but default didn't work for me. since this is scoped to the cidrs, accept might be okay.

PR #1861, opened 2025-02-10
fix: support for amazon linux 2

emosbaugh:

```release-note
Fixes an issue when installing on Amazon Linux 2 and other older Linux distributions that causes the installation to timeout waiting for storage to be ready.
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -109,7 +109,9 @@ func InstallFlags(nodeIP string) []string {
 
 func AdditionalInstallFlags(nodeIP string) []string {
 	return []string{
-		"--kubelet-extra-args", fmt.Sprintf(`"--node-ip=%s"`, nodeIP),
+		// NOTE: quotes are not supported in older systemd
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -109,7 +109,9 @@ func InstallFlags(nodeIP string) []string {
 
 func AdditionalInstallFlags(nodeIP string) []string {
 	return []string{
-		"--kubelet-extra-args", fmt.Sprintf(`"--node-ip=%s"`, nodeIP),
+		// NOTE: quotes are not supported in older systemd
+		// kardianos/service will escape spaces with "\x20"
+		"--kubelet-extra-args", fmt.Sprintf("--node-ip=%s", nodeIP),
```
sgalsaleh, line 114: are you sure this sets the node ip correctly though? i intentionally used backticks because this didn’t work IIRC

sgalsaleh (approved):

PR #1865, opened 2025-02-11
Default rp_filter to loose mode (2) instead of disabled (0)

sgalsaleh:

Looks like Calico [doesn't rely on sysctls](https://github.com/projectcalico/calico/blob/v3.12.0/release-notes/v3.12.0-release-notes.md#other-changes) to determine RPF anymore, so setting it to `2` shouldn't be an issue and is what's recommended and common.

This PR makes it so that we try to set `rp_filter` to loose mode (`2`) instead of disabled (`0`), but if we can't, we only fail if `rp_filter` is set to strict mode (`1`).

[SC-117729](https://app.shortcut.com/replicated/story/117729)

```release-note
Adds preflight checks to ensure that reverse path filtering (rp_filter) is either disabled (0) or set to loose mode (2) for both default and all interfaces via kernel parameters `net.ipv4.conf.default.rp_filter` and `net.ipv4.conf.all.rp_filter`.
```

NONE

### DIFF
--- a/pkg/configutils/static/99-embedded-cluster.conf
+++ b/pkg/configutils/static/99-embedded-cluster.conf
@@ -14,5 +14,5 @@ net.ipv4.conf.all.arp_ignore = 0
 # particularly when communicating with cluster services from the host. This is
 # because Kubernetes networking paths are not always seen as the "best reverse
 # path" by the kernel.
-net.ipv4.conf.default.rp_filter = 0
-net.ipv4.conf.all.rp_filter = 0
+net.ipv4.conf.default.rp_filter = 2
+net.ipv4.conf.all.rp_filter = 2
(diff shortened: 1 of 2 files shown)


pkg/configutils/static/99-embedded-cluster.conf
```diff
@@ -14,5 +14,5 @@ net.ipv4.conf.all.arp_ignore = 0
 # particularly when communicating with cluster services from the host. This is
 # because Kubernetes networking paths are not always seen as the "best reverse
 # path" by the kernel.
```
laverya, line 16: I feel like this comment is now inaccurate, no?

sgalsaleh: yes, let me update. ty

laverya (dismissed):

laverya (approved):

PR #1883, opened 2025-02-14
fix(lam): wait for service after start

emosbaugh:

Checks systemctl status of LAM after installing and starting before proceeding with the rest of the install.

```release-note

```

### DIFF
--- a/cmd/installer/cli/install.go
+++ b/cmd/installer/cli/install.go
@@ -608,7 +608,7 @@ func installAndStartCluster(ctx context.Context, networkInterface string, airgap
 		return nil, fmt.Errorf("create config file: %w", err)
 	}
 	logrus.Debugf("creating systemd unit files")
-	if err := createSystemdUnitFiles(false, proxy); err != nil {
+	if err := createSystemdUnitFiles(ctx, false, proxy); err != nil {
 		return nil, fmt.Errorf("create systemd unit files: %w", err)
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/installer/kotscli/kotscli.go
```diff
@@ -78,21 +75,14 @@ func Install(opts InstallOptions, msg *spinner.MessageWriter) error {
...
 	defer msg.SetLineBreaker(nil)
 
-	stdout := bytes.NewBuffer(nil)
-	stderr := bytes.NewBuffer(nil)
-
 	runCommandOptions := helpers.RunCommandOptions{
-		Stdout: io.MultiWriter(msg, stdout),
-		Stderr: stderr,
+		Stdout:       msg,
+		LogOnSuccess: true,
```
emosbaugh, line 80: i was not aware of this option before

cmd/installer/cli/install.go
```diff
@@ -839,15 +839,50 @@ func installAndEnableLocalArtifactMirror() error {
...
 
+func waitForLocalArtifactMirror(ctx context.Context) error {
+	consecutiveSuccesses := 0
+	requiredSuccesses := 3
+	maxAttempts := 30
+	checkInterval := 2 * time.Second
+
+	var lastErr error
+	for attempt := 0; attempt < maxAttempts; attempt++ {
+		_, err := helpers.RunCommand("systemctl", "status", "local-artifact-mirror")
```
laverya, line 864: should this use the same expbackoff with jitter as we do elsewhere? I see we're looking for consecutive successes and I'm not sure if that would work though

cmd/installer/kotscli/kotscli.go
laverya: oh neat

laverya (approved):

PR #1892, opened 2025-02-18
feat: configure firewalld

emosbaugh:

Configures firewalld for single and multi node installations.

If there is an error it will log and move on.

```release-note
Automatically configures firewalld to allow the pod and service networks to communicate, as well as opens host ports for Kubernetes networking.
```


cmd/installer/cli/firewalld.go
```diff
@@ -0,0 +1,178 @@
...
+		return fmt.Errorf("add pod network source: %w", err)
+	}
+
+	err = firewalld.AddSourceToZone(ctx, serviceNetwork, opts...)
+	if err != nil {
+		return fmt.Errorf("add service network source: %w", err)
+	}
+
+	// Add the calico interfaces
+	// This is redundant and overlaps with the pod network but we add it anyway
```
sgalsaleh, line 115: why though?

e2e/scripts/yum-install-firewalld.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+yum update -y
+yum install -y firewalld
```
sgalsaleh, line 5: why not include this in the docker image(s)?

cmd/installer/cli/firewalld.go
emosbaugh: well i cant see any harm in it. and because it is what was recommended on calico issues but preflights will not pass without adding sources since the interfaces do not exist when running preflights.

e2e/scripts/yum-install-firewalld.sh
emosbaugh: updated

sgalsaleh (approved):

PR #1894, opened 2025-02-19
fix(host-preflights): buildtin kernel modules file from wrong path

emosbaugh:

Updates troubleshoot with this change

https://github.com/replicatedhq/troubleshoot/pull/1741

```release-note
Fixes host preflight failures for kernel modules in environments where kernel modules are built in.
```


emosbaugh:
This will not pass until this change is merged
https://github.com/replicatedhq/embedded-cluster/pull/1892

sgalsaleh (approved):

PR #1898, opened 2025-02-19
feat(host-preflights): check that the data directory is executable

emosbaugh:

Screenshot 2025-02-20 at 9 24 00 PM

```release-note
Adds a host preflight to check that the data directory is executable.
```

### DIFF
--- a/cmd/installer/cli/install_runpreflights.go
+++ b/cmd/installer/cli/install_runpreflights.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 
 	"github.com/replicatedhq/embedded-cluster/pkg/configutils"
 	"github.com/replicatedhq/embedded-cluster/pkg/netutils"
@@ -108,6 +109,10 @@ func runInstallPreflights(ctx context.Context, flags InstallCmdFlags, metricsRep
 		AssumeYes:            flags.assumeYes,
 		MetricsReporter:      metricsReported,
 	}); err != nil {
+		if errors.Is(err, os.ErrPermission) {
+			logrus.Errorf("Please make sure that the filesystem at %s is executable.", runtimeconfig.EmbeddedClusterHomeDirectory())
+			return NewErrorNothingElseToAdd(err)
+		}
 		return err
 	}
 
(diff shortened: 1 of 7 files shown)


cmd/installer/cli/install_runpreflights.go
```diff
@@ -108,6 +109,10 @@ func runInstallPreflights(ctx context.Context, flags InstallCmdFlags, metricsRep
 		AssumeYes:            flags.assumeYes,
 		MetricsReporter:      metricsReported,
 	}); err != nil {
+		if errors.Is(err, os.ErrPermission) {
+			logrus.Errorf("Please make sure that the filesystem at %s is executable.", runtimeconfig.EmbeddedClusterHomeDirectory())
```
sgalsaleh, line 113: this could be an inaccurate message though if it's the binary (or one of the subdirectories / parent directories) that isn't executable, right?

sgalsaleh: Also, it might be executable for root, but not for group or other which is what's necessary for etcd for example.

sgalsaleh: Ok, yeah I guess that's fine then. But I remember a discussion around etcd being run by a different user which might not have execute permissions on one of the folders in the path and I don't think this error will catch that, but I think that is a separate problem to this and can be fixed in a different PR.

sgalsaleh (approved):

cmd/installer/cli/install_runpreflights.go
```diff
@@ -108,6 +109,10 @@ func runInstallPreflights(ctx context.Context, flags InstallCmdFlags, metricsRep
 		AssumeYes:            flags.assumeYes,
 		MetricsReporter:      metricsReported,
 	}); err != nil {
+		if errors.Is(err, os.ErrPermission) {
```
emosbaugh, line 112: move this to the root command

cmd/installer/cli/join_runpreflights.go
```diff
@@ -110,6 +111,10 @@ func runJoinPreflights(ctx context.Context, jcmd *kotsadm.JoinCommandResponse, f
 		TCPConnectionsRequired: jcmd.TCPConnectionsRequired,
 		IsJoin:                 true,
 	}); err != nil {
+		if errors.Is(err, os.ErrPermission) {
```
emosbaugh, line 114: move this to the root command

sgalsaleh (approved):

PR #1901, opened 2025-02-21
Add preflight check for directory permissions to prevent etcd startup failures

sgalsaleh:

This PR ensures that the k0s data directory and its parent directories have execute permissions. This is necessary for the k0s etcd user to execute the etcd binary. Without proper execute permissions, the etcd service will fail to start, causing cluster initialization failures.

Fixes:
- [SC-120515](https://app.shortcut.com/replicated/story/120515)
- [SC-120528](https://app.shortcut.com/replicated/story/120528)

```release-note
* Adds a preflight check to verify execute permissions on the data directory and its parent directories, preventing cluster startup failures due to etcd permission issues.
```

NONE

### DIFF
--- a/pkg/runtimeconfig/runtimeconfig.go
+++ b/pkg/runtimeconfig/runtimeconfig.go
@@ -34,10 +34,16 @@ func Cleanup() {
 // EmbeddedClusterHomeDirectory returns the parent directory. Inside this parent directory we
 // store all the embedded-cluster related files.
 func EmbeddedClusterHomeDirectory() string {
+	dir := ecv1beta1.DefaultDataDir
 	if runtimeConfig.DataDir != "" {
-		return runtimeConfig.DataDir
+		dir = runtimeConfig.DataDir
 	}
-	return ecv1beta1.DefaultDataDir
+	if err := os.Chmod(dir, 0755); err != nil {
+		// don't fail as there are cases where we can't change the permissions (bind mounts, selinux, etc...),
+		// and we handle and surface those errors to the user later (host preflights, checking exec errors, etc...)
+		logrus.Debugf("unable to chmod embedded-cluster home dir: %s", err)
+	}
+	return dir
 }
 
 // EmbeddedClusterTmpSubDir returns the path to the tmp directory where embedded-cluster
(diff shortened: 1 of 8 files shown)


ajp-io:
do you have a screenshot of what that failure message looks like in your terminal? i'd like to see it rendered as opposed to just the templated stuff in the code.

pkg/runtimeconfig/runtimeconfig.go
```diff
@@ -34,10 +34,16 @@ func Cleanup() {
 // EmbeddedClusterHomeDirectory returns the parent directory. Inside this parent directory we
 // store all the embedded-cluster related files.
 func EmbeddedClusterHomeDirectory() string {
```
laverya, line 36: how often is this function called? I just don't want to run this 100 times

laverya (approved):

pkg/runtimeconfig/runtimeconfig.go
```diff
@@ -34,10 +34,16 @@ func Cleanup() {
...
+		dir = runtimeConfig.DataDir
 	}
-	return ecv1beta1.DefaultDataDir
+	if err := os.Chmod(dir, 0755); err != nil && !os.IsNotExist(err) {
+		// don't fail as there are cases where we can't change the permissions (bind mounts, selinux, etc...),
+		// and we handle and surface those errors to the user later (host preflights, checking exec errors, etc...)
+		logrus.Debugf("unable to chmod embedded-cluster home dir: %s", err)
+	}
+	return dir
 }
```
laverya, line 47:
```suggestion
	if chmodHasRun {
		return dir
	}
	if err := os.Chmod(dir, 0755); err != nil && !os.IsNotExist(err) {
		// don't fail as there are cases where we can't change the permissions (bind mounts, selinux, etc...),
		// and we handle and surface those errors to the user later (host preflights, checking exec errors, etc...)
		logrus.Debugf("unable to chmod embedded-cluster home dir: %s", err)
	} else if err == nil {
		// only set that the chmod has run if the directory existed and there was no error
		chmodHasRun = true
	}
	
	return dir
}
var chmodHasRun bool
```

something like this

laverya (changes requested):

laverya (approved):

PR #1915, opened 2025-02-27
worker profiles investigation

laverya:

```release-note

```


hedge-sparrow:
To pick this back up, the plan is to change the config section so that we have one single globally defined profile that will be installed as the `default` worker profile on all nodes if defined.

hedge-sparrow:
Ok so the plan going forward should be to take this work and simplify the config sections so that only a single worker profile can be defined and applied to all nodes, but we should validate the name so the user can't override the `default` worker profile.

PR #1917, opened 2025-02-27
chore(ci): lxd copy dir tar

emosbaugh:

Tars directories before copying them to lxc container rather than copying files one at a time.

```release-note

```

### DIFF
--- a/e2e/cluster/lxd/cluster.go
+++ b/e2e/cluster/lxd/cluster.go
@@ -1,6 +1,7 @@
 package lxd
 
 import (
+	"archive/tar"
 	"bytes"
 	"context"
 	"fmt"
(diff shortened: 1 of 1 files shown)


e2e/cluster/lxd/cluster.go
```diff
@@ -498,28 +499,115 @@ func CopyDirsToNode(in *ClusterInput, node string) {
...
 }
 
-// CopyDirToNode copies a single directory to a node.
+// CopyDirToNode copies a single directory to a node by creating a tar archive and streaming it
 func CopyDirToNode(in *ClusterInput, node string, dir Dir) {
+	// Create a temporary tar file
+	tmpFile, err := os.CreateTemp("", "dir-*.tar")
+	if err != nil {
+		in.T.Fatalf("Failed to create temp file: %v", err)
+	}
```
laverya, line 508: why is this a real file and not a bytes.buffer/pipe that we stream through? The filepath.walk should be in a goroutine and we can do read+tar+stream-to-lxd in one pass

laverya (approved):

PR #1922, opened 2025-02-27
chore(ci): concurrent node create

emosbaugh:

Speed up e2e tests

```release-note

```

### DIFF
--- a/e2e/cluster/docker/cluster.go
+++ b/e2e/cluster/docker/cluster.go
@@ -71,16 +71,9 @@ func NewNode(in *ClusterInput, name string) *Container {
 }
 
 func (c *Cluster) Run() {
-	wg := sync.WaitGroup{}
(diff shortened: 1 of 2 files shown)


e2e/cluster/docker/cluster.go
```diff
@@ -28,13 +28,24 @@ type ClusterInput struct {
...
+		go func(i int) {
+			defer wg.Done()
+			node := NewNode(in, fmt.Sprintf("node%d", i))
+			if i == 0 {
+				node = node.WithPort("30003:30003")
+			}
+			c.Nodes[i] = node
+		}(i)
 	}
+	wg.Wait()
```
sgalsaleh, line 47: I don't think NewNode runs anything. You'll probably need to make your changes in the "Run" function below as that's what starts the containers.

PR #1961, opened 2025-03-13
add embedded cluster custom domain types, use those domains for preflights

laverya:

This is the first part of the custom domains ticket - the second largely requires kotsadm to be updated to include the type updates in this PR.

This contains code to add the new fields to the cluster config type, and adds code changes to allow for the new custom app and proxy domains to be used in preflights but not fully in installations.

```release-note

```


emosbaugh (approved):

laverya:
I'm going to merge this despite the alma linux test failures - those are unrelated to this code and appear to have been from a change in the OS image we're running

PR #1962, opened 2025-03-13
feat(adminconsole): support for custom domains

emosbaugh:

Passes custom domains as admin console chart values.

```release-note

```


emosbaugh:
Merging as test failures are unrelated.

PR #1974, opened 2025-03-17
worker profiles (#1915)

hedge-sparrow:

The current state of this PR is that worker profiles can now be defined in unsupported overrides, and the first defined worker profile is always applied on install. 

TODO: 

- [ ] Investigate kots changes required to enable worker profiles being set on join
- [x] confirm config UX 

```release-note

```


pkg/config/config.go
```diff
@@ -102,9 +103,10 @@ func InstallFlags(nodeIP string) []string {
 		"--no-taints",
 		"-c", runtimeconfig.PathToK0sConfig(),
 	}
+	flags = append(flags, ProfileInstallFlag(cfg))
 	flags = append(flags, AdditionalInstallFlags(nodeIP)...)
 	flags = append(flags, AdditionalInstallFlagsController()...)
-	return flags
+	return flags, nil
```
laverya, line 109: if you only ever return nil better to leave it without an error in the signature

e2e/kots-release-install/cluster-config.yaml
```diff
@@ -40,6 +40,11 @@ spec:
         spec:
           telemetry:
             enabled: false
+          workerProfiles:
+          - name: ip-forward
+            values:
+              allowedUnsafeSysctls:
+              - net.ipv4.ip_forward
```
laverya, line 47: I would prefer that we add this to the `-upgrade` and `-install-noop` directories too

pkg/config/config.go
```diff
@@ -38,8 +39,8 @@ func RenderK0sConfig(proxyRegistryDomain string) *k0sconfig.ClusterConfig {
 	return cfg
 }
 
-// extractK0sConfigPatch extracts the k0s config portion of the provided patch.
-func extractK0sConfigPatch(raw string) (string, error) {
+// ExtractK0sConfigPatch extracts the k0s config portion of the provided patch.
+func ExtractK0sConfigPatch(raw string) (string, error) {
```
laverya, line 43: Why was this promoted to an external function? I don't see it called anywhere new

pkg/config/config.go
```diff
@@ -123,6 +125,14 @@ func AdditionalInstallFlagsController() []string {
 	}
 }
 
+func ProfileInstallFlag(cfg *k0sconfig.ClusterConfig) string {
+	controllerProfile := controllerWorkerProfile()
+	if controllerProfile == "" {
+		return ""
+	}
+	return "--profile=" + controllerProfile
+}
```
laverya, line 134: if this doesn't use the cluster config it shouldn't be passed the clusterconfig

pkg/kotsadm/types.go
```diff
@@ -35,6 +35,10 @@ func (j JoinCommandResponse) extractK0sConfigOverridePatch(data []byte) (dig.Map
 	if storage := config.DigMapping("config", "spec", "storage"); len(storage) > 0 {
 		result.DigMapping("config", "spec")["storage"] = storage
 	}
+	workerProfiles := config.Dig("config", "spec", "workerProfiles")
```
laverya, line 38: mind updating the comment for this function too?

laverya (dismissed):

pkg/config/config.go
```diff
@@ -161,6 +171,25 @@ func additionalControllerLabels() map[string]string {
 	return map[string]string{}
 }
 
+func controllerWorkerProfile() string {
```
emosbaugh, line 174: this should return an error and we should handle this error

cmd/installer/cli/join.go
```diff
@@ -459,9 +462,26 @@ func applyJoinConfigurationOverrides(jcmd *kotsadm.JoinCommandResponse) error {
 	return nil
 }
 
+func getFirstDefinedProfile() string {
```
emosbaugh, line 465: this should return an error and we should handle this error

emosbaugh (changes requested):

emosbaugh (approved):

emosbaugh (approved):

PR #1979, opened 2025-03-17
chore(ci): use GITHUB_TOKEN to avoid rate limits

emosbaugh:

- Use GITHUB_TOKEN to avoid rate limiting
- increases restore timeouts
- Outputs tcp denied logs url path

```release-note

```


cmd/installer/goods/bins/.placeholder
laverya, line 1: I think keeping these is good for local dev reasons (it keeps the directory around)

laverya (approved):

PR #1989, opened 2025-03-18
Default to release domains and don't fallback to license

sgalsaleh:

Defaults to release domains embedded in the binary (or installation) and doesn't fall back to license

Yes

```release-note
NONE
```

NONE

### DIFF
--- a/cmd/installer/cli/install.go
+++ b/cmd/installer/cli/install.go
@@ -34,6 +34,7 @@ import (
 	"github.com/replicatedhq/embedded-cluster/pkg/k0s"
 	"github.com/replicatedhq/embedded-cluster/pkg/kubeutils"
 	"github.com/replicatedhq/embedded-cluster/pkg/metrics"
+	"github.com/replicatedhq/embedded-cluster/pkg/netutil"
 	"github.com/replicatedhq/embedded-cluster/pkg/netutils"
 	"github.com/replicatedhq/embedded-cluster/pkg/preflights"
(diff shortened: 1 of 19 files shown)


cmd/installer/cli/install.go
```diff
@@ -780,6 +779,24 @@ func validateAdminConsolePassword(password, passwordCheck string) bool {
...
+}
+
+func proxyRegistryURL() string {
+	var embCfgSpec *ecv1beta1.ConfigSpec
+	if embCfg := release.GetEmbeddedClusterConfig(); embCfg != nil {
+		embCfgSpec = &embCfg.Spec
+	}
+	domains := runtimeconfig.GetDomains(embCfgSpec)
+	return netutil.MaybeAddHTTPS(domains.ProxyRegistryDomain)
+}
```
laverya, line 798: why aren't these in the runtimeconfig package?

For that matter, if you're not going to return an error here, why not do that for a GetDomains function?

sgalsaleh: it's an intentional decision to put these here in the cli package because they use the embedded config spec, whereas other packages might have to use the installation config spec, and i didn't want us getting this from runtimeconfig by mistake.

sgalsaleh: because that's hard to read :)

sgalsaleh: also, `GetDomains` takes a config spec, not a config

laverya (approved):

PR #1990, opened 2025-03-19
test without custom domains in cluster config

laverya:

```release-note

```


laverya:
This was actually much more successful than I expected - the replicated.app tests worked, it's just the locally bundled ones that failed
@sgalsaleh your changes were effective

sgalsaleh:
We should probably do a similar test when Diamon merges his changes with the new charts to make sure they work without custom domains. Unless his PR already covers that.

PR #2001, opened 2025-03-20
remove 'http_access allow localnet' from test squid config

laverya:

AFAICT this line isn't required for our tests, but it makes the proxy fail-open if you use this config in GCP

```release-note

```


laverya:
apparently that is required in CI, never mind

PR #2002, opened 2025-03-20
chore(proxy): fix bad squid conf

emosbaugh:

```release-note

```

### DIFF
--- a/e2e/scripts/install-and-configure-squid.sh
+++ b/e2e/scripts/install-and-configure-squid.sh
@@ -11,7 +11,11 @@ ssl_bump bump all
 
 acl whitelist dstdomain \"/etc/squid/sites.whitelist.txt\"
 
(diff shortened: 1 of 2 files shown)


e2e/scripts/install-and-configure-squid.sh
```diff
@@ -9,10 +9,9 @@ acl step1 at_step SslBump1
 ssl_bump peek step1
 ssl_bump bump all
 
-http_access allow localnet
-
 acl whitelist dstdomain \"/etc/squid/sites.whitelist.txt\"
-http_access allow whitelist
+
+http_access allow localnet
```
laverya, line 14:
```suggestion

# this will allow all access to the internet from local IPs
http_access allow localnet

# to restrict access so only local IPs can access the internet and only sites on the whitelist, instead use
# http_access allow localnet whitelist
```

PR #2011, opened 2025-03-24
chore: print sbctl shell instead of sbctl serve in GitHub actions log

banjoh:
Instruct users to use `sbctl shell` instead of `sbctl serve` in GitHub actions log. This was users do not have to create a new shell to run kubectl commands.

```release-note
NONE
```


sgalsaleh:
Merging as this PR doesn't affect tests.

PR #2012, opened 2025-03-24
feat(ha): node local loadbalancing

emosbaugh:

Adds no local load balancing support for new installations.

Upgrades are not supported from previous versions of EC.

According to the [k0s docs](https://docs.k0sproject.io/head/nllb/).

> The k0s worker process on worker nodes that are already running must be restarted for the new configuration to take effect.

Adds a test that resets the first controller node and checks that a pod will still start on a worker node.

```release-note
For new installations, the control plane is now set up as highly available within the cluster, enabling the removal of controller nodes from multi-node clusters without affecting pod scheduling.
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -113,6 +113,8 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v4
     - name: Make manifests
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: make -C operator manifests
     - name: Check CRDs
       run: |
--- a/e2e/scripts/check-nllb.sh
+++ b/e2e/scripts/check-nllb.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# This script is meant to be run after deleting the first control plane node to ensure the cluster
+# is still functional.
+
 set -euxo pipefail
 
 DIR=/usr/local/bin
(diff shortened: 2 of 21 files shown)


.github/workflows/ci.yaml
```diff
@@ -113,6 +113,8 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v4
     - name: Make manifests
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
laverya, line 117: why does this need a token?

e2e/scripts/check-nllb.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+DIR=/usr/local/bin
+. $DIR/common.sh
+
+function main() {
```
laverya, line 8: I would add a comment to this saying that it's meant to be run after deleting the first controller node to ensure the cluster is still functional

laverya (dismissed):

.github/workflows/ci.yaml
emosbaugh: It uses auth to install kustomize

e2e/scripts/check-nllb.sh
emosbaugh: done

laverya (approved):

PR #2017, opened 2025-03-24
warn about joining controllers in parallel

ajp-io:

Warn that you shouldn't join another node when a controller is being joined.

Also update other warns to use `logrus.Warn`.

Screenshot 2025-03-24 at 2 44 41 PM

https://app.shortcut.com/replicated/story/106755/warn-that-you-should-only-join-one-node-at-a-time

NONE

```release-note
When joining a controller node, warn that the user shouldn't join another node until the controller joins successfully.
```

https://github.com/replicatedhq/replicated-docs/pull/3126


emosbaugh (approved):

emosbaugh:
Failures are unrelated to changes.

PR #2020, opened 2025-03-24
add jobs to scan released images

St0rmz1:

- Adds new image scanning process
- Scan the actual public image on dockerhub
- Use Grype as the scanner
- Use Sarif format as attachment to the repo so it gets added to the code scan results
- Enrich the Sarif with properties we use for reporting purposes
- To determine the image tag we use the release version from this repo
- We scan the public image to ensure we are scanning what people are using. If we were to build the image and scan, we would actually be scanning a new image instead of the current release.

None

None

```release-note
NONE
```

No

### DIFF
--- a/.github/workflows/daily-container-scans.yml
+++ b/.github/workflows/daily-container-scans.yml
@@ -1,0 +1,78 @@
+# Container Security Scans
+# This workflow orchestrates security scanning of container images using Anchore scanner.
+# It runs nightly and can be triggered manually to scan various container images for vulnerabilities.
+name: Container Security Scans
+
+# Trigger configuration
+on:
+  schedule:
+    - cron: '0 0 * * *'  # Runs nightly at midnight (UTC)
+  workflow_dispatch:      # Allows manual triggering through GitHub UI
+
+# Security hardening: Start with no permissions and grant only what's needed
+permissions: {}  # Remove all permissions by default
+
+# Prevent multiple workflow runs from interfering with each other
+# This ensures only one scan runs at a time and new triggers cancel old runs
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  # Get the latest release tag first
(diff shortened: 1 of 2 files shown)


.github/workflows/daily-container-scans.yml
```diff
@@ -0,0 +1,78 @@
...
+      security-events: write # Needed to upload SARIF results
+    with:
+      # Use the latest release tag from the previous job
+      image: replicated/ec-registry:${{ needs.get-latest-tag.outputs.tag_name }}
+      # Report findings of medium severity or higher
+      severity-cutoff: medium
+      # Continue even if vulnerabilities are found
+      fail-build: false
+      # Specify platform to scan
+      platform: linux/amd64
```
laverya, line 78: Do you want to scan the registry image, or the LAM image? (Comment says one, code says the other)

St0rmz1: thanks for catching that incorrect comment. Will edit and push the update

.github/workflows/daily-container-scans.yml
```diff
@@ -0,0 +1,78 @@
...
```
laverya, line 78:
```suggestion
  # Scan local artifact mirror image using latest release tag
  scan-registry:
    name: Scan Local Artifact Mirror Image
    needs: get-latest-tag  # Wait for tag to be fetched
    uses: ./.github/workflows/scan-container-image.yml
    # Grant required permissions to the reusable workflow
    permissions:
      contents: read        # Needed to read workflow files
      security-events: write # Needed to upload SARIF results
    with:
      # Use the latest release tag from the previous job
      image: replicated/embedded-cluster-local-artifact-mirror:${{ needs.get-latest-tag.outputs.tag_name }}
      # Report findings of medium severity or higher
      severity-cutoff: medium
      # Continue even if vulnerabilities are found
      fail-build: false
      # Specify platform to scan
      platform: linux/amd64 
```

PR #2027, opened 2025-03-25
fix(ha): make more resilient

emosbaugh:

- Adds an enable-ha command
- Runs the registry data migration in a pod rather than an exec
- Retries uploading individual files on failure during registry data migration

```release-note

```


pkg/addons/embeddedclusteroperator/upgrade.go
```diff
@@ -15,7 +15,7 @@ func (e *EmbeddedClusterOperator) Upgrade(ctx context.Context, kcli client.Clien
 		return errors.Wrap(err, "check if release exists")
 	}
 	if !exists {
-		slog.Info("Release not found, installing", "release", releaseName, "namespace", namespace)
+		logrus.Debugf("Release not found, installing release %s in namespace %s", releaseName, namespace)
```
sgalsaleh, line 18: why change to debug? debug logs aren't easily accessibly in the pods like the cli no?

pkg/addons/highavailability.go
```diff
@@ -15,85 +17,76 @@ import (
...
-			return errors.Wrap(err, "upgrade registry")
+			loading.Infof("Migrating data for high availability")
+			logrus.Debugf("Migrating data for high availability")
+			err = migrateRegistryData(ctx, kcli, kclient, cfgspec, loading)
+			if err != nil {
+				return errors.Wrap(err, "migrate registry data")
+			}
+			logrus.Debugf("Data migration complete!")
+
+			loading.Infof("Enabling registry high availability")
```
sgalsaleh, line 79: i don't think we want to log this to the user. i think the previous log message is sufficient for this step as well

pkg/addons/highavailability.go
```diff
@@ -117,6 +110,74 @@ func EnableHA(ctx context.Context, kcli client.Client, hcli helm.Client, isAirga
...
+		return errors.Wrap(err, "get latest installation")
+	}
+
+	operatorImage, err := getOperatorImage()
+	if err != nil {
+		return errors.Wrap(err, "get operator image")
+	}
+	domains := runtimeconfig.GetDomains(cfgspec)
+	if domains.ProxyRegistryDomain != "" {
+		operatorImage = strings.Replace(operatorImage, "proxy.replicated.com", domains.ProxyRegistryDomain, 1)
```
sgalsaleh, line 126: why not make this logic part of the `getOperatorImage` function?

pkg/addons/registry/migrate/pod.go
```diff
@@ -0,0 +1,371 @@
...
+)
+
+const (
+	StatusConditionType = "RegistryMigrationStatus"
+
+	dataMigrationPodName = "registry-data-migration"
+	serviceAccountName   = "registry-data-migration-serviceaccount"
+
+	// seaweedfsS3SecretName is the name of the secret containing the s3 credentials.
+	// This secret name is defined in the chart in the release metadata.
```
sgalsaleh, line 33: > This secret name is defined in the chart in the release metadata.
idk what this means, but i don't think it applies anymore?

pkg/addons/registry/migrate/pod.go
```diff
@@ -0,0 +1,371 @@
...
+	// TODO: should we check seaweedfs health?
+
+	logrus.Debug("Ensuring migration service account")
+	err = ensureMigrationServiceAccount(ctx, cli)
+	if err != nil {
+		return nil, nil, fmt.Errorf("ensure service account: %w", err)
+	}
+
+	logrus.Debug("Ensuring s3 access credentials secret")
+	err = ensureS3Secret(ctx, cli)
```
sgalsaleh, line 60: this gets created by the seaweedfs chart, there's no need to do it here. it was needed previously when we used k0s extensions, but not anymore.

sgalsaleh: actually, nevermind, this is in the registry namespace, not seaweedfs namespace. ignore

sgalsaleh: hmm, it's not used in this migration though, and the registry chart creates it on upgrade. any reason to create it here?

pkg/addons/registry/migrate/pod.go
```diff
@@ -0,0 +1,371 @@
...
+		return fmt.Errorf("create secret: %w", err)
+	}
+	return nil
+}
+
+func applyRegistryLabels(labels map[string]string, component string) map[string]string {
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+	// TODO: why would we want to back this up?
```
sgalsaleh, line 365: i don't think we should. i'd say get rid of these labels for the migration pod

pkg/addons/registry/registry.go
```diff
@@ -101,3 +105,27 @@ func (r *Registry) ChartLocation() string {
 	}
 	return strings.Replace(Metadata.Location, "proxy.replicated.com", r.ProxyRegistryDomain, 1)
 }
+
+// IsRegistryHA checks if the registry has been configured for HA by looking for the
+// REGISTRY_STORAGE_S3_ACCESSKEY environment variable in the docker-registry container.
```
sgalsaleh, line 110: why not simply check if the replicas are > 1?

sgalsaleh: nevermind, that happens after the upgrade, not the migration.

pkg/addons/seaweedfs/install.go
```diff
@@ -4,66 +4,68 @@ import (
...
-	if err := s.createPreRequisites(ctx, kcli); err != nil {
-		return errors.Wrap(err, "create prerequisites")
-	}
+	return s.Upgrade(ctx, kcli, hcli, overrides)
+}
 
-	values, err := s.GenerateHelmValues(ctx, kcli, overrides)
-	if err != nil {
-		return errors.Wrap(err, "generate helm values")
+func (s *SeaweedFS) Uninstall(ctx context.Context, kcli client.Client) error {
```
sgalsaleh, line 24: i assume helm uninstall isn't sufficient?

pkg/kubeutils/kubeutils.go
```diff
@@ -117,44 +117,70 @@ func (k *KubeUtils) WaitForJob(ctx context.Context, cli client.Client, ns, name
...
+			return lasterr
 		}
+		return fmt.Errorf("timed out waiting for job %s", name)
 	}
 	return nil
 }
 
 // WaitForPod waits for a pod to be completed.
-func (k *KubeUtils) WaitForPodComplete(ctx context.Context, cli client.Client, ns, name string, opts *WaitOptions) error {
+func (k *KubeUtils) WaitForPodComplete(ctx context.Context, cli client.Client, ns, name string, opts *WaitOptions) (*corev1.Pod, error) {
```
sgalsaleh, line 158: my first impression when reading this function name is that it would return an error if the pod fails, but looks like it's up to the caller to make that decision based on the returned pod status?

pkg/addons/registry/registry.go
sgalsaleh: is this env var added after upgrade or migration?

pkg/addons/seaweedfs/install.go
```diff
@@ -114,14 +116,14 @@ func ensureService(ctx context.Context, kcli client.Client, serviceCIDR string)
 		}
 	}
 
-	if err := kcli.Create(ctx, obj); err != nil && !k8serrors.IsAlreadyExists(err) {
+	if err := kcli.Create(ctx, obj); err != nil {
```
sgalsaleh, line 119: i think this will fail when seaweedfs is upgraded later and the service already exists.

pkg/addons/seaweedfs/upgrade.go
```diff
@@ -2,31 +2,43 @@ package seaweedfs
...
-	"log/slog"
 
 	"github.com/pkg/errors"
 	"github.com/replicatedhq/embedded-cluster/pkg/helm"
+	"github.com/sirupsen/logrus"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 func (s *SeaweedFS) Upgrade(ctx context.Context, kcli client.Client, hcli helm.Client, overrides []string) error {
+	if err := s.ensurePreRequisites(ctx, kcli); err != nil {
```
sgalsaleh, line 13: i don't think this is idempotent. some it's functionality will fail if a resource already exists.

pkg/addons/embeddedclusteroperator/upgrade.go
emosbaugh: It is this way in all other addons
Screenshot 2025-03-26 at 8 05 22 AM

pkg/addons/highavailability.go
emosbaugh: It sits at that 100% for a long time. Are you sure there is no other message we want to log?

pkg/addons/registry/migrate/pod.go
emosbaugh: It is used here https://github.com/replicatedhq/embedded-cluster/pull/2027/files#diff-acb8fb3dc0e105f16b5eff96df7f6b3a3c3444c9bb9145965aa490608a06fed5R250

pkg/addons/registry/registry.go
emosbaugh: Good call. Thanks!

pkg/addons/seaweedfs/install.go
emosbaugh: Its deleted above here https://github.com/replicatedhq/embedded-cluster/pull/2027/files#diff-1ba7f2b58909b59c50d80f035bb11b2a496751bf84ed2ea2cc675cfa84d11880R113

pkg/kubeutils/kubeutils.go
emosbaugh: Completed != succeeded

pkg/addons/seaweedfs/install.go
```diff
@@ -4,66 +4,68 @@ import (
...
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 func (s *SeaweedFS) Install(ctx context.Context, kcli client.Client, hcli helm.Client, overrides []string, writer *spinner.MessageWriter) error {
-	if err := s.createPreRequisites(ctx, kcli); err != nil {
-		return errors.Wrap(err, "create prerequisites")
-	}
+	return s.Upgrade(ctx, kcli, hcli, overrides)
```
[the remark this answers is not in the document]
emosbaugh: I like this. I think we should move all addons to this "ensure" logic

pkg/addons/seaweedfs/upgrade.go
emosbaugh: I believe this pr makes it idempotent. Do you think otherwise?

pkg/addons/registry/migrate/pod.go
emosbaugh: updated to make it more clear

pkg/addons/registry/migrate/pod.go
emosbaugh: updated

pkg/addons/registry/migrate/migrate.go
```diff
@@ -0,0 +1,230 @@
...
+
+const (
+	s3Bucket        = "registry"
+	s3RootDirectory = "registry"
+	labelSelector   = "app=docker-registry"
+)
+
+// RegistryData runs a migration that copies data on disk in the registry-data PVC to the seaweedfs
+// s3 store. If it fails, it will scale the registry deployment back to 1. If it succeeds, it will
+// create a secret used to indicate success to the operator.
```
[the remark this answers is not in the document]
emosbaugh: fixed. also moved the scaling to the caller

pkg/addons/highavailability.go
sgalsaleh: i see. i would ask Alex what the copy should be then

emosbaugh: @ajp-io ?

pkg/addons/seaweedfs/install.go
sgalsaleh: only real benefit is if there are non-namespaced resources as helm would delete those. other note is just that we're using helm to install / upgrade. not strong objection though to keep it this way.

emosbaugh: What if i tried to helm uninstall and on failure delete the namespace?

cmd/installer/cli/enable_ha.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/replicatedhq/embedded-cluster/pkg/versions"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+// EnableHACmd is the command for enabling HA mode.
+func EnableHACmd(ctx context.Context, name string) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "enable-ha",
+		Short: fmt.Sprintf("Enable high availability for the %s cluster", name),
```
emosbaugh, line 23: @ajp-io can you review this?

cmd/installer/cli/root.go
```diff
@@ -100,6 +100,7 @@ func RootCmd(ctx context.Context, name string) *cobra.Command {
 	cmd.AddCommand(JoinCmd(ctx, name))
 	cmd.AddCommand(ShellCmd(ctx, name))
 	cmd.AddCommand(NodeCmd(ctx, name))
+	cmd.AddCommand(EnableHACmd(ctx, name))
```
emosbaugh, line 103: @ajp-io should this be nested under root?

ajp-io: i don't see a place that makes sense other than `install`, but I don't know that i like that either. i'm open to it though.

ajp-io: but then it inherits all the command line flags from `install`, and it doesn't need those

cmd/installer/cli/enable_ha.go
ajp-io: lgtm

pkg/addons/highavailability.go
```diff
@@ -2,99 +2,105 @@ package addons
...
+			}
+
+			logFn("Migrating data for high availability")
+			err = migrateRegistryData(ctx, kcli, kclient, cfgspec, logFn)
+			if err != nil {
+				return errors.Wrap(err, "migrate registry data")
+			}
+			logrus.Debugf("Data migration complete!")
+
+			logFn("Enabling registry high availability")
```
emosbaugh, line 94: @ajp-io can you review this message to the user? thanks

cmd/installer/cli/install.go
```diff
@@ -321,6 +319,15 @@ func runInstall(ctx context.Context, name string, flags InstallCmdFlags, metrics
 
 	// TODO (@salah): update installation status to reflect what's happening
 
+	logrus.Debugf("adding insecure registry")
+	registryIP, err := registry.GetRegistryClusterIP(flags.cidrCfg.ServiceCIDR)
+	if err != nil {
+		return fmt.Errorf("unable to get registry cluster IP: %w", err)
+	}
+	if err := airgap.AddInsecureRegistry(fmt.Sprintf("%s:5000", registryIP)); err != nil {
```
sgalsaleh, line 327: curious why is this needed here since it's added as part of the registry chart installation

sgalsaleh (approved):

pkg/addons/highavailability.go
ajp-io:
```suggestion
			logFn("Enabling high availability for the registry")
```

cmd/installer/cli/install.go
emosbaugh: I removed it from the registry chart installation. It seemed wrong there since its touching the host containerd conf files.

sgalsaleh (approved):

PR #2036, opened 2025-03-27
ToLower() hostnames where it'll be used to address nodes

hedge-sparrow:

If a host has a mixed case hostname, k0s will normalise the case to lowercase for the kubernetes node name, but we still check for the node by it's mixed-case name in several places. 

```release-note
Allow running clusters on nodes with hostnames that include uppercase letters.
```

### DIFF
--- a/cmd/installer/cli/install.go
+++ b/cmd/installer/cli/install.go
@@ -983,7 +983,8 @@ func waitForNode(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("get hostname: %w", err)
 	}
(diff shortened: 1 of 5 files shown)


laverya (approved):

hedge-sparrow:
the current failing TestMultiNodeAirgapUpgradePreviousStable test is failing in an expected manner. the new e2e test with mixed-case hostnames is throwing the install off because it can't find the node name

laverya (approved):

emosbaugh (approved):

PR #2043, opened 2025-03-28
refactor image scans

St0rmz1:

Enhancements to the daily image scanning process

file and job name changes for clarity
added input validations
the reusable workflow now creates a sarif even if no vulns detected. This is so external tools will always have a recent file to pull for reporting.

St0rmz1/chore/refactor image scans

NONE

NONE

```release-note
NONE
```

NONE


.github/workflows/daily-image-scans.yml
```diff
@@ -0,0 +1,67 @@
...
+    with:
+      image: replicated/embedded-cluster-operator-image:${{ needs.get-latest-tag.outputs.tag_name }}
+      severity-cutoff: negligible
+      fail-build: false
+      output-file: results.sarif
+      timeout-minutes: 30
+      retention-days: 90
+      category-prefix: container-scan-
+
+  scan-registry:
```
emosbaugh, line 52:
```suggestion
  scan-local-artifact-mirror:
```

emosbaugh (changes requested):
In the current scan we detect all images and scan. Is this something we intend to keep? I'm ok with this in a follow up

https://github.com/replicatedhq/embedded-cluster/blob/main/.github/workflows/image-scan.yaml#L69

PR #2066, opened 2025-04-03
chore: previous k0s version v1.29.14+k0s.0

emosbaugh:

```release-note

```


Makefile
```diff
@@ -12,8 +12,8 @@ ADMIN_CONSOLE_MIGRATIONS_IMAGE_OVERRIDE =
 ADMIN_CONSOLE_KURL_PROXY_IMAGE_OVERRIDE =
 K0S_VERSION = v1.30.9+k0s.0
 K0S_GO_VERSION = v1.30.9+k0s.0
-PREVIOUS_K0S_VERSION ?= v1.29.9+k0s.0-ec.0
-PREVIOUS_K0S_GO_VERSION ?= v1.29.9+k0s.0
+PREVIOUS_K0S_VERSION ?= v1.29.14+k0s.0-ec.0
```
emosbaugh, line 15: we need to make a custom build here with patch

laverya: I don't think we do actually
v1.29.14+k0s.0 works fine in the 1.29 branch

HOWEVER
if you update to .14 here then the upgrade tests in the 1.29 branch won't work

In total I don't think this is a change that is useful - all it's changing is the version we upgrade from in e2e testing

PR #2086, opened 2025-04-08
[bug] don't set profiles flag if no profiles

hedge-sparrow:

if no worker profiles are set, the profile flag still gets applied but is empty. this causes some problems

```release-note

```


sgalsaleh (dismissed):

laverya (dismissed):

laverya:
image

I'm forcibly merging this

PR #2094, opened 2025-04-08
fix: handle ctrl+c in prompts

emosbaugh:

> Ordinarily, when you type Ctrl-C, the terminal recognizes this as the QUIT button and delivers a SIGINT signal to the process, which terminates it. However, Survey temporarily configures the terminal to deliver control codes as ordinary input bytes. When Survey reads a ^C byte (ASCII \x03, "end of text"), it interrupts the current survey and returns a github.com/AlecAivazis/survey/v2/terminal.InterruptErr from Ask or AskOne. If you want to stop the process, handle the returned error in your code:

https://github.com/AlecAivazis/survey?tab=readme-ov-file#why-isnt-ctrl-c-working

```release-note

```

### DIFF
--- a/cmd/installer/cli/reset.go
+++ b/cmd/installer/cli/reset.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/AlecAivazis/survey/v2/terminal"
 	autopilot "github.com/k0sproject/k0s/pkg/apis/autopilot/v1beta2"
 	"github.com/k0sproject/k0s/pkg/etcd"
 	"github.com/replicatedhq/embedded-cluster/pkg/config"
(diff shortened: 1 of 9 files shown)


emosbaugh:
Test failures are unrelated to the change.

cmd/installer/cli/reset.go
```diff
@@ -234,7 +241,14 @@ func checkErrPrompt(noPrompt bool, force bool, err error) bool {
 		return false
 	}
 	logrus.Info("Continuing may leave the cluster in an unexpected state.")
-	return prompts.New().Confirm("Do you want to continue anyway?", false)
+	confirmed, err := prompts.New().Confirm("Do you want to continue anyway?", false)
+	if errors.Is(err, terminal.InterruptErr) {
```
sgalsaleh, line 245: why do we care about this type of error here if we're not reporting it?

sgalsaleh (dismissed):

cmd/installer/cli/reset.go
emosbaugh: removed

sgalsaleh (approved):

PR #2116, opened 2025-04-15
feat: generate controller join commands from the CLI

laverya:

https://www.loom.com/share/0284975afe7b495b872a9737411b575b

Adds `embedded-cluster join print-command` that prints a controller node join command.

```release-note

```


e2e/scripts/check-installation-state.sh
```diff
@@ -51,6 +51,7 @@ main() {
     # if this is the current version in CI
     if echo "$version" | grep -qvE "(pre-minio-removal|1.8.0-k8s|previous-stable)" ; then
         validate_data_dirs
+        check_join_command
```
sgalsaleh, line 54: feels like this should be a dryrun test instead similar to this? https://github.com/replicatedhq/embedded-cluster/blob/c74d25cc007c258532e0af3fe55e7fd60a1284d6/tests/dryrun/join_test.go#L206

sgalsaleh: actually, nevermind. can keep it here to test the feature e2e.

sgalsaleh (approved):

e2e/scripts/check-installation-state.sh
laverya: Could be both

sgalsaleh (approved):

PR #2122, opened 2025-04-22
feat(lam): support for pull binaries command in online mode

emosbaugh:

In preparation for a better join experience, run the LAM jobs in online mode to update the binaries and artifacts on the host similar to how we do it for airgap.

```release-note

```

### DIFF
--- a/cmd/local-artifact-mirror/pull_binaries.go
+++ b/cmd/local-artifact-mirror/pull_binaries.go
@@ -137,7 +137,8 @@ func PullBinariesCmd(cli *CLI) *cobra.Command {
 }
 
 // fetchBinaryWithLicense downloads the binary from the Replicated app using basic auth with license ID
-func fetchBinaryWithLicense(url, licenseID, binaryName string) (string, error) { // Create a temporary directory to store the binary
+func fetchBinaryWithLicense(url, licenseID, binaryName string) (string, error) {
+	// Create a temporary directory to store the binary
 	tmpdir, err := os.MkdirTemp("", "lam-artifact-*")
 	if err != nil {
 		return "", fmt.Errorf("create temp dir: %w", err)
--- a/cmd/local-artifact-mirror/pull_binaries.go
+++ b/cmd/local-artifact-mirror/pull_binaries.go
@@ -137,58 +137,37 @@ func PullBinariesCmd(cli *CLI) *cobra.Command {
 }
 
 // fetchBinaryWithLicense downloads the binary from the Replicated app using basic auth with license ID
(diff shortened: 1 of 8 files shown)


cmd/local-artifact-mirror/pull_binaries.go
```diff
@@ -73,5 +128,116 @@ func PullBinariesCmd(cli *CLI) *cobra.Command {
...
+	// Make the request
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("HTTP request failed: %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		// Read response body for error details
+		body, readErr := io.ReadAll(resp.Body)
+		resp.Body.Close()
```
sgalsaleh, line 198: why not defer close?

cmd/local-artifact-mirror/pull_binaries.go
```diff
@@ -29,22 +56,54 @@ func PullBinariesCmd(cli *CLI) *cobra.Command {
 				return fmt.Errorf("unable to create kube client: %w", err)
 			}
 
-			in, err := fetchAndValidateInstallation(ctx, kcli, args[0])
+			isAirgap := licenseID == ""
```
sgalsaleh, line 59: why isn't this just driven by `installation.Spec.AirGap`? using the license id makes it so we have to ensure both variables are taken into account when we use this command.

cmd/local-artifact-mirror/pull_binaries.go
emosbaugh: I am not sure why I would defer close. ive already read it all

sgalsaleh: i meant outside of this if block, there's nothing that closes the body.

emosbaugh: the function returns a read closer

cmd/local-artifact-mirror/pull_binaries.go
sgalsaleh: but you don't need to. the installation is the source of truth. it should simply be updated to not fail if it's not airgap.

emosbaugh: Updated https://github.com/replicatedhq/embedded-cluster/pull/2122/commits/81bc5d5fa42d879dcb6df77e14da71a3feca78f6

sgalsaleh (approved):

sgalsaleh (approved):

PR #2124, opened 2025-04-22
feat(multinode): online distribute artifacts on upgrade

emosbaugh:

Run the LAM jobs in online mode to update the binaries and artifacts on the host similar to how we do it for airgap.

This is needed so that kots can serve the updated binary for joining nodes.

```release-note

```


cmd/local-artifact-mirror/pull_binaries.go
```diff
@@ -71,7 +63,7 @@ func PullBinariesCmd(cli *CLI) *cobra.Command {
 				u := releaseURL(in.Spec.MetricsBaseURL, appSlug, channelSlug, appVersion)
 				logrus.Infof("fetching embedded cluster binary from %s using license ID", u)
 
-				location, err = fetchBinaryWithLicense(u, licenseID, in.Spec.BinaryName)
+				location, err = fetchBinaryWithLicense(u, licenseID, appSlug)
```
emosbaugh, line 66: This seems wrong but it doesnt work in the dev env...

operator/pkg/cli/upgrade.go
```diff
@@ -74,6 +75,10 @@ func UpgradeCmd() *cobra.Command {
 
 	// TODO(upgrade): local-artifact-mirror-image should be included in the installation object
 	cmd.Flags().StringVar(&localArtifactMirrorImage, "local-artifact-mirror-image", "", "Local artifact mirror image")
```
emosbaugh, line 77: If we were using the cli binary rather than the operator im not sure any of these flags would be required and this would be more straightforward

operator/pkg/cli/upgrade.go
```diff
@@ -74,6 +75,10 @@ func UpgradeCmd() *cobra.Command {
...
+	cmd.Flags().StringVar(&licenseID, "license-id", "", "License ID for online upgrades")
+	cmd.Flags().StringVar(&appSlug, "app-slug", "", "App slug for online upgrades")
+	cmd.Flags().StringVar(&channelSlug, "channel-slug", "", "Channel slug for online upgrades")
+	cmd.Flags().StringVar(&appVersion, "app-version", "", "App version for online upgrades")
```
emosbaugh, line 81: If we were using the cli binary rather than the operator im not sure any of these flags would be required and this would be more straightforward

pkg/addons/embeddedclusteroperator/install.go
```diff
@@ -29,3 +38,46 @@ func (e *EmbeddedClusterOperator) Install(ctx context.Context, kcli client.Clien
...
+func installEnsureCAConfigmap(ctx context.Context, kcli client.Client, privateCAs []string) error {
+	cas, err := privateCAsToMap(privateCAs)
+	if err != nil {
+		return errors.Wrap(err, "create private cas map")
+	}
+	return ensureCAConfigmap(ctx, kcli, cas)
+}
+
+func ensureCAConfigmap(ctx context.Context, cli client.Client, cas map[string]string) error {
+	kotsCAConfigmap := corev1.ConfigMap{
```
sgalsaleh, line 51:
```suggestion
	ecoCAConfigmap := corev1.ConfigMap{
```

pkg/addons/embeddedclusteroperator/install.go
```diff
@@ -29,3 +38,46 @@ func (e *EmbeddedClusterOperator) Install(ctx context.Context, kcli client.Clien
...
+			Name:      "private-cas",
+			Namespace: namespace,
+			Labels:    getBackupLabels(),
+		},
+		Data: cas,
+	}
+
+	err := cli.Create(ctx, &kotsCAConfigmap)
+	if client.IgnoreAlreadyExists(err) != nil {
+		return errors.Wrap(err, "create kotsadm-private-cas configmap")
```
sgalsaleh, line 66:
```suggestion
		return errors.Wrap(err, "create private-cas configmap")
```

pkg/addons/embeddedclusteroperator/install.go
emosbaugh: updated

pkg/addons/embeddedclusteroperator/install.go
emosbaugh: updated

sgalsaleh (approved):

PR #2125, opened 2025-04-24
Install extension on upgrade if doesn't exist

sgalsaleh:

Installs extension on upgrade if Helm release doesn't exist in the cluster.

[SC-122672](https://app.shortcut.com/replicated/story/122672)

Yes

```release-note
* Improves Helm extension upgrade behavior to automatically install extensions if they don't exist during an upgrade operation.
```

NONE


dev/dockerfiles/local-artifact-mirror/Dockerfile.ttlsh
```diff
@@ -22,6 +22,3 @@ FROM debian:bookworm-slim
 
 COPY --from=build /app/local-artifact-mirror/bin/local-artifact-mirror /usr/bin/local-artifact-mirror
 RUN ln -s /usr/bin/local-artifact-mirror /usr/local/bin/local-artifact-mirror
-
-RUN groupadd -r lam && useradd -r -u 1000 -g lam lam
-USER 1000
```
sgalsaleh, line 27: this caused airgap upgrades in dev env to fail because cli commands need to run as root in the artifacts job

emosbaugh (approved):

PR #2128, opened 2025-04-25
fix: join should use same logic as install to determine network interface if unspecified

emosbaugh:

This PR enhances the join command by using the same logic as the install command to determine the network interface when not explicitly specified.

```release-note
Fixes an issue where node joins can fail if the --network-interface flag is not specified, caused by the installer automatically selecting a different interface than the one used by the install command on the first node."
```


divolgin (dismissed):

cmd/installer/cli/join.go
```diff
@@ -618,3 +627,23 @@ func maybeEnableHA(ctx context.Context, kcli client.Client, flags JoinCmdFlags,
 		loading,
 	)
 }
+
+func joinDetermineBestNetworkInterface(joinAddress string) (string, error) {
+	// Try to use the interface that contains the ip address from the join command
```
sgalsaleh, line 632: this assumes that the interface containing the first node's ip address is what's going to be used for traffic going to the first node, which might not always be the case no? specially in environments like GCP where routing is managed outside the VM.

emosbaugh: That is why it fails silently and falls back to the same detection behavior of the install command

sgalsaleh (approved):

PR #2148, opened 2025-05-07
feat: release k0s 1.31

JGAntunes:

```release-note

```


pkg/config/images.go
```diff
@@ -33,15 +32,10 @@ func ListK0sImages(cfg *k0sv1beta1.ClusterConfig) []string {
...
+			cfg.Spec.Images.Konnectivity.URI(),
+			cfg.Spec.Images.PushGateway.URI():
 		default:
-			if strings.Contains(image, constant.KubePauseContainerImage) {
-				// there's a bug in GetImageURIs where it always returns the original pause image
-				// instead of the one in the config, make sure to use the one in the config.
-				images = append(images, cfg.Spec.Images.Pause.URI())
-			} else {
-				images = append(images, image)
-			}
```
JGAntunes, line 44: https://github.com/k0sproject/k0s/pull/5520 addressed this I believe.

pkg/config/images.go
```diff
@@ -33,15 +32,10 @@ func ListK0sImages(cfg *k0sv1beta1.ClusterConfig) []string {
 		// skip these images
 		case cfg.Spec.Images.KubeRouter.CNI.URI(),
 			cfg.Spec.Images.KubeRouter.CNIInstaller.URI(),
-			cfg.Spec.Images.Konnectivity.URI():
+			cfg.Spec.Images.Konnectivity.URI(),
+			cfg.Spec.Images.PushGateway.URI():
```
JGAntunes, line 36: With https://github.com/k0sproject/k0s/pull/5520 push gateway now shows up on the airgap image list. My understanding is that we don't need/use this since it's used for metrics scrapping - https://docs.k0sproject.io/head/system-monitoring/ - that we don't enable. Based on that we can probably skip it altogether.

pkg/config/images.go
JGAntunes: However we can't do this until we drop support for 1.30 and 1.29 😅

operator/pkg/upgrade/upgrade_test.go
```diff
@@ -157,7 +158,7 @@ config:
 			},
 			validate: func(t *testing.T, updatedConfig *k0sv1beta1.ClusterConfig) {
 				// Verify that the unsupported override was applied to the telemetry config
-				assert.Equal(t, false, updatedConfig.Spec.Telemetry.Enabled)
+				assert.Equal(t, false, *updatedConfig.Spec.Telemetry.Enabled)
```
sgalsaleh, line 161: would this compile in the 1.29 branch? i assume it wouldn't be a pointer and this wouldn't compile

JGAntunes: ah right and once we merge this in the release branch we're 💀 good call

JGAntunes: Fixed via - https://github.com/replicatedhq/embedded-cluster/pull/2148/commits/0c54c17835e7b39de8158842b34ecfe8c89aba7a

sgalsaleh (approved):

PR #2163, opened 2025-05-14
hide install run-preflights command

ajp-io:

hide `install run-preflights` because it's not fully productized or documented yet.

```release-note

```


jtuchscherer (approved):
LGTM. See previous comment

PR #2168, opened 2025-05-15
feat: detect host ca bundle

emosbaugh:

This is a precursor to deprecating the --private-ca flag and instead using the CA bundle from the host to inject into containers to support private CAs.

There is no functional change here and the product should continue to work as is.

```release-note

```

### DIFF
--- a/cmd/installer/cli/ca.go
+++ b/cmd/installer/cli/ca.go
@@ -1,0 +1,45 @@
+package cli
+
+import (
+	"errors"
+	"fmt"
+	"os"
+)
+
+// findHostCABundle locates the system CA certificate bundle on the host.
+// It first checks the SSL_CERT_FILE environment variable, then searches
+// common file paths used by various Linux distributions.
+//
+// The search follows the same order as the Go standard library's crypto/x509 package.
+//
+// Returns the path to the first found CA certificate bundle and nil error on success.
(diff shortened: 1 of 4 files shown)


cmd/installer/cli/ca.go
```diff
@@ -0,0 +1,45 @@
...
+		"/etc/pki/tls/certs/ca-bundle.crt",                  // Fedora/RHEL 6
+		"/etc/ssl/ca-bundle.pem",                            // OpenSUSE
+		"/etc/pki/tls/cacert.pem",                           // OpenELEC
+		"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem", // CentOS/RHEL 7
+		"/etc/ssl/cert.pem",                                 // Alpine Linux
+	}
+
+	// Check each file in the order of preference returning the first found
+	for _, file := range certFiles {
+		if _, err := os.Stat(file); err == nil {
```
sgalsaleh, line 39: should this only ignore NotExist errors?

sgalsaleh (approved):

cmd/installer/cli/ca.go
emosbaugh: I think it follows the go logic more closely as is

https://github.com/golang/go/blob/master/src/crypto/x509/root_unix.go#L47C25-L79

sgalsaleh (approved):

PR #2173, opened 2025-05-19
feat(operator): improved private ca support from host

emosbaugh:

Pass the ca bundle detected from the host to the operator chart.

Mounts the CA into the operator container at path /certs/ca-certificates.crt and sets the SSL_CERT_DIR env var so that golang will use the certificate.

Adds extraVolumes and extraVolumeMounts keys to the helm values file to pass through to the operator deployment similar to the [velero helm chart](https://github.com/vmware-tanzu/helm-charts/blob/main/charts/velero/templates/node-agent-daemonset.yaml#L86-L89).

Related https://github.com/replicatedhq/embedded-cluster/pull/2168

```release-note

```


diamonwiggins (approved):
lgtm

PR #2181, opened 2025-05-22
chore(e2e): create multiple VMs in one go through the CLI

JGAntunes:

Based on the following thread:
- https://replicated.slack.com/archives/C04T9AAMJUS/p1747707927071029

Seems like a lot of the E2E test flakiness and errors we've been getting with CMX can be addressed if instead of manually creating the network and each node in parallel we leverage the VM API and create all the required VMs in one go (which will be under the same network by default).

NA

NONE

```release-note
NONE
```

NONE

### DIFF
--- a/e2e/cluster/cmx/cluster.go
+++ b/e2e/cluster/cmx/cluster.go
@@ -125,7 +125,7 @@ func NewNodes(in *ClusterInput) ([]Node, error) {
 			return nil, fmt.Errorf("copy scripts to node %s: %v", nodes[i].ID, err)
 		}
 
-		if in.Nodes == 1 {
+		if i == 0 {
 			in.T.Logf("exposing port 30003 on node %s", nodes[i].ID)
 			hostname, err := exposePort(nodes[i], "30003")
 			if err != nil {
--- a/e2e/cluster/cmx/cluster.go
+++ b/e2e/cluster/cmx/cluster.go
@@ -299,7 +299,7 @@ func (c *Cluster) Destroy() {
 func (c *Cluster) removeNode(node Node) {
 	output, err := exec.Command("replicated", "vm", "rm", node.ID).CombinedOutput()
 	if err != nil {
-		c.t.Logf("failed to destroy node %s: %v: %s", node.Name, err, string(output))
+		c.t.Logf("failed to destroy node %s: %v: %s", node.ID, err, string(output))
 	}
 }
 
--- a/e2e/cluster/cmx/cluster.go
+++ b/e2e/cluster/cmx/cluster.go
@@ -12,8 +12,6 @@ import (
 	"sync"
 	"testing"
 	"time"
-
-	"github.com/google/uuid"
 )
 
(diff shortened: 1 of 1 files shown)


e2e/cluster/cmx/cluster.go
```diff
@@ -36,87 +33,40 @@ type Cluster struct {
...
-	nodeName := fmt.Sprintf("node%d", index)
-	in.T.Logf("creating node %s", nodeName)
+func NewNodes(in *ClusterInput) ([]Node, error) {
+	in.T.Logf("creating %s nodes", strconv.Itoa(in.Nodes))
 
 	args := []string{
 		"vm", "create",
-		"--name", nodeName,
-		"--network", networkID,
+		"--name", "ec-test-suite",
```
JGAntunes, line 68: Since we're creating multiple nodes now they'll all gonna have the same name, I've replaced the ID in most of the logs in here and also added a couple of initial log lines post cluster creation that maps the index in the Nodes slice to the node ID and private IP. This is how it looks:
```
   cluster.go:64: creating 3 nodes
    cluster.go:136: node 0 created with ID b8309fa6 and private IP 10.0.0.247
    cluster.go:136: node 1 created with ID 1a08d35f and private IP 10.0.0.192
    cluster.go:136: node 2 created with ID 5054f8b6 and private IP 10.0.0.229
    cluster.go:56: cluster created with network ID 624bf1704b41342375aa887fcce206124bfbd01da194720556d06976ec9f435f
    restore_test.go:649: 2025-05-22T12:46:48+01:00: deploying minio on node 0
    restore_test.go:655: 2025-05-22T12:49:16+01:00: downloading airgap files
    airgap.go:92: downloaded airgap bundle on node 0 to /assets/ec-release.tgz (2.0 GB) in 1m17.617605375s
    airgap.go:92: downloaded airgap bundle on node 0 to /assets/ec-release-upgrade.tgz (2.1 GB) in 1m21.8185815s
    restore_test.go:668: 2025-05-22T12:50:37+01:00: installing expect package on node 0
    restore_test.go:673: 2025-05-22T12:50:42+01:00: installing expect package on node 2
    restore_test.go:678: 2025-05-22T12:50:46+01:00: airgapping cluster
    cluster.go:236: node 0 is airgapped successfully
    cluster.go:236: node 1 is airgapped successfully
    cluster.go:236: node 2 is airgapped successfully
```

sgalsaleh (approved):

divolgin:
Just heads up, this does not help with the "host unreachable" error.  We should have a fix for that soon though.

sgalsaleh (dismissed):

emosbaugh (approved):
one question but lgtm

e2e/cluster/cmx/cluster.go
JGAntunes: I would imagine it's fine since we do all the operations based on node IDs, it's also not that different from having multiple runs using `node0´ for example.

PR #2184, opened 2025-05-22
Create CMX nodes sequentially

sgalsaleh:

This is to avoid a race condition in CMX where the network bridge isn't added to all nodes when creating nodes in parallel.

Yes

```release-note
NONE
```

NONE


emosbaugh (approved):

sgalsaleh:
Merging as the failing 3 tests are flaky and not related to this change.

PR #2187, opened 2025-05-23
Installer Experience v2 - Milestone 1

sgalsaleh:
#### Does this PR require a release note?

```release-note

```

#### Does this PR require documentation?

NONE

PR #2191, opened 2025-05-27
fix(api-client): addresses bugs in API client

JGAntunes:
This PR branches out of:
- https://github.com/replicatedhq/embedded-cluster/pull/2187

Adds tests to the API client and fixes the underlying bugs found as part of it.

```release-note

```

### DIFF
--- a/api/client/auth.go
+++ b/api/client/auth.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 )
 
+// Login sends a login request to the API server with the provided password and retrieves a session token. The token is stored in the client struct for subsequent requests.
 func (c *client) Login(password string) error {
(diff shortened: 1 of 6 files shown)


api/client/auth.go
```diff
@@ -35,13 +36,13 @@ func (c *client) Login(password string) error {
...
-		SessionToken string `json:"sessionToken"`
+		Token string `json:"token"`
 	}
 	err = json.NewDecoder(resp.Body).Decode(&loginResp)
 	if err != nil {
 		return err
 	}
 
-	c.token = loginResp.SessionToken
+	c.token = loginResp.Token
```
JGAntunes, line 46: Matching the API response body.

PR #2192, opened 2025-05-27
feat(web): initial work to setup root handler

JGAntunes:

The idea is to setup a root handler that we can template our initial HTML we serve with any needed state we need before rendering the app.

```release-note

```


JGAntunes:
https://github.com/replicatedhq/embedded-cluster/pull/2201 should address the CI failures. It's probably better than creating a placeholder index.html file in the `dist` folder.

cmd/installer/cli/install_test.go
```diff
@@ -581,7 +581,7 @@ func Test_runInstallAPI(t *testing.T) {
 		},
 	}
 	resp, err := httpClient.Get(url)
-	assert.NoError(t, err)
+	require.NoError(t, err)
```
JGAntunes, line 584: Later it would panic leading to a red herring because there's no `resp`.

cmd/installer/cli/install.go
```diff
@@ -97,6 +98,9 @@ type InstallCmdFlags struct {
 	tlsKeyBytes  []byte
 }
 
+// webAssetsFS is the filesystem to be used by the web component. Defaults to nil allowing the web server to use the default assets embedded in the binary. Useful for testing.
+var webAssetsFS fs.FS = nil
```
JGAntunes, line 102: Not particularly happy about this, unsure if we should keep this somewhere else, open to ideas.

cmd/installer/cli/install_test.go
```diff
@@ -560,6 +561,29 @@ func Test_runInstallAPI(t *testing.T) {
 	certPool := x509.NewCertPool()
 	certPool.AddCert(cert.Leaf)
 
+	// We need a release object to pass over to the Web component.
+	dataMap := map[string][]byte{
+		"kots-app.yaml": []byte(`
```
sgalsaleh, line 566: should probably include an icon and a title

sgalsaleh (approved):

cmd/installer/cli/install_test.go
JGAntunes: we can, albeit it's not really being used right now. Unless we also want to run some requests within this test against the web components?

emosbaugh (approved):

sgalsaleh:
Merging as the changes aren't related to the flaky e2e tests.

PR #2199, opened 2025-05-27
chore(api): generated docs - login

emosbaugh:

Allow for logging in to get the bearer token

```release-note

```


api/auth.go
```diff
@@ -48,8 +40,19 @@ func (a *API) authMiddleware(next http.Handler) http.Handler {
 	})
 }
 
+// postAuthLogin handler to authenticate a user
+//
+//	@Summary		Authenticate a user
+//	@Description	Authenticate a user
```
sgalsaleh, line 46: indentation looks wrong, don't think it matters though.

emosbaugh: The fmt command in the makefile is opinionated. This is correct

emosbaugh:
Test failures are unrelated. Merging

PR #2201, opened 2025-05-28
chore: include generated assets in repo

emosbaugh:

Adds generated assets to repo and checks that they are up to date or fails the CI run.

```release-note

```

### DIFF
--- a/web/vite.config.ts
+++ b/web/vite.config.ts
@@ -7,7 +7,9 @@ import path from 'path';
 export default defineConfig({
   plugins: [
     react(),
-    // Preserve README.md from .gitignore
+    // When building, vite removes README.md from the dist directory which makes the git tree dirty as it is in the .gitignore.
+    // This is needed because otherwise the go build fails with "pattern dist: no matching files found".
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -112,21 +112,74 @@ jobs:
...
+        run: npm run build
+      - name: Check web is up to date
+        run: |
+          git diff --exit-code --name-only
+          if [ $? -eq 0 ]; then
+            echo "Web is up to date"
+          else
+            echo "Web is out of date"
+            exit 1
+          fi
```
JGAntunes, line 182: Alternatively, why don't we build the web assets on CI VS having to locally build it and commit it? Things such as `npm install` are non-deterministic, meaning my local web build can be different from yours. In CI we can make sure we have reproducible builds via specific node+npm versions and commands such as `npm ci` for example.

JGAntunes (commented):
Overall LGTM, have one suggestion and one question

JGAntunes (approved):

PR #2203, opened 2025-05-28
fix: empty tmp dir instead of removing it

banjoh:
The tmp directory is used for a number of things including when editig k8s resources using kubectl edit. Removing it after running commands will lead to failures in certain commands.

[sc-124106](https://app.shortcut.com/replicated/story//collecting-a-support-bundle-deletes-tmp-directory-causing-failures-when-editing-k8s-resources-in-cli)

Unit test added

```release-note
Empty <data-dir>/tmp dir instead of deleting in when cleaning up after running commands
```

NONE

### DIFF
--- a/pkg/runtimeconfig/runtimeconfig.go
+++ b/pkg/runtimeconfig/runtimeconfig.go
@@ -27,7 +27,23 @@ func Get() *ecv1beta1.RuntimeConfigSpec {
 }
 
 func Cleanup() {
-	os.RemoveAll(EmbeddedClusterTmpSubDir())
+	emptyTmpDir()
+}
+
+func emptyTmpDir() {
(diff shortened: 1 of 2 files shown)


pkg/runtimeconfig/runtimeconfig.go
```diff
@@ -27,7 +27,23 @@ func Get() *ecv1beta1.RuntimeConfigSpec {
 }
 
 func Cleanup() {
-	os.RemoveAll(EmbeddedClusterTmpSubDir())
+	emptyTmpDir()
```
sgalsaleh, line 30: we already have `helpers.RemoveAll`. can you please use that instead? it also has tests

sgalsaleh (approved):

sgalsaleh:
Merging as change is unrelated and tests are flaky right now.

PR #2239, opened 2025-06-03
Set SSL Cert ConfigMap in KOTS and Improve Detection of Unhealthy Pods in Tests

diamonwiggins:

Set `SSL_CERT_CONFIGMAP` so KOTS knows what configmap to use for the Replicated SDK. See here - https://github.com/replicatedhq/kots-helm/blob/main/templates/kotsadm-deployment.yaml#L108-L113. Without this, the SDK  doesn't work in proxied environments.

Also in this PR is an adjustment to our function which validates if there are unhealthy pods. Right now it doesn't account for pods which are stuck in `0/1 Running` like the SDK is currently in proxied environments

```release-note

```

### DIFF
--- a/pkg/addons/adminconsole/static/values.tpl.yaml
+++ b/pkg/addons/adminconsole/static/values.tpl.yaml
@@ -19,3 +19,6 @@ passwordSecretRef:
     name: kotsadm-password
 service:
     enabled: false
+privateCAs:
+    enabled: true
+    configmapName: "kotsadm-private-cas"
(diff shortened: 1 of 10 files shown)


sgalsaleh (approved):

pkg/addons/adminconsole/static/values.tpl.yaml
```diff
@@ -19,3 +19,6 @@ passwordSecretRef:
     name: kotsadm-password
 service:
     enabled: false
+extraEnv:
```
sgalsaleh, line 22: is this needed anymore if we're adding it on generate helm values? doesn't look like we include the others so we should be consistent and maybe not include it here

sgalsaleh (approved):

PR #2243, opened 2025-06-04
Increase preflight checks timeout on upgrade

sgalsaleh:

Increases preflight checks timeout on upgrade to 30 seconds

```release-note
NONE
```

NONE


jtuchscherer (approved):
LGTM

PR #2249, opened 2025-06-05
Revert "Use default CMX instance type to avoid quota limit"

sgalsaleh:
Reverts replicatedhq/embedded-cluster#2242


JGAntunes (approved):

sgalsaleh:
Merging as this only reverts back to large CMX instances and doesn't touch test logic.

PR #2250, opened 2025-06-05
Increase preflight initial load timeout on upgrade

sgalsaleh:

Fixes some flakiness on upgrade tests due to short timeouts.

```release-note
NONE
```

NONE


sgalsaleh:
Merging as this only increases the timeout of a flaky test.

PR #2251, opened 2025-06-05
Wait for SSH to be available on CMX node

sgalsaleh:

Waits for SSH to be available on CMX node before running ssh commands.

```release-note
NONE
```


squizzi (approved):

sgalsaleh:
Merging as the failing test is flaky.

PR #2252, opened 2025-06-05
Log stdout and stderr if ssh isn't available

sgalsaleh:

Log stdout and stderr if ssh isn't available for easier debuggability.

```release-note
NONE
```

NONE


screspod (approved):

sgalsaleh:
Merging as this doesn't affect actual test logic.

PR #2258, opened 2025-06-07
Improvements to manager preflights UI

sgalsaleh:

Improvements to manager preflights UI

```release-note
NONE
```

NONE

#### Loom
https://www.loom.com/share/71e59fcfa7af44709787b0c14344093e?sid=b7689c2c-4a0e-40b3-af63-ce3e7e5eefb6


web/src/components/wizard/preflight/LinuxPreflightCheck.tsx
```diff
@@ -54,7 +72,8 @@ const LinuxPreflightCheck: React.FC<LinuxPreflightCheckProps> = ({ onComplete })
         },
       });
       if (!response.ok) {
-        throw new Error("Failed to run preflight checks");
+        const errorData = await response.json().catch(() => ({}));
+        throw new Error(errorData.message || "Failed to run preflight checks");
```
JGAntunes, line 76: If I'm reading correctly, we're ignoring the `json()` error here and has a fallback returning an Error `"Failed to run preflight checks"`.

- I wonder what happens when `json()` fails, I believe errorData can be `undefined` which will result in `... errorData.message ...` to fail.
- Wondering if we want to capture the response body's `text` for analysis since this shouldn't really happen. Maybe we should bubble up that error?

web/src/components/wizard/preflight/LinuxPreflightCheck.tsx
```diff
@@ -79,14 +100,15 @@ const LinuxPreflightCheck: React.FC<LinuxPreflightCheckProps> = ({ onComplete })
         },
       });
       if (!response.ok) {
-        throw new Error("Failed to get installation config status");
+        const errorData = await response.json().catch(() => ({}));
```
JGAntunes, line 103: See ☝️

web/src/components/wizard/preflight/LinuxPreflightCheck.tsx
```diff
@@ -99,7 +121,8 @@ const LinuxPreflightCheck: React.FC<LinuxPreflightCheckProps> = ({ onComplete })
         },
       });
       if (!response.ok) {
-        throw new Error("Failed to get preflight status");
+        const errorData = await response.json().catch(() => ({}));
+        throw new Error(errorData.message || "Failed to get preflight status");
```
JGAntunes, line 125: See ☝️

web/src/components/wizard/preflight/LinuxPreflightCheck.tsx
```diff
@@ -187,87 +166,135 @@ const LinuxPreflightCheck: React.FC<LinuxPreflightCheckProps> = ({ onComplete })
...
+        >
+          <CheckCircle className="w-6 h-6" style={{ color: themeColor }} />
+        </div>
+        <p className="text-lg font-medium text-gray-900">Host validation successful!</p>
+      </div>
+    );
+  }
+
+  // If there are no failures and no warnings then we have an api error
+  if (!hasFailures(preflightResponse?.output) && !hasWarnings(preflightResponse?.output)) {
```
JGAntunes, line 195: Can we use the API response as a way to infer the error state VS using the absence of failures/warnings?

web/src/contexts/ConfigContext.tsx
```diff
@@ -1,39 +1,24 @@
...
-  };
 }
 
 interface PrototypeSettings {
   skipValidation: boolean;
   failPreflights: boolean;
   failInstallation: boolean;
   failHostPreflights: boolean;
-  clusterMode: "existing" | "embedded";
+  clusterMode: 'existing' | 'embedded';
```
JGAntunes, line 21: Seems like we're struggling with formatting/ style options. Thoughts on setting up - https://www.npmjs.com/package/prettier ?

web/src/components/wizard/preflight/LinuxPreflightCheck.tsx
sgalsaleh: ditto

web/src/components/wizard/preflight/LinuxPreflightCheck.tsx
sgalsaleh: ditto

web/src/contexts/ConfigContext.tsx
sgalsaleh: yes, we have a bad history with prettier 😅: https://github.com/replicatedhq/kots/pull/5260

web/src/components/wizard/preflight/LinuxPreflightCheck.tsx
JGAntunes: Ah I didn't know we returned a non 2xx code when the preflights failed.
Yeah I think we wanna keep those 2 separate. Shall we have a follow up for it?

web/src/contexts/ConfigContext.tsx
JGAntunes: Well, honestly I don't really care about what we use, as long as we have something setup in the project we can all rely on without the need to bikeshed too much around it.

sgalsaleh: yeah, we should look into it

web/src/components/wizard/preflight/LinuxPreflightCheck.tsx
JGAntunes: Ah thanks for clarifying that, I mixing some concepts here 😅 in that case I still think we should be able to distinguish between those 2 failure modes. Not on this PR though. Shall we have a follow up story to add that?

JGAntunes (approved):
We should create stories for the 2 action items but otherwise lgtm

PR #2226, opened 2025-06-02
chore(test): add tests to memory store

JGAntunes:
#### What this PR does / why we need it:

Follow up to https://github.com/replicatedhq/embedded-cluster/pull/2187

See:
- https://github.com/replicatedhq/embedded-cluster/pull/2187#discussion_r2107367897

#### Which issue(s) this PR fixes:

https://app.shortcut.com/replicated/story/124364/add-tests-to-memory-store

#### Does this PR require a release note?

```release-note

```

PR #2263, opened 2025-06-09
fix(install): keep a single instance of runtime config within the API context

JGAntunes:

This PR addresses something that came up once I rebased - https://github.com/replicatedhq/embedded-cluster/pull/2226
See the failure:
- https://github.com/replicatedhq/embedded-cluster/actions/runs/15537684440/job/43740622893?pr=2226#step:4:1401

This failure is a consequence of how we're currently passing around the runtime config between controller, managers, etc. This doesn't surface in production because we let the install controller instantiate everything, however I'm worried we might run into this in the future with the way we have our structs wired up. The idea here is that instead of keeping a runtime config property in multiple places that can lead to a disconnect between which instance is being used, we keep one copy of it at the controller level and pass it around to all the manager methods when they need it. We could eventually look into either keeping the runtime config in a specific manager (Installation?) or creating its dedicated manager (overkill?) but I believe that for now this should suffice.

Updated

```release-note
NONE
```

NONE

### DIFF
--- a/api/controllers/install/controller.go
+++ b/api/controllers/install/controller.go
@@ -125,7 +125,6 @@ func NewInstallController(opts ...InstallControllerOption) (*InstallController, 
 
 	if controller.installationManager == nil {
 		controller.installationManager = installation.NewInstallationManager(
-			installation.WithRuntimeConfig(controller.rc),
 			installation.WithLogger(controller.logger),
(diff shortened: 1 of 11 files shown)


JGAntunes:
Opted for parking this for now until we get a bit more clarity in terms of how we want to use `runtimeconfig` and how that ties down with `InstallationConfig`

PR #2284, opened 2025-06-09
feat(ui): new manager-experience sub-command, use app-slug for data dir, respect flags as defaults

emosbaugh:

- use the app slug as the default dir only for the manager-experience install flow
- changes `--manager-experience` flag to sub-command to `embedded-cluster install manager-experience`
- respects flags passed in by end user as defaults in the manager-experience config form

```release-note

```


cmd/installer/cli/install.go
```diff
@@ -290,6 +350,47 @@ func preRunInstall(cmd *cobra.Command, flags *InstallCmdFlags, rc runtimeconfig.
...
+	}
+	flags.cidrCfg = cidrCfg
+
+	// if a network interface flag was not provided, attempt to discover it
+	if flags.networkInterface == "" {
+		autoInterface, err := newconfig.DetermineBestNetworkInterface()
+		if err == nil {
+			flags.networkInterface = autoInterface
+		}
+	}
```
emosbaugh, line 381: This is probably all redundant in the api right now where defaults are set

cmd/installer/cli/install.go
```diff
@@ -376,74 +477,57 @@ func preRunInstall(cmd *cobra.Command, flags *InstallCmdFlags, rc runtimeconfig.
...
-		flags.proxy = proxy
-
-		if err := verifyProxyConfig(flags.proxy, prompts.New(), flags.assumeYes); err != nil {
-			return err
-		}
-		logrus.Debug("User confirmed prompt to proceed installing with `http_proxy` set and `https_proxy` unset")
-
-		if err := validateCIDRFlags(cmd); err != nil {
-			return err
+		if err := updateRuntimeConfigFromInstallCmdFlags(flags, rc); err != nil {
```
emosbaugh, line 485: rather than a configChan, we should just have a runtimeConfigChan and set it directly since its all computed in the api now.

api/controllers/install/installation.go
```diff
@@ -34,35 +33,12 @@ func (c *InstallController) ConfigureInstallation(ctx context.Context, config *t
...
-
 	if err := c.installationManager.SetConfig(*config); err != nil {
 		return fmt.Errorf("write: %w", err)
 	}
 
-	// TODO (@team): discuss the distinction between the runtime config and the installation config
-	// update the runtime config
-	c.rc.SetDataDir(config.DataDirectory)
-	c.rc.SetLocalArtifactMirrorPort(config.LocalArtifactMirrorPort)
-	c.rc.SetAdminConsolePort(config.AdminConsolePort)
```
emosbaugh, line 49: this was moved to the manager

api/controllers/install/controller_test.go
```diff
@@ -190,65 +171,6 @@ func TestConfigureInstallation(t *testing.T) {
 	}
 }
 
-// TestIntegrationComputeCIDRs tests the CIDR computation with real networking utility
-func TestIntegrationComputeCIDRs(t *testing.T) {
```
emosbaugh, line 194: this comment is not correct. it is redundant and tested already in the utils package

pkg/config/config.go
```diff
@@ -30,7 +30,7 @@ var k0sConfigPathOverride string
 func RenderK0sConfig(proxyRegistryDomain string) *k0sconfig.ClusterConfig {
 	cfg := k0sconfig.DefaultClusterConfig()
 	// Customize the default k0s configuration to our taste.
-	cfg.Name = runtimeconfig.BinaryName()
+	cfg.Name = "k0s"
```
emosbaugh, line 33: i checked in the cluster and it seems to be named this regardless

api/internal/managers/installation/manager.go
```diff
@@ -104,12 +97,8 @@ func NewInstallationManager(opts ...InstallationManagerOption) *installationMana
 		manager.logger = logger.NewDiscardLogger()
 	}
 
-	if manager.installation == nil {
-		manager.installation = types.NewInstallation()
-	}
-
 	if manager.installationStore == nil {
-		manager.installationStore = NewMemoryStore(manager.installation)
+		manager.installationStore = NewMemoryStore(manager.rc, types.NewStatus())
```
sgalsaleh, line 101: The status needs to be tied to the one in the "install" type i think, can't be a new one, otherwise it wouldn't update correctly there. That's why the installation object was passed down as a reference

cmd/installer/cli/install.go
sgalsaleh: I got rid of the config chan in my branch that installs everything in the api, so there won't be a need for it

operator/controllers/installation_controller.go
```diff
@@ -78,7 +78,16 @@ var copyHostPreflightResultsJob = &batchv1.Job{
...
+					},
+					{
+						Name: "k0s",
+						VolumeSource: corev1.VolumeSource{
+							HostPath: &corev1.HostPathVolumeSource{
+								Path: runtimeconfig.K0sBinaryPath,
+								Type: ptr.To(corev1.HostPathFile),
 							},
 						},
 					},
```
emosbaugh, line 93: i have no idea why im all of a sudden seeing this error

```
$ kubectl  -n embedded-cluster logs copy-host-preflight-results-node3-4d8ks
/embedded-cluster/bin/kubectl: /embedded-cluster/bin/kubectl: exec: exec: line 2: line 2: /usr/local/bin/k0s: not found
```

api/internal/managers/installation/manager.go
sgalsaleh: before this change, the `manager.installation` would be a pointer to the one on the `install` object here: https://github.com/replicatedhq/embedded-cluster/blob/7f687075fa235d9ec4a916f4f74b84bce1ec0698/api/controllers/install/controller.go#L130

so the changes reflect in the `install` object. with this change, it's creating a new status and changes wouldn't reflect in the `install` object.

emosbaugh: Perhaps we need to refactor out these pointers? It is hard to understand and has side effects.

sgalsaleh: Yeah agreed

PR #2327, opened 2025-06-16
feat(ui): hide validation from install wizard navigation

diamonwiggins:

This PR consolidates the ValidationStep functionality into the SetupStep component, simplifying the installation wizard flow from 5 steps to 4 steps. Instead of having validation as a separate wizard step, validation now occurs as an internal state within the setup step.

[SC-124824](https://app.shortcut.com/replicated/story/124824/put-host-preflights-in-the-setup-page-instead-of-its-own-validation-page)

yes, added. 

```release-note
NONE
```

NONE


web/src/components/wizard/InstallWizard.tsx
```diff
@@ -14,29 +13,19 @@ const InstallWizard: React.FC = () => {
...
   };
 
-  const goToPreviousStep = () => {
-    const steps: WizardStep[] = ["welcome", "setup", "validation", "installation", "completion"];
-    const currentIndex = steps.indexOf(currentStep);
-    if (currentIndex > 0) {
-      setCurrentStep(steps[currentIndex - 1]);
-    }
-  };
-
```
diamonwiggins, line 31: I removed this function onBack functionality from the SetupStep because it looks like you can only reach the setup page if you're already authenticated, and the WelcomeStep automatically redirects authenticated users back to SetupStep, so the back button effectively didn't do anything.

web/src/components/wizard/SetupStep.tsx
```diff
@@ -142,38 +164,69 @@ const SetupStep: React.FC<SetupStepProps> = ({ onNext }) => {
...
       </Card>
 
-      <div className="flex justify-end">
-        <Button onClick={handleNext} icon={<ChevronRight className="w-5 h-5" />}>
-          Next: Validate Host
-        </Button>
+      <div className="flex justify-between">
+        {currentView === 'validation' && (
+          <Button variant="outline" onClick={handleBackToConfiguration} icon={<ChevronLeft className="w-5 h-5" />}>
+            Back to Configuration
```
ajp-io, line 211: I prefer just "Back" on this button. Configuration is going to be confusing anyway once we add the actual config screen into the installer.

web/src/components/wizard/StepNavigation.tsx
```diff
@@ -13,20 +13,26 @@ const StepNavigation: React.FC<StepNavigationProps> = ({ currentStep }) => {
...
   ];
 
+  // All wizard steps for progress calculation
+  const allSteps: WizardStep[] = ['welcome', 'setup', 'validation', 'installation', 'completion'];
+
   const getStepStatus = (step: { id: string }) => {
-    const stepIndex = steps.findIndex((s) => s.id === step.id);
-    const currentIndex = steps.findIndex((s) => s.id === currentStep);
+    const allStepIndex = allSteps.indexOf(step.id as WizardStep);
+    const currentAllStepIndex = allSteps.indexOf(currentStep);
```
sgalsaleh, line 29: i think we can keep the old names, they're easier to grok
```suggestion
    const stepIndex = allSteps.indexOf(step.id as WizardStep);
    const currentStepIndex = allSteps.indexOf(currentStep);
```

sgalsaleh (approved):

sgalsaleh:
Merging as failing test is flaky and not related to PR changes.

PR #2333, opened 2025-06-17
chore: Make `isRunning` atomic at the Manager level

screspod:

Renames the isRunning function to isState, makes it atomic, and relocates it from the Store level to the Manager level.

https://app.shortcut.com/replicated/story/124680/make-isrunning-atomic-at-the-manager-level

Existing tests were updated.

```release-note
NONE
```

NONE

### DIFF
--- a/api/internal/managers/installation/config.go
+++ b/api/internal/managers/installation/config.go
@@ -207,7 +207,7 @@ func (m *installationManager) ConfigureHost(ctx context.Context, rc runtimeconfi
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
-	running, err := m.isRunning()
+	running, err := m.IsInState(types.StateRunning)
 	if err != nil {
 		return fmt.Errorf("check if installation is running: %w", err)
 	}
--- a/api/internal/managers/installation/manager.go
+++ b/api/internal/managers/installation/manager.go
@@ -21,6 +21,7 @@ type InstallationManager interface {
 	SetConfig(config types.InstallationConfig) error
 	GetStatus() (types.Status, error)
 	SetStatus(status types.Status) error
+	IsInState(state types.State) (bool, error)
 	ValidateConfig(config types.InstallationConfig, managerPort int) error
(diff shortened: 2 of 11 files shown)


api/internal/managers/installation/manager.go
```diff
@@ -23,6 +23,7 @@ type InstallationManager interface {
 	ValidateConfig(config *types.InstallationConfig, managerPort int) error
 	SetConfigDefaults(config *types.InstallationConfig) error
 	ConfigureHost(ctx context.Context, rc runtimeconfig.RuntimeConfig) error
+	isState(state types.State) (bool, error)
```
emosbaugh, line 26: I've gotta say I'm not sure I've seen this pattern of putting a private method on an interface. Does that work when mocking from a different package? I will have to test.

screspod: I believe types outside the package can use the exported methods, but not the unexported or implement the interface. I would expect that you can't mock from a different package with this pattern. Should I export it?

api/internal/managers/installation/config.go
```diff
@@ -204,17 +204,19 @@ func (m *installationManager) setCIDRDefaults(config *types.InstallationConfig)
...
 		return fmt.Errorf("check if installation is running: %w", err)
 	}
 	if running {
 		return fmt.Errorf("installation configuration is already running")
 	}
 
+	// Now acquire the lock for the actual configuration
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
```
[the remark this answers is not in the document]
screspod: Discussed over standup and agreed to move the methods to the store level

emosbaugh (changes requested):
Lets not merge this in favor of https://github.com/replicatedhq/embedded-cluster/pull/2334

screspod:
Closing in favor of https://app.shortcut.com/replicated/story/124677/have-a-method-in-the-status-type-to-dictate-which-states-are-final

PR #2338, opened 2025-06-18
feat: add preflight to check for XFS filesystem with ftype=0

banjoh:
Some systems use an XFS filesystem that is configured with `ftype=0`, which is a requirement for `OverlayFS`. This causes problems for container runtimes like containerd, and can lead to failures during or after installation.

The failure and success (never shown) messages will be

```
FAIL: The XFS filesystem on /var/lib/embedded-cluster path is configured with ftype=0 which containers
      cannot run on. Please reconfigure the filesystem or use a different data directory.

PASS: File system on /var/lib/embedded-cluster path is fit for use with containers
```

[sc-124752](https://app.shortcut.com/replicated/story/124752/add-preflight-check-for-xfs-with-ftype-0-to-prevent-container-runtime-issues)

NONE

```release-note
Add preflight to check for XFS filesystem with ftype=0
```

NONE

### DIFF
--- a/pkg-new/preflights/host-preflight.yaml
+++ b/pkg-new/preflights/host-preflight.yaml
@@ -1215,7 +1215,7 @@ spec:
         outcomes:
           - fail:
               when: "true"
-              message: "The XFS filesystem on {{ .DataDir }} path is configured with ftype=0 which containers cannot run on. Please reconfigure the filesystem or use a different data directory."
+              message: "The XFS filesystem at {{ .DataDir }} is configured with ftype=0, which is not supported. Reformat the filesystem with ftype=1, or choose a different data directory on a supported filesystem."
           - pass:
               when: "false"
               message: "File system on {{ .DataDir }} path is fit for use with containers"
--- a/pkg-new/preflights/host-preflight.yaml
+++ b/pkg-new/preflights/host-preflight.yaml
@@ -172,6 +172,20 @@ spec:
             dir="{{ .DataDir }}"
             while [ "$dir" != "/" ]; do find "$dir" -maxdepth 0 ! -perm -111; dir=$(dirname "$dir"); done
             find "/" -maxdepth 0 ! -perm -111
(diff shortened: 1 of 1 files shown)


pkg-new/preflights/host-preflight.yaml
```diff
@@ -1194,3 +1208,14 @@ spec:
...
+        checkName: Check filesystem on data directory path
+        fileName: host-collectors/run-host/xfs_info-data-dir.txt
+        regex: 'ftype=0'
+        outcomes:
+          - fail:
+              when: "true"
+              message: "The XFS filesystem on {{ .DataDir }} path is configured with ftype=0 which containers cannot run on. Please reconfigure the filesystem or use a different data directory."
+          - pass:
+              when: "false"
+              message: "File system on {{ .DataDir }} path is fit for use with containers"
```
ajp-io, line 1221:
```suggestion
              message: "The XFS filesystem at {{ .DataDir }} is configured with ftype=1."
```

ajp-io (commented):
let me know if these make sense and sound good

pkg-new/preflights/host-preflight.yaml
banjoh: Perhaps a nit here. A pass will also include all other file systems

diamonwiggins: @banjoh are you saying that you would expand this same preflight in the future for additional hypothetical filesystem checks? If not, and we would add new checks in the future if we had to check something else for ext4 or zfs, then we'd probably have specific pass messages for those, so i think i'm in favor of what Alex is suggesting here. `"File system on {{ .DataDir }} path is fit for use with containers"` just seems like a bold claim to make when this check is very specific.

diamonwiggins (dismissed):
lgtm

diamonwiggins (approved):
lgtm

PR #2337, opened 2025-06-18
feat(reporting): add metrics reporting to the new manager experience

JGAntunes:
#### What this PR does / why we need it:
This ended up tackling a bit more than intended as we started digging through the state transitions and we spotted that:
- The lack of some states complicated our overall flow and reporting
- Some transitions were not happening or just wrong (e.g. we were conflating the PreflightsFailed state).

The bulk of this PR is around adding new states and fixing the ones that we have as well the transition mechanisms. This mermaid diagram illustrates the final graph:
```mermaid
graph TD

    %% Initial & config states
    StateNew -->|Success| StateInstallationConfigured
    StateNew -->|Fail| StateInstallationConfigurationFailed

    StateInstallationConfigurationFailed -->|Retry| StateInstallationConfigured
    StateInstallationConfigurationFailed -->|Retry| StateInstallationConfigurationFailed

    StateInstallationConfigured -->|Success| StateHostConfigured
    StateInstallationConfigured -->|Fail| StateHostConfigurationFailed

    StateHostConfigurationFailed -->|Retry| StateInstallationConfigured
    StateHostConfigurationFailed -->|Retry| StateInstallationConfigurationFailed

    StateHostConfigured -->|Success| StatePreflightsRunning
    StateHostConfigured -->|Retry| StateInstallationConfigured
    StateHostConfigured -->|Retry| StateInstallationConfigurationFailed

    %% Preflights
    StatePreflightsRunning -->|Success| StatePreflightsSucceeded
    StatePreflightsRunning -->|SuccessWithFailures| StatePreflightsFailed
    StatePreflightsRunning -->|FailureToExecute| StatePreflightsExecutionFailed

    StatePreflightsExecutionFailed -->|Retry| StatePreflightsRunning
    StatePreflightsExecutionFailed -->|RetryConfig| StateInstallationConfigured
    StatePreflightsExecutionFailed -->|RetryConfig| StateInstallationConfigurationFailed

    StatePreflightsSucceeded -->|Success| StateInfrastructureInstalling
    StatePreflightsSucceeded -->|Retry| StatePreflightsRunning
    StatePreflightsSucceeded -->|RetryConfig| StateInstallationConfigured
    StatePreflightsSucceeded -->|RetryConfig| StateInstallationConfigurationFailed

    StatePreflightsFailed -->|Success| StatePreflightsFailedBypassed
    StatePreflightsFailed -->|Retry| StatePreflightsRunning
[...]

PR #2363, opened 2025-06-23
chore: add statemachine event handlers

JGAntunes:
This PR relates to https://github.com/replicatedhq/embedded-cluster/pull/2337. Where, in order to support our reporting strategy we introduce event handlers that are triggered when successful transitions happen on the state machine. This has the added benefits of:
- All the reporting logic can be aggregated under the same code area
- No potential for drift between reported state VS actual state
- Managers become entirely independent/unaware of the metrics reporting mechanism
- Potential to easily extend the events we report in future.
- Potential to extend with other reporting mechanisms in the future.

https://app.shortcut.com/replicated/story/125431/spike-evaluate-setting-up-reporting-handlers-as-part-of-state-machine-transitions-in-new-manager-api

Yes, added

```release-note
NONE
```

NONE

### DIFF
--- a/api/internal/statemachine/event_handler.go
+++ b/api/internal/statemachine/event_handler.go
@@ -7,7 +7,44 @@ import (
 	"time"
 )
 
-// eventHandler is a struct that holds the event handler function and its timeout.
+var (
+	_ EventHandler = &eventHandler{}
+)
+
+// EventHandler is an interface for handling state transition events in the state machine.
+type EventHandler interface {
+	// TriggerHandler triggers the event handler for a state transition.
+	TriggerHandler(ctx context.Context, fromState, toState State) error
+}
+
+// EventHandlerFunc is a function that handles state transition events. Used to report state changes.
+type EventHandlerFunc func(ctx context.Context, fromState, toState State)
+
+// EventHandlerOption is a configurable state machine option.
+type EventHandlerOption func(*eventHandler)
+
+// WithHandlerTimeout sets the timeout for the event handler to complete.
+func WithHandlerTimeout(timeout time.Duration) EventHandlerOption {
+	return func(eh *eventHandler) {
+		eh.timeout = timeout
+	}
+}
+
+// NewEventHandler creates a new event handler with the provided function and options.
+func NewEventHandler(handler EventHandlerFunc, options ...EventHandlerOption) EventHandler {
+	eh := &eventHandler{
+		handler: handler,
+		timeout: 5 * time.Second, // Default timeout
+	}
+
+	for _, option := range options {
+		option(eh)
+	}
+
+	return eh
+}
+
+// eventHandler is a struct that implements the EventHandler interface. It contains a handler function that is called when a state transition occurs, and it supports a timeout for the handler to complete.
 type eventHandler struct {
 	handler EventHandlerFunc
 	timeout time.Duration // Timeout for the handler to complete, default is 5 seconds
@@ -26,9 +63,9 @@ func (eh *eventHandler) TriggerHandler(ctx context.Context, fromState, toState S
 				err := fmt.Errorf("event handler panic from %s to %s: %v: %s\n", fromState, toState, r, debug.Stack())
 				done <- err
 			}
+			close(done)
 		}()
 		eh.handler(ctx, fromState, toState)
-		close(done)
 	}()
 
 	select {
--- a/api/internal/statemachine/event_handler.go
+++ b/api/internal/statemachine/event_handler.go
@@ -1,0 +1,41 @@
+package statemachine
+
+import (
+	"context"
+	"fmt"
+	"runtime/debug"
+	"time"
+)
+
+// eventHandler is a struct that holds the event handler function and its timeout.
+type eventHandler struct {
+	handler EventHandlerFunc
+	timeout time.Duration // Timeout for the handler to complete, default is 5 seconds
+}
+
+// TriggerHandler triggers the event handler for a state transition. The trigger is blocking and will wait for the handler to complete or timeout.
+func (eh *eventHandler) TriggerHandler(ctx context.Context, fromState, toState State) error {
+	ctx, cancel := context.WithTimeout(ctx, eh.timeout)
+	defer cancel()
+	done := make(chan error, 1)
+
+	go func() {
+		defer func() {
+			if r := recover(); r != nil {
+				// Capture panic but don't affect the transition
+				err := fmt.Errorf("event handler panic from %s to %s: %v: %s\n", fromState, toState, r, debug.Stack())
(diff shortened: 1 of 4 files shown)


api/internal/statemachine/statemachine.go
```diff
@@ -123,11 +162,55 @@ func (sm *stateMachine) Transition(lock Lock, nextState State) error {
...
+func (sm *stateMachine) triggerHandlers(from, next State) {
+	handlers, exists := sm.eventHandlers[next]
+	if !exists || len(handlers) == 0 {
+		return
+	}
+	handlersCopy := make([]EventHandler, len(handlers))
+	// Make a copy of the handlers to avoid potential changes to the slice during execution
+	copy(handlersCopy, handlers)
+
+	go func(from, to State, handlerList []EventHandler) {
```
sgalsaleh, line 196: i wonder if running in a goroutine should be an option that can be overridden / specified by the client

api/internal/statemachine/statemachine.go
```diff
@@ -123,11 +162,55 @@ func (sm *stateMachine) Transition(lock Lock, nextState State) error {
...
+		ctx, cancel := context.WithTimeout(context.Background(), sm.handlerTimeout)
```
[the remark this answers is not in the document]
JGAntunes: We can make it configurable when we register the handler I guess 🤔 the general idea was to have an initial working approach. If we feel like we should make this configurable from the get go we can do it.

api/internal/statemachine/statemachine.go
JGAntunes: We could make it so that the client can decide if they want an async or an sync handler for example. My question would be if we have an actual use for that right now or not?

sgalsaleh: well, the reporting events we send today are actually all sync, not async, no?

api/internal/statemachine/statemachine.go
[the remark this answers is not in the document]
emosbaugh: i suppose i felt like the context here is unnecessary and should be removed

api/internal/statemachine/statemachine.go
emosbaugh: i think i agree with salah here that it should be sync. async is hard because we may need to catch signals to ensure that all handlers execute before the process exits.

emosbaugh: i think i agree with salah here that it should be synchronous as it is one more thing to gracefully shutdown before the process exits

JGAntunes: happy to make them sync in that case, we just need to be mindful that wtv we hook up to these handlers in the future will be sync and will mean blocking the transition while it happens.

api/internal/statemachine/statemachine.go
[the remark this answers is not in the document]
JGAntunes: with us making this sync we definitely need it now so I might make it so that it's configurable when you register the handler instead.

api/internal/statemachine/statemachine.go
JGAntunes: Done via https://github.com/replicatedhq/embedded-cluster/pull/2363/commits/9581235c330930803d2ce6a96f287a810013739a

api/internal/statemachine/statemachine.go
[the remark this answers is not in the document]
JGAntunes: Done via https://github.com/replicatedhq/embedded-cluster/pull/2363/commits/9581235c330930803d2ce6a96f287a810013739a

api/internal/statemachine/event_handler.go
```diff
@@ -0,0 +1,41 @@
...
+	go func() {
+		defer func() {
+			if r := recover(); r != nil {
+				// Capture panic but don't affect the transition
+				err := fmt.Errorf("event handler panic from %s to %s: %v: %s\n", fromState, toState, r, debug.Stack())
+				done <- err
+			}
+		}()
+		eh.handler(ctx, fromState, toState)
+		close(done)
```
sgalsaleh, line 31: i think this should happen in the defer above so it also gets closed on panic

api/internal/statemachine/interface.go
```diff
@@ -0,0 +1,78 @@
...
+
+// WithHandlerTimeout sets the timeout for the event handler to complete.
+func WithHandlerTimeout(timeout time.Duration) EventHandlerOption {
+	return func(eh *eventHandler) {
+		eh.timeout = timeout
+	}
+}
+
+// NewEventHandler creates a new event handler with the provided function and options.
+func NewEventHandler(handler EventHandlerFunc, options ...EventHandlerOption) EventHandler {
```
sgalsaleh, line 67: should this and related types live in the event_handler.go file instead of interface.go?

api/internal/statemachine/statemachine.go
```diff
@@ -123,11 +117,41 @@ func (sm *stateMachine) Transition(lock Lock, nextState State) error {
 		return fmt.Errorf("invalid transition from %s to %s", sm.currentState, nextState)
 	}
 
+	fromState := sm.currentState
 	sm.currentState = nextState
 
+	// Trigger event handlers after successful transition
+	sm.triggerHandlers(fromState, nextState)
```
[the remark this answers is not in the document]
JGAntunes: This is the only way we can guarantee order when triggering handlers for multiple transitions right?

Given two transitions, T1 and T2, we can only be certain that the handlers from T1 will execute before the handlers of T2 if we rely on a mutex (or semaphore, or other similar mechanism. When I say "only" here I mean re-using what we have).

In practical terms for our use case, this is the only way we can ensure you wouldn't get an installation succeeded before a preflights bypassed event for example.

api/internal/statemachine/event_handler.go
JGAntunes: Done via - https://github.com/replicatedhq/embedded-cluster/pull/2363/commits/f7e59592be57196c85b6ba9392f574308f669a3b

api/internal/statemachine/interface.go
JGAntunes: Done via - https://github.com/replicatedhq/embedded-cluster/pull/2363/commits/f7e59592be57196c85b6ba9392f574308f669a3b

api/internal/statemachine/statemachine.go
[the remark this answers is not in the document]
JGAntunes: Done via - https://github.com/replicatedhq/embedded-cluster/pull/2363/commits/51a362b7da32b9e960de25e3aacfa896a4dbe069

sgalsaleh:
Failing test is flaky.

PR #2370, opened 2025-06-25
feat(ui): show modal when frontend loses connection to backend

diamonwiggins:

This PR implements backend disconnect detection and handling for the embedded cluster installation UI using a simplified component approach based on code review feedback.

- A ConnectionMonitor component that manages its own internal state
- Automatic health checks every 10 seconds via /api/health endpoint
- A modal UI that appears when the backend becomes unavailable using conditional rendering
- Automatic retry mechanism with manual retry option

Fixes: [SC-125010](https://app.shortcut.com/replicated/story/125010/detect-that-the-ui-isn-t-being-served-anymore-and-show-a-message)

Test included for the new componented

```release-note
NONE
```

NONE

### DIFF
--- a/web/src/components/common/ConnectionMonitor.tsx
+++ b/web/src/components/common/ConnectionMonitor.tsx
@@ -120,4 +120,4 @@ const ConnectionMonitor: React.FC = () => {
   );
 };
 
-export default ConnectionMonitor; 
+export default ConnectionMonitor;
--- a/web/src/components/common/ConnectionMonitor.tsx
+++ b/web/src/components/common/ConnectionMonitor.tsx
@@ -1,32 +1,15 @@
-import React, { createContext, useContext, useEffect, useState, ReactNode } from 'react';
-
-interface ConnectionContextType {
-  isConnected: boolean;
-  isChecking: boolean;
-}
-
-const ConnectionContext = createContext<ConnectionContextType | undefined>(undefined);
-
-interface ConnectionProviderProps {
-  children: ReactNode;
-}
+import React, { useEffect, useState } from 'react';
 
 // Connection modal component
-const ConnectionModal: React.FC<{ isVisible: boolean; onRetry: () => void; isRetrying: boolean }> = ({ isVisible, onRetry, isRetrying }) => {
+const ConnectionModal: React.FC<{ onRetry: () => void; isRetrying: boolean }> = ({ onRetry, isRetrying }) => {
   const [retryCount, setRetryCount] = useState(0);
 
   useEffect(() => {
-    if (isVisible) {
-      const interval = setInterval(() => {
-        setRetryCount(count => count + 1);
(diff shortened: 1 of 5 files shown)


web/src/contexts/ConnectionContext.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+  return (
+    <ConnectionContext.Provider value={value}>
+      {children}
+      <ConnectionModal 
+        isVisible={!isConnected} 
+        onRetry={checkConnection}
+        isRetrying={isChecking}
+      />
+    </ConnectionContext.Provider>
```
JGAntunes, line 141: I wonder if we actually need this to be a context or if this could just be a stateless component handling props only instead 🤔

My understanding of context is it allows us to share state between deeply nested components without the need to pass props around. This would make sense for this case if we ever wanted other components to be aware of `ConnectionContextType` and check `isConnected` or `isChecking`, which I don't see happening?

Thoughts?

web/src/contexts/ConnectionContext.tsx
```diff
@@ -0,0 +1,151 @@
...
+  const value: ConnectionContextType = {
+    isConnected,
+    isChecking,
+  };
+
+  return (
+    <ConnectionContext.Provider value={value}>
+      {children}
+      <ConnectionModal 
+        isVisible={!isConnected}
```
JGAntunes, line 137: Wondering if the components visibility should be decided at the parent level, where we decide if we render it or not instead? i.e.

```js
{!isConnected && <ConnectionModal ... />}
```

web/src/contexts/ConnectionContext.tsx
diamonwiggins: Makes sense. I think when I started the idea was going to be to maybe present this modal more broadly on other components for other types of API calls as well, but decided to just use the the single `api/health` endpoint and keep things pretty simple. So not using react context i think is the right choice for now.

JGAntunes (approved):

web/src/components/common/ConnectionMonitor.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const [isConnected, setIsConnected] = useState(true);
+  const [isChecking, setIsChecking] = useState(false);
+
+  const checkConnection = async () => {
+    setIsChecking(true);
+    
+    try {
+      // Try up to 3 times before marking as disconnected
+      let attempts = 0;
+      const maxAttempts = 3;
```
sgalsaleh, line 62: let's make this 5

sgalsaleh: actually, nevermind i see it happens every 10 seconds and 5 would be too long

sgalsaleh:
Merging as failing test is flaky.

PR #2371, opened 2025-06-25
feat(v3): customize install command help output based on targets

emosbaugh:

Screenshot 2025-06-24 at 3 49 38 PM

```release-note

```

### DIFF
--- a/cmd/installer/cli/install.go
+++ b/cmd/installer/cli/install.go
@@ -98,7 +98,7 @@ func InstallCmd(ctx context.Context, name string) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "install",
 		Short:   short,
-		Example: installCmdExample(),
+		Example: installCmdExample(name),
 		PostRun: func(cmd *cobra.Command, args []string) {
 			rc.Cleanup()
 			cancel() // Cancel context when command completes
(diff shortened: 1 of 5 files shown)


cmd/installer/cli/install.go
```diff
@@ -153,6 +161,85 @@ func InstallCmd(ctx context.Context, name string) *cobra.Command {
...
+      --target linux \
+      --data-dir /opt/embedded-cluster \
+      --cidr 10.244.0.0/16 \
+      --license ./license.yaml \
+      --yes
+
+  # Install into an existing Kubernetes cluster
+  replicated install \
+      --target kubernetes \
+      --k8s-admin-console-ns replicated-admin \
```
ajp-io, line 181: this flag isn't in the help menu.

we're not yet prefixing with k8s, so i think this could be `--admin-console-namespace` and be available for kubernetes only.

cmd/installer/cli/install.go
```diff
@@ -153,6 +161,85 @@ func InstallCmd(ctx context.Context, name string) *cobra.Command {
...
+	return `
+  # Install on a Linux host
+  replicated install \
+      --target linux \
+      --data-dir /opt/embedded-cluster \
+      --cidr 10.244.0.0/16 \
+      --license ./license.yaml \
+      --yes
+
+  # Install into an existing Kubernetes cluster
```
ajp-io, line 178:
```suggestion
  # Install in a Kubernetes cluster
```

cmd/installer/cli/install.go
```diff
@@ -153,6 +161,85 @@ func InstallCmd(ctx context.Context, name string) *cobra.Command {
...
+	if os.Getenv("ENABLE_V3") != "1" {
+		return ""
+	}
+
+	return `
+  # Install on a Linux host
+  replicated install \
+      --target linux \
+      --data-dir /opt/embedded-cluster \
+      --cidr 10.244.0.0/16 \
```
ajp-io, line 174: this is the default value, so either drop it or change the value

cmd/installer/cli/install.go
```diff
@@ -153,6 +161,85 @@ func InstallCmd(ctx context.Context, name string) *cobra.Command {
...
 }
 
+func installCmdExample() string {
+	if os.Getenv("ENABLE_V3") != "1" {
+		return ""
+	}
+
+	return `
+  # Install on a Linux host
+  replicated install \
```
ajp-io, line 171: maybe this is just a placeholder for now, but this should use the app slug rather than `replicated`. same with other example.

ajp-io (commented):
that looks really good!

sgalsaleh (dismissed):

cmd/installer/cli/install.go
emosbaugh: changed to --kubeconfig

cmd/installer/cli/install.go
emosbaugh: fixed

PR #2380, opened 2025-06-27
feat: use appslug for default data directory in install wizard

diamonwiggins:

Uses the app slug for the default data directory to provide users with a more white labeled experienced. If a user specifies the `--data-dir` CLI flag then that will be populated in the UI by default.

Screenshot 2025-06-26 at 9 24 40 PM

[SC-124300](https://app.shortcut.com/replicated/story/124300/use-the-app-slug-for-the-default-data-directory-name)

```release-note
NONE
```

NONE

### DIFF
--- a/api/controllers/linux/install/installation.go
+++ b/api/controllers/linux/install/installation.go
@@ -21,6 +21,13 @@ func (c *InstallController) GetInstallationConfig(ctx context.Context) (types.In
 		return types.InstallationConfig{}, fmt.Errorf("set defaults: %w", err)
 	}
 
+	// Use the RuntimeConfig's data directory instead of the hardcoded default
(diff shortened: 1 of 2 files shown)


api/controllers/linux/install/installation.go
```diff
@@ -21,6 +21,10 @@ func (c *InstallController) GetInstallationConfig(ctx context.Context) (types.In
 		return types.InstallationConfig{}, fmt.Errorf("set defaults: %w", err)
 	}
 
+	// Always use the RuntimeConfig's data directory
+	// This ensures the UI shows the same data directory that was set by the CLI
+	config.DataDirectory = c.rc.EmbeddedClusterHomeDirectory()
```
sgalsaleh, line 26: this should be in SetConfigDefaults

PR #2394, opened 2025-06-30
Resolve selinux installation issues

ashjones-replicated:

Enables Embedded Cluster to install in selinux environments by:
- Setting selinux `bin_t` file context on our bin directory.
- Restoring selinux contexts for data-dir after creation. 

This should resolve selinux installation issues for most common selinux setups. 
administrators with custom policies and rules will still have to take extra steps to allow EC to install. 

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ K0S_GO_VERSION = v1.31.11+k0s.0
 PREVIOUS_K0S_VERSION ?= v1.30.9+k0s.0
 PREVIOUS_K0S_GO_VERSION ?= v1.30.9+k0s.0
 K0S_BINARY_SOURCE_OVERRIDE =
-TROUBLESHOOT_VERSION = v0.120.3
+TROUBLESHOOT_VERSION = v0.121.2
 
 KOTS_VERSION = v$(shell awk '/^version/{print $$2}' pkg/addons/adminconsole/static/metadata.yaml | sed -E 's/([0-9]+\.[0-9]+\.[0-9]+)(-ec\.[0-9]+)?.*/\1\2/')
 # If KOTS_BINARY_URL_OVERRIDE is set to a ttl.sh artifact, there's NO need to update the KOTS_VERSION above as it will be dynamically generated
--- a/api/internal/managers/app/release/manager.go
+++ b/api/internal/managers/app/release/manager.go
@@ -2,6 +2,7 @@ package release
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/replicatedhq/embedded-cluster/api/pkg/logger"
 	"github.com/replicatedhq/embedded-cluster/api/pkg/template"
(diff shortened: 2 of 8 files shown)


pkg-new/hostutils/initialize.go
```diff
@@ -39,6 +40,26 @@ func (h *HostUtils) ConfigureHost(ctx context.Context, rc runtimeconfig.RuntimeC
...
+
+		// Set selinux fcontext for embedded-cluster binary directory to bin_t
+		out, err := exec.Command("semanage", "fcontext", "-a", "-s", "system_u", "-t", "bin_t", rc.EmbeddedClusterBinsSubDir()+"(/.*)?").CombinedOutput()
+		if err != nil {
+			h.logger.Debugf("unable to set contexts on binary directory: %v", err)
+			h.logger.Debugln(out)
+		}
+
+		// Relabel whole embedded-cluster data directory since it's created with unconfined_u
+		out, err = exec.Command("restorecon", "-RvF", rc.EmbeddedClusterHomeDirectory()).CombinedOutput()
```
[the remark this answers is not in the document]
hedge-sparrow: yes, I'd expect `restorecon` to be available if `semanage` is. they're both provided by the policycoreutils package. 

The check for `semanage` is only really to judge if selinux is present on the system.

hedge-sparrow:
I've rolled back my test changes after a discussion with @chris-sanders. 

I've performed manual testing on selinux enabled systems that prove that this works, but was having significant trouble porting the single node alma linux test over to the cmx testing framework so that we can test automatically with selinux enabled.

e2e/cluster/cmx/cluster.go
```diff
@@ -363,7 +371,7 @@ func (c *Cluster) SetupPlaywrightAndRunTest(testName string, args ...string) (st
 
 func (c *Cluster) SetupPlaywright(envs ...map[string]string) error {
 	c.t.Logf("%s: bypassing kurl-proxy", time.Now().Format(time.RFC3339))
-	_, stderr, err := c.RunCommandOnNode(0, []string{"bypass-kurl-proxy.sh"}, envs...)
+	_, stderr, err := c.RunCommandOnNode(0, []string{"/usr/local/bin/bypass-kurl-proxy.sh"}, envs...)
```
sgalsaleh, line 374: is adding `/usr/local/bin` here necessary if it's being added to `PATH` above?

pkg-new/hostutils/initialize.go
```diff
@@ -39,6 +40,31 @@ func (h *HostUtils) ConfigureHost(ctx context.Context, rc runtimeconfig.RuntimeC
 		}
 	}
 
+	h.logger.Debugln("checking for semanage binary in $PATH")
```
sgalsaleh, line 43: nit: how about moving these two selinux related changes to an selinux.go and add them as methods on the host utils interface so they can be mocked?

e2e/cluster/cmx/cluster.go
hedge-sparrow: I _believe_ the `$PATH` variable being stripped might actually be _because_ of selinux 😓

sgalsaleh (approved):

PR #2406, opened 2025-07-03
Installer Update CLI Command V3 Design

diamonwiggins:

```release-note

```

### DIFF
--- a/design/embedded-cluster-cli-driven-update.md
+++ b/design/embedded-cluster-cli-driven-update.md
@@ -4,7 +4,7 @@
 
 This document outlines the implementation of **CLI-driven updates** to support the new architecture where customers will download a new binary to get the new release of software, compared to before where customers could check for versions in the UI to download and apply new versions.
 
-The implementation will **modify the existing `./binary update` command** to perform CLI-driven binary updating when `ENABLE_V3=1` is set, instead of its current behavior of updating applications with airgap bundles.
+This implementation would **modify the existing `./binary update` command** to perform CLI-driven binary updates when `ENABLE_V3=1` is set.
 
 ## Architecture Overview
 
--- a/design/embedded-cluster-cli-driven-update.md
+++ b/design/embedded-cluster-cli-driven-update.md
@@ -1,12 +1,14 @@
-# Embedded Cluster Self-Update Implementation Plan
+# Embedded Cluster CLI-Driven Update Design
 
 ## Overview
 
-This document outlines the high-level changes required to implement a self-updating mechanism for embedded cluster binaries, similar to the Replicated CLI's `version upgrade` command. The implementation will **modify the existing `./binary update` command** to perform binary self-updating when `ENABLE_V3=1` is set, instead of its current behavior of updating applications with airgap bundles.
+This document outlines the implementation of **CLI-driven updates** to support the new architecture where customers will download a new binary to get the new release of software, compared to before where customers could check for versions in the UI to download and apply new versions.
+
+The implementation will **modify the existing `./binary update` command** to perform CLI-driven binary updating when `ENABLE_V3=1` is set, instead of its current behavior of updating applications with airgap bundles.
 
 ## Architecture Overview
 
-The self-update mechanism will leverage the existing embedded cluster download infrastructure while adding new components for version management, update detection, and safe binary replacement.
+The CLI-driven update mechanism will leverage the existing embedded cluster download infrastructure while adding new components for version management, update detection, and safe binary replacement.
 
 ```
 Embedded Cluster Binary → Version Check → Download New Binary → Atomic Replacement
@@ -21,7 +23,7 @@ Embedded Cluster Binary → Version Check → Download New Binary → Atomic Rep
 #### Modified Update Command Behavior
 - **Conditional behavior**: Check `ENABLE_V3` environment variable to determine update mode
 - **V2 mode**: Current behavior (application airgap bundle updates)
-- **V3 mode**: Binary self-updating behavior
+- **V3 mode**: CLI-driven binary updating behavior
 - **Command signature (V3)**: `./binary update --license <license-file> [--version=<version>] [--force]`
 - **License requirement**: The `--license` flag is **required** when `ENABLE_V3=1` for authentication with replicated.app
 - **Validation**: Verify the binary can be updated (not installed via package manager)
@@ -123,7 +125,7 @@ Following the KOTS pattern, the embedded cluster should handle these channel typ
 
 ### 3. Update Discovery Mechanism
 
-The system needs to discover available updates by calling the replicated.app API.
+The binary needs to discover available updates by calling the replicated.app API.
 
 **New API Endpoint:** `GET /embedded/:appSlug/:channelSlug/versions`
 - **Architecture**: Follow same proxy pattern as download endpoint (replicated-app → market-api)
@@ -170,9 +172,9 @@ The system needs to discover available updates by calling the replicated.app API
 - **Error handling**: Consistent error responses and HTTP status codes with download endpoint
 - **Performance**: Reasonable response times with appropriate caching headers
 
-### 4. Binary Download and Replacement (TODO - Self-Update Logic)
+### 4. Binary Download and Replacement (CLI-Driven Update Logic)
 
-**Current State:** No existing self-update logic in embedded cluster binary.
+**Current State:** No existing binary update logic in embedded cluster binary.
 
 #### Download Process
 Leverage existing embedded cluster download infrastructure:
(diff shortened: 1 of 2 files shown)


design/embedded-cluster-cli-driven-update.md
```diff
@@ -0,0 +1,215 @@
...
+→ "Cannot update to v1.7.0 from beta channel. Required release v1.6.5 from stable channel must be installed first."
+```
+
+**Key Point:** This leverages existing infrastructure rather than building new version management from scratch.
+
+### 3. Update Discovery Mechanism
+
+The binary needs to discover available updates by calling the replicated.app API.
+
+**New API Endpoint:** `GET /embedded/:appSlug/:channelSlug/versions`
```
emosbaugh, line 130: does this send the current version? otherwise this list could get long.

design/embedded-cluster-cli-driven-update.md
```diff
@@ -0,0 +1,215 @@
...
+- **Reuse existing infrastructure**: Same authentication, validation, and proxy patterns as download endpoint
+- **Database optimization**: Efficient queries to avoid N+1 problems with large version lists
+- **Error handling**: Consistent error responses and HTTP status codes with download endpoint
+- **Performance**: Reasonable response times with appropriate caching headers
+
+### 4. Binary Download and Replacement (CLI-Driven Update Logic)
+
+**Current State:** No existing binary update logic in embedded cluster binary.
+
+#### Download Process
```
emosbaugh, line 179: I think it is built on the fly so this may not be possible but worth putting in here

design/embedded-cluster-cli-driven-update.md
```diff
@@ -0,0 +1,215 @@
...
+Leverage existing embedded cluster download infrastructure:
+- **Reuse download endpoint**: Use current `/embedded/:appSlug/:channelSlug/:versionLabel` pattern
+- **Authentication**: Pass license ID from `--license` flag in Authorization header
+- **Streaming download**: Handle large binaries efficiently without excessive memory usage
+- **Progress indication**: Show download progress with percentage and transfer rates
```
emosbaugh, line 184: I think it is built on the fly so this may not be possible but worth keeping in here

design/embedded-cluster-cli-driven-update.md
```diff
@@ -0,0 +1,215 @@
...
+- **Resume capability**: Support interrupted download resumption for reliability
```
emosbaugh, line 185: same since it is built on the fly

design/embedded-cluster-cli-driven-update.md
```diff
@@ -0,0 +1,215 @@
...
+- **Backup creation**: Always create backup of current binary before replacement
+- **Atomic operation**: Use filesystem moves/renames for atomicity where possible
+- **Permission preservation**: Maintain original executable permissions and ownership
+- **Cross-platform support**: Handle platform-specific file replacement quirks
+- **Verification post-replacement**: Confirm new binary works correctly
+
+#### Error Handling and Rollback
+Comprehensive error recovery mechanisms:
+- **Download failures**: Network timeouts, authentication errors, disk space issues
+- **Verification failures**: Checksum mismatches, corrupted downloads
```
emosbaugh, line 199: same since its built on the fly

design/embedded-cluster-cli-driven-update.md
```diff
@@ -0,0 +1,278 @@
...
+- **V2 mode**: Current behavior (application airgap bundle updates)
+- **V3 mode**: Binary self-updating behavior
+- **Command signature (V3)**: `./binary update --license <license-file> [--version=<version>] [--force]`
+- **License requirement**: The `--license` flag is **required** when `ENABLE_V3=1` for authentication with replicated.app
+- **Validation**: Verify the binary can be updated (not installed via package manager)
+
+#### Command Options (V3 Mode)
+- `--license`: **Required** - License file for authentication with replicated.app
+- `--version`: Optional parameter to specify target version (defaults to latest)
+- `--force`: Skip confirmation prompts for automated scenarios
```
sgalsaleh, line 48: if it's just for skipping prompts, then i believe it should be `--yes`

design/embedded-cluster-cli-driven-update.md
```diff
@@ -0,0 +1,278 @@
...
+**Dual Sequence Handling:**
+The embedded cluster must handle multiple versioning schemes simultaneously, each serving different purposes:
+
+- **Upstream cursors**: Channel sequence numbers from API (e.g., 1247, 1248, 1251, 1255)
+  - Used for API synchronization and incremental updates
+  - Can have gaps when releases are skipped or removed
+  - Channel-specific and may restart across different channels
+  - Essential for efficient API communication
+
+- **Version labels**: User-facing identifiers (e.g., v1.2.0, v1.2.1)
```
sgalsaleh, line 73: why not just support version labels? what's the need to support channel sequences and embedded cluster technical versions?

design/embedded-cluster-cli-driven-update.md
```diff
@@ -0,0 +1,278 @@
...
+**Example Mapping:**
+```
+API Release          User Experience      Technical Details
+─────────────       ─────────────────     ─────────────────
+Cursor: 1247    →   Version: v1.2.0   →   EC: 1.5.0+k8s-1.29
+Cursor: 1248    →   Version: v1.2.1   →   EC: 1.5.0+k8s-1.29  
+Cursor: 1251    →   Version: v1.3.0   →   EC: 1.6.0+k8s-1.30
+```
+
+This multi-layered approach allows the binary to handle API efficiency, user experience, and technical compatibility as separate concerns.
```
sgalsaleh, line 94: not sure i understand what this means. seems like complexity that isn't necessary?

design/embedded-cluster-cli-driven-update.md
sgalsaleh: why does this need to be a new endpoint? can't we just use the existing endpoint? we would have all the information we need about the current release and can send the channel cursor, version label, etc... and the current endpoint would return available updates

design/embedded-cluster-cli-driven-update.md
```diff
@@ -0,0 +1,278 @@
...
+- Tracking user's previous update path for required release validation
+- Maintaining cursor positions for incremental API synchronization
+
+### 2. Cross-Channel Semver Support
+
+**Question:** Do we need to account for cross-channel semantic version comparisons, or can we simplify by restricting version comparisons to the same channel?
+
+**Context:** The current design includes cross-channel semver functionality where:
+- Semver-enabled channels can compare versions across different channels (e.g., v1.2.3 from stable > v1.2.1 from beta)
+- Required releases can block updates across ALL channels when semver is enabled
```
sgalsaleh, line 263: That's not how it works in KOTS, and I don't think this makes sense either. I can mark the same version / semver as required on a certain channel, but not required on another.

design/embedded-cluster-cli-driven-update.md
```diff
@@ -0,0 +1,278 @@
...
+- This adds significant complexity to the version comparison logic
+
+### 3. Cursor-Based vs Semver-Only Version Management
```
sgalsaleh, line 266: i don't think the client (EC CLI) should even care about semver being enabled or not. the sorting mechanism lives in replicated.app, and the client should simply check a linear list for required / non-required releases. for example, if you're trying to deploy a specific release, can just check if there's a required release before it in the list of available updates returned from replicated.app. there's no need to sort or even know if it's semver or not in this case.

design/embedded-cluster-cli-driven-update.md
```diff
@@ -0,0 +1,278 @@
...
+API Release          User Experience      Technical Details
+─────────────       ─────────────────     ─────────────────
+Cursor: 1247    →   Version: v1.2.0   →   EC: 1.5.0+k8s-1.29
+Cursor: 1248    →   Version: v1.2.1   →   EC: 1.5.0+k8s-1.29  
+Cursor: 1251    →   Version: v1.3.0   →   EC: 1.6.0+k8s-1.30
+```
+
+This multi-layered approach allows the binary to handle API efficiency, user experience, and technical compatibility as separate concerns.
+
+#### Version Comparison Logic
```
sgalsaleh, line 96: i don't think this should be a thing in embedded cluster. all this logic should only live in replicated.app

design/embedded-cluster-cli-driven-update.md
diamonwiggins: Which existing endpoint? The one where we download the binary currently, `GET /embedded/:appSlug/:channelSlug/:versionLabel`? Or the existing endpoints already being used to check for available releases, `/release/{appSlug}/pending`?

design/embedded-cluster-cli-driven-update.md
diamonwiggins: i had the same question, and raised it at the bottom of the doc in the questions section. I think just supported version labels is the way to go as well.

design/embedded-cluster-cli-driven-update.md
diamonwiggins: Makes sense. Is this how it works in KOTS as well? Either way, with us not worrying about sequences and just versions, this is the natural conclusion as well i think.

PR #2434, opened 2025-07-08
Add API endpoint to set config values

sgalsaleh:

Adds an API endpoint to set config values

[SC-126236](https://app.shortcut.com/replicated/story/126236)

Yes

```release-note
NONE
```

NONE


diamonwiggins (dismissed):
lgtm

api/internal/managers/app/config/config.go
```diff
@@ -4,12 +4,39 @@ import (
...
+				value, ok := configValues.Spec.Values[c.Name]
+				if ok {
+					updatedConfig.Spec.Groups[idxG].Items[idxI].Items[idxC].Value = multitype.FromString(value.Value)
+				}
+			}
+		}
+	}
+
+	return updatedConfig, nil
 }
```
JGAntunes, line 42: The other two methods are just proxying to the store but seems like it isn't the case for this method, should we add a test for it?

api/internal/handlers/kubernetes/install.go
```diff
@@ -128,15 +129,65 @@ func (h *Handler) GetInfraStatus(w http.ResponseWriter, r *http.Request) {
 //	@Tags			kubernetes-install
 //	@Security		bearerauth
 //	@Produce		json
-//	@Success		200	{object}	v1beta1.Config
+//	@Success		200	{object}	kotsv1beta1.Config
 //	@Router			/kubernetes/install/app/config [get]
 func (h *Handler) GetAppConfig(w http.ResponseWriter, r *http.Request) {
-	config, err := h.installController.GetAppConfig(r.Context())
+	// import is used in swagger annotation above
+	_ = kotsv1beta1.Config{}
```
JGAntunes, line 136: Should this be outside of the handler? It's probably cheap enough but we're creating a config struct every single time on request.

api/types/requests.go
```diff
@@ -0,0 +1,8 @@
+package types
+
+import kotsv1beta1 "github.com/replicatedhq/kotskinds/apis/kots/v1beta1"
+
+// SetAppConfigValuesRequest represents the request when setting the app config values
+type SetAppConfigValuesRequest struct {
+	Values map[string]kotsv1beta1.ConfigValue `json:"values"`
```
emosbaugh, line 7: Why is this not just a map[string]string?

api/internal/handlers/kubernetes/install.go
```diff
@@ -128,15 +129,65 @@ func (h *Handler) GetInfraStatus(w http.ResponseWriter, r *http.Request) {
...
+
+	if h.cfg.ReleaseData == nil || h.cfg.ReleaseData.AppConfig == nil {
+		utils.LogError(r, errors.New("app config not found"), h.logger)
+		utils.JSONError(w, r, errors.New("app config not found"), h.logger)
+		return
+	}
```
JGAntunes, line 142: Shouldn't this live in the controller?

api/controllers/kubernetes/install/app_test.go
```diff
@@ -1,112 +0,0 @@
...
-	"github.com/stretchr/testify/require"
-)
-
-func TestInstallController_GetAppConfig(t *testing.T) {
-	tests := []struct {
-		name           string
-		releaseData    *release.ReleaseData
-		expectedConfig kotsv1beta1.Config
-		expectedError  bool
-	}{
```
JGAntunes, line 20: I get that right now the controller isn't doing much besides proxying requests to the manager but I would imagine that soon enough that might not be the case(?) E.g. doing the validation that we're currently doing in the handler, or the state transitions. Should we keep this test file around even if the only tests we have are 2 with the manager mocked where it errors if the manager returns an error and returns the manager response otherwise?

No strong feelings about it though.

api/internal/handlers/linux/install.go
```diff
@@ -207,15 +208,65 @@ func (h *Handler) GetInfraStatus(w http.ResponseWriter, r *http.Request) {
 //	@Tags			linux-install
 //	@Security		bearerauth
 //	@Produce		json
-//	@Success		200	{object}	v1beta1.Config
+//	@Success		200	{object}	kotsv1beta1.Config
 //	@Router			/linux/install/app/config [get]
 func (h *Handler) GetAppConfig(w http.ResponseWriter, r *http.Request) {
-	config, err := h.installController.GetAppConfig(r.Context())
+	// import is used in swagger annotation above
+	_ = kotsv1beta1.Config{}
```
JGAntunes, line 215: Same as the other handler ☝️

api/internal/handlers/linux/install.go
```diff
@@ -207,15 +208,65 @@ func (h *Handler) GetInfraStatus(w http.ResponseWriter, r *http.Request) {
...
+
+	if h.cfg.ReleaseData == nil || h.cfg.ReleaseData.AppConfig == nil {
+		utils.LogError(r, errors.New("app config not found"), h.logger)
+		utils.JSONError(w, r, errors.New("app config not found"), h.logger)
+		return
+	}
```
JGAntunes, line 221: Same as the other handler ☝️

api/internal/managers/app/config/manager.go
```diff
@@ -14,8 +14,9 @@ var _ AppConfigManager = &appConfigManager{}
 
 // AppConfigManager provides methods for managing appConfigstructure setup
 type AppConfigManager interface {
```
JGAntunes, line 16: Happy to keep it separate since you're no longer touching that code but we can remove this mutex since it's unused I think.
https://github.com/replicatedhq/embedded-cluster/blob/fefb871ad0751c8dc37a75b81043b2c514c37093/api/internal/managers/app/config/manager.go#L26

emosbaugh (approved):

PR #2440, opened 2025-07-09
feat: pass config values from memory store to kots cli

diamonwiggins:

This PR implements the ability to pass user-provided config values from the API memory store to KOTS installation during Linux installs.

Fixes: https://app.shortcut.com/replicated/story/126282/pass-user-provided-config-values-to-kots-install-for-linux-installs

**YES**

The tests verify:
- Correct YAML format generation (`apiVersion: kots.io/v1beta1`, `kind: ConfigValues`)
- File creation in proper directory structure
- Priority logic (CLI file > memory store values)
- Backward compatibility with existing file-based workflows

```release-note
NONE
```

**NONE**

#### Additional Details

The following POST results in an updated config values in the API's memory store
```
embedded-cluster git:(diamonwiggins/sc-126282/pass-config-values-v3) ✗ curl -k -X POST ${API_BASE}/api/linux/install/app/config/values \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{
    "values": {
      "hostname": "my-server.example.com",
      "pw": "*********"
    }
  }' | jq
```

The config values are retrieved from the memory store and persisted in a tmp directory so they can be used by the KOTS CLI when installing the application.


cmd/installer/kotscli/kotscli.go
```diff
@@ -79,8 +84,16 @@ func Install(opts InstallOptions) error {
 		installArgs = append(installArgs, "--airgap-bundle", opts.AirgapBundle)
 		maskfn = MaskKotsOutputForAirgap()
 	}
+
 	if opts.ConfigValuesFile != "" {
 		installArgs = append(installArgs, "--config-values", opts.ConfigValuesFile)
+	} else if opts.ConfigValues != nil {
+		configValuesFile, err := createConfigValuesFile(opts.ConfigValues, opts.RuntimeConfig)
```
sgalsaleh, line 91: i think the caller should simply create the file and pass it via `ConfigValuesFile` instead of having the kots cli do this step. that keeps the kotscli package a light wrapper around the kots cli and easy to take out in the future.

sgalsaleh: actually, a better suggestion maybe is to pass config values as a map and then this package converts it to a `kind: ConfigValues` file. that way, the ConfigValues logic is only coupled to the KOTS package.

api/controllers/linux/install/infra.go
```diff
@@ -67,6 +67,10 @@ func (c *InstallController) SetupInfra(ctx context.Context, ignoreHostPreflights
 			}
 		}()
 
+		if err := c.updateInfraManagerWithLatestConfigValues(); err != nil {
```
sgalsaleh, line 70: why not simply get the values and pass them to the `Install(` function below?

api/internal/managers/linux/infra/manager.go
```diff
@@ -175,3 +191,9 @@ func NewInfraManager(opts ...InfraManagerOption) *infraManager {
 func (m *infraManager) Get() (types.Infra, error) {
 	return m.infraStore.Get()
 }
+
+func (m *infraManager) UpdateConfigValues(configValues map[string]string) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.configValues = configValues
```
sgalsaleh, line 198: this should be passed to the infraManager.Install function as an arg instead of updating a global variable like this

cmd/installer/kotscli/kotscli.go
```diff
@@ -79,8 +84,16 @@ func Install(opts InstallOptions) error {
 		installArgs = append(installArgs, "--airgap-bundle", opts.AirgapBundle)
 		maskfn = MaskKotsOutputForAirgap()
 	}
+
 	if opts.ConfigValuesFile != "" {
 		installArgs = append(installArgs, "--config-values", opts.ConfigValuesFile)
+	} else if len(opts.ConfigValues) > 0 {
+		configValuesFile, err := createConfigValuesFile(opts.ConfigValues, opts.RuntimeConfig.EmbeddedClusterHomeDirectory())
```
sgalsaleh, line 91: let's store this file in the embedded cluster tmp dir and defer remove it as it can contain sensitive info

api/controllers/linux/install/infra.go
diamonwiggins: felt like that was going to sprawl into a lot more changes if the function signature needed to change. can do that thought if its what we want

api/controllers/linux/install/controller_test.go
```diff
@@ -1145,6 +1149,12 @@ func TestSetupInfra(t *testing.T) {
 			mockInfraManager := &infra.MockInfraManager{}
 			mockMetricsReporter := &metrics.MockReporter{}
 			mockStore := &store.MockStore{}
+			mockAppConfigManager := &appconfig.MockAppConfigManager{}
+
+			// Setup store mock instead of appConfigManager mock since updateInfraManagerWithLatestConfigValues
+			// now calls c.store.AppConfigStore().GetConfigValues() instead of c.appConfigManager.GetConfigValues()
+			mockStore.AppConfigMockStore.On("GetConfigValues").Return(map[string]string{}, nil)
```
sgalsaleh, line 1156: why isn't this part of the "setupMocks" function?

api/controllers/linux/install/infra.go
```diff
@@ -67,7 +67,16 @@ func (c *InstallController) SetupInfra(ctx context.Context, ignoreHostPreflights
 			}
 		}()
 
-		if err := c.infraManager.Install(ctx, c.rc); err != nil {
+		configValues := make(map[string]string)
+		if c.store != nil && c.store.AppConfigStore() != nil {
```
sgalsaleh, line 71: nit: no need to check if stores are nil, nothing would function if that was true.

api/controllers/linux/install/infra.go
```diff
@@ -67,7 +67,16 @@ func (c *InstallController) SetupInfra(ctx context.Context, ignoreHostPreflights
...
+			if values, err := c.store.AppConfigStore().GetConfigValues(); err != nil {
+				c.logger.WithError(err).Warn("reading config values from store")
```
sgalsaleh, line 73: this should return an error, not just log it

sgalsaleh (approved):

sgalsaleh (approved):

PR #2447, opened 2025-07-09
feat(ui): add support for textarea config option

diamonwiggins:

This PR adds support for `textarea` config item types in the application configuration UI, enabling users to input multi-line text values for configuration options.

Fixes: [SC-126347](https://app.shortcut.com/replicated/story/126347/support-the-textarea-config-item-type) 

**YES**

```release-note
NONE
```

**NONE**

#### Additional Details
kind: Config
```yaml
apiVersion: kots.io/v1beta1
kind: Config
spec:
  groups:
    - name: config_group
      title: The First Config Group
      items:
        - name: hostname
          title: Hostname
          type: text
        - name: pw
          title: Password
          type: password
        - name: some-text-area
          title: Application Description
          type: textarea
```
**NOTE**: Screenshot is out of data as the default value is no longer shown as a placeholder
Screenshot 2025-07-09 at 5 30 14 PM
Screenshot 2025-07-09 at 5 31 03 PM


web/src/components/wizard/config/ConfigurationStep.tsx
```diff
@@ -147,6 +148,18 @@ const ConfigurationStep: React.FC<ConfigurationStepProps> = ({ onNext }) => {
           />
         );
 
+      case 'textarea':
+        return (
+          <Textarea
+            id={item.name}
+            label={item.title}
+            value={item.value || ''}
+            placeholder={item.default}
```
ajp-io, line 157: it was probably this way in bolt, but from what i read, it's an anti-pattern to use the default value as a placeholder. i don't think the KOTS config uses placeholders at all, so we can just not have placehodlers on text or textarea items.

web/src/components/wizard/config/ConfigurationStep.tsx
```diff
@@ -141,12 +142,22 @@ const ConfigurationStep: React.FC<ConfigurationStepProps> = ({ onNext }) => {
             id={item.name}
             label={item.title}
             value={item.value || ''}
-            placeholder={item.default}
```
sgalsaleh, line 144: Why remove this?

diamonwiggins: see discussion above with Alex

ajp-io: https://github.com/replicatedhq/embedded-cluster/pull/2447#discussion_r2196280441

ajp-io: i actually just tested with KOTS, and it does use the default as the placeholder. but like i said above, i do see online that that's an anti-pattern because it's not what placeholders are intended for, so i still support showing the default value beneath the text area instead.

PR #2450, opened 2025-07-10
feat(api): handle config item "when" attribute

emosbaugh:

Only returns items with when != "false" to the ui.

Filters values for items with when == "false"

```release-note

```

### DIFF
--- a/api/internal/managers/app/config/config.go
+++ b/api/internal/managers/app/config/config.go
@@ -1,13 +1,82 @@
 package config
 
 import (
-	"context"
+	"fmt"
+
+	kotsv1beta1 "github.com/replicatedhq/kotskinds/apis/kots/v1beta1"
+	"github.com/replicatedhq/kotskinds/multitype"
+	"github.com/tiendc/go-deepcopy"
 )
+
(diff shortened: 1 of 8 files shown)


api/internal/managers/app/config/config.go
```diff
@@ -1,13 +1,82 @@
...
-func (m *appConfigManager) SetConfigValues(ctx context.Context, values map[string]string) error {
-	return m.appConfigStore.SetConfigValues(values)
+func (m *appConfigManager) SetConfigValues(config kotsv1beta1.Config, configValues map[string]string) error {
+	filteredValues := make(map[string]string)
+
+	// only include values for enabled groups and items, otherwise keep the original value
+	for _, g := range config.Spec.Groups {
+		for _, i := range g.Items {
+			if !isItemEnabled(g.When) || !isItemEnabled(i.When) {
+				filteredValues[i.Name] = i.Value.String()
```
sgalsaleh, line 26: wouldn't this be applying the value from the config to config values and potentially be adding values the user never submitted? i think we should add to filtered values if item is enabled, but not add at all if not, no?

PR #2460, opened 2025-07-11
chore(web): consistent usage of the intial state context

JGAntunes:
Quick cleanup of the initial state context component:
- Renamed it from Branding* to what it actually is (Initial State)
- Added some types and type guards
- Use the InitialStateContext everywhere instead of accessing `window.*` directly.
- Some tests
- Some type cleanups.

NA

Adds tests

```release-note
NONE
```

NONE


diamonwiggins (approved):
lgtm

PR #2479, opened 2025-07-14
feat(api): support Go templating with sprig in config resource

sgalsaleh:

Adds support for Go templating with Sprig functions for the app kotsv1beta1.Config resource. This allows dynamic configuration using template expressions like {{ upper "text" }}, {{ add 9000 90 }}, and {{printf "Port: %d" 8080 }} in the config. The AppConfigManager now stores the raw config internally and processes templates on-demand.

[SC-126445](https://app.shortcut.com/replicated/story/126445)

Yes

```release-note
NONE
```

NONE

### DIFF
--- a/api/controllers/kubernetes/install/app.go
+++ b/api/controllers/kubernetes/install/app.go
@@ -11,15 +11,15 @@ import (
 )
 
 func (c *InstallController) GetAppConfig(ctx context.Context) (kotsv1beta1.Config, error) {
-	if c.releaseData == nil || c.releaseData.AppConfig == nil {
+	if c.appConfigManager == nil {
 		return kotsv1beta1.Config{}, errors.New("app config not found")
 	}
 
-	return c.appConfigManager.GetConfig(*c.releaseData.AppConfig)
+	return c.appConfigManager.GetConfig()
 }
 
 func (c *InstallController) PatchAppConfigValues(ctx context.Context, values map[string]string) (finalErr error) {
-	if c.releaseData == nil || c.releaseData.AppConfig == nil {
+	if c.appConfigManager == nil {
 		return errors.New("app config not found")
 	}
 
@@ -51,12 +51,12 @@ func (c *InstallController) PatchAppConfigValues(ctx context.Context, values map
 		}
 	}()
 
-	err = c.appConfigManager.ValidateConfigValues(*c.releaseData.AppConfig, values)
+	err = c.appConfigManager.ValidateConfigValues(values)
 	if err != nil {
 		return fmt.Errorf("validate app config values: %w", err)
 	}
 
-	err = c.appConfigManager.PatchConfigValues(*c.releaseData.AppConfig, values)
+	err = c.appConfigManager.PatchConfigValues(values)
 	if err != nil {
 		return fmt.Errorf("patch app config values: %w", err)
 	}
@@ -70,11 +70,8 @@ func (c *InstallController) PatchAppConfigValues(ctx context.Context, values map
 }
 
 func (c *InstallController) GetAppConfigValues(ctx context.Context, maskPasswords bool) (map[string]string, error) {
-	// Get the app config to determine which fields are password type
-	appConfig, err := c.GetAppConfig(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("get app config: %w", err)
+	if c.appConfigManager == nil {
+		return nil, errors.New("app config not found")
 	}
-
-	return c.appConfigManager.GetConfigValues(appConfig, maskPasswords)
+	return c.appConfigManager.GetConfigValues(maskPasswords)
 }
--- a/api/controllers/kubernetes/install/controller.go
+++ b/api/controllers/kubernetes/install/controller.go
@@ -223,11 +223,11 @@ func NewInstallController(opts ...InstallControllerOption) (*InstallController, 
 		if controller.releaseData == nil || controller.releaseData.AppConfig == nil {
 			return nil, errors.New("app config not found")
 		}
-		err := controller.appConfigManager.ValidateConfigValues(*controller.releaseData.AppConfig, controller.configValues)
+		err := controller.appConfigManager.ValidateConfigValues(controller.configValues)
 		if err != nil {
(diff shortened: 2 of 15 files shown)


api/internal/managers/app/config/template.go
```diff
@@ -0,0 +1,25 @@
...
+	}
+
+	var buf bytes.Buffer
+	err = tmpl.Execute(&buf, nil) // No custom context - vanilla Go templates only
+	if err != nil {
+		return "", fmt.Errorf("template execution error: %w", err)
+	}
+
+	return buf.String(), nil
+}
```
emosbaugh, line 25: looks like you need to run go fmt

api/internal/managers/app/config/template_test.go
```diff
@@ -0,0 +1,139 @@
...
+        help: Port number (default: 8080)
+`,
+			hasError: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result, err := manager.processTemplate(tt.input)
+
```
emosbaugh, line 129: looks like you need to run go fmt

api/internal/managers/app/config/config.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/replicatedhq/kotskinds/multitype"
 	"github.com/tiendc/go-deepcopy"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
```
emosbaugh, line 13: in the api this is imported everywhere else as kyaml

```
kyaml "sigs.k8s.io/yaml"
```

api/internal/managers/app/config/config.go
```diff
@@ -22,16 +23,34 @@ var (
...
+
+	return filterAppConfig(processedConfig)
 }
 
-func (m *appConfigManager) ValidateConfigValues(config kotsv1beta1.Config, configValues map[string]string) error {
+func (m *appConfigManager) ValidateConfigValues(configValues map[string]string) error {
 	var ve *types.APIError
 
-	filteredConfig, err := filterAppConfig(config)
+	filteredConfig, err := m.GetConfig()
```
emosbaugh, line 51: for consistency

```suggestion
	processedConfig, err := m.GetConfig()
```

api/controllers/kubernetes/install/app.go
```diff
@@ -11,15 +11,15 @@ import (
 )
 
 func (c *InstallController) GetAppConfig(ctx context.Context) (kotsv1beta1.Config, error) {
-	if c.releaseData == nil || c.releaseData.AppConfig == nil {
+	if c.appConfigManager == nil {
```
emosbaugh, line 14: are these checks still needed? we dont check that a manager is nil anywhere else

api/controllers/kubernetes/install/controller.go
```diff
@@ -195,24 +194,25 @@ func NewInstallController(opts ...InstallControllerOption) (*InstallController,
...
-			appconfig.WithAppConfigStore(controller.store.AppConfigStore()),
-		)
-	}
-
-	if controller.configValues != nil {
-		if controller.releaseData == nil || controller.releaseData.AppConfig == nil {
-			return nil, errors.New("app config not found")
+	if controller.releaseData != nil && controller.releaseData.AppConfig != nil {
+		// Initialize the app config manager if an app config is provided
+		if controller.appConfigManager == nil {
```
emosbaugh, line 199: shouldnt we always instantiate this and instead fail on nil controller.releaseData.AppConfig?

api/controllers/linux/install/controller.go
```diff
@@ -256,17 +248,25 @@ func NewInstallController(opts ...InstallControllerOption) (*InstallController,
 		)
 	}
 
-	if controller.configValues != nil {
-		if controller.releaseData == nil || controller.releaseData.AppConfig == nil {
-			return nil, errors.New("app config not found")
+	if controller.releaseData != nil && controller.releaseData.AppConfig != nil {
+		// Initialize the app config manager if an app config is provided
+		if controller.appConfigManager == nil {
```
emosbaugh, line 253: shouldnt we always instantiate this and instead fail on nil controller.releaseData.AppConfig?

api/internal/managers/app/config/template.go
```diff
@@ -0,0 +1,25 @@
...
+
+func (m *appConfigManager) processTemplate(configYAML string) (string, error) {
+	if configYAML == "" {
+		return configYAML, nil
+	}
+
+	tmpl, err := m.templateEngine.Parse(configYAML)
+	if err != nil {
+		return "", fmt.Errorf("template parse error: %w", err)
+	}
```
emosbaugh, line 16: why marshal and parse this each time we get the config? why not do it when we instantiate appConfigManager and store it as a property?

api/internal/managers/app/config/config.go
```diff
@@ -22,16 +23,34 @@ var (
...
 )
 
-func (m *appConfigManager) GetConfig(config kotsv1beta1.Config) (kotsv1beta1.Config, error) {
-	return filterAppConfig(config)
+func (m *appConfigManager) GetConfig() (kotsv1beta1.Config, error) {
+	// Convert config to YAML string
+	configYAML, err := yaml.Marshal(m.rawConfig)
+	if err != nil {
+		return kotsv1beta1.Config{}, fmt.Errorf("marshal config to yaml: %w", err)
+	}
```
emosbaugh, line 31: why marshal this each time we get the config?

emosbaugh (changes requested):

api/controllers/kubernetes/install/app.go
sgalsaleh: right now there's a limitation where v3 doesn't work without a config. i can create a story to remove that limitation as i see the controller returns an error if there's no config at the moment.

JGAntunes: Agree, if it's data we provide at instantiation time we should do it there and only once, no point in spreading this nil check logic around unless we absolutely need to.

emosbaugh (approved):

PR #2480, opened 2025-07-15
feat(ui:) add support for default config values

diamonwiggins:

This PR implements frontend support for displaying and handling default values for configuration items in the UI.

Fixes [SC-126434](https://app.shortcut.com/replicated/story/126434/support-the-default-config-item-property)

**YES**

```release-note
NONE
```

NONE

#### Screenshots
Screenshot 2025-07-14 at 10 43 06 PM


web/src/components/wizard/config/ConfigurationStep.tsx
```diff
@@ -141,6 +142,40 @@ const ConfigurationStep: React.FC<ConfigurationStepProps> = ({ onNext }) => {
     return configValues?.[item.name] ?? (item.value || item.default || '');
   };
 
+  // Helper function to determine if default value should be shown
+  const shouldShowDefault = (item: AppConfigItem): boolean => {
+    return !!(item.default && ['text', 'password', 'textarea'].includes(item.type));
+  };
+
+  // Helper function to render help text with default value inline
+  const renderHelpTextWithDefault = (helpText?: string, defaultValue?: string, error?: string) => {
```
diamonwiggins, line 151: @JGAntunes was on the fence about this since `default` and `help_text` are two separate properties in the config, but the desire here was to keep them on the same line with each other since they are also very related and both technically "help text". Let me know what you think.

web/src/components/wizard/config/ConfigurationStep.tsx
```diff
@@ -141,6 +141,13 @@ const ConfigurationStep: React.FC<ConfigurationStepProps> = ({ onNext }) => {
     return configValues?.[item.name] ?? (item.value || item.default || '');
   };
 
+  // Helper function to determine if default value should be shown
+  const shouldShowDefault = (item: AppConfigItem): boolean => {
+    return !!(item.default && ['text', 'password', 'textarea'].includes(item.type));
+  };
+
+
```
sgalsaleh, line 149: there's extra lines here

web/src/components/common/HelpText.tsx
```diff
@@ -0,0 +1,38 @@
...
+  let combinedText = helpText || '';
+  if (defaultValue) {
+    const defaultText = `(Default: \`${defaultValue}\`)`;
+    combinedText = helpText ? `${helpText} ${defaultText}` : defaultText;
+  }
+  
+  return (
+    <div className="mt-1 text-sm text-gray-500">
+      <ReactMarkdown
+        components={{
```
emosbaugh, line 23: why not use the same <a /> alias as Label component?

web/src/components/wizard/config/ConfigurationStep.tsx
```diff
@@ -192,6 +199,7 @@ const ConfigurationStep: React.FC<ConfigurationStepProps> = ({ onNext }) => {
         return (
           <Input
             {...sharedProps}
+            defaultValue={shouldShowDefault(item) ? item.default : undefined}
```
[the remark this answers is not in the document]
diamonwiggins: After a discussion with Alex, we don't want to show defaults for certain config items like "radio, checkbox, and label"

web/src/components/wizard/tests/ConfigurationStep.test.tsx
```diff
@@ -1507,4 +1509,145 @@ describe.each([
     const enableSslField = screen.getByTestId("bool-input-enable_ssl") as HTMLInputElement;
     expect(enableSslField.checked).toBe(true); // default "1" is used since getEffectiveValue includes default
   });
+
+  describe("Default value utility functions", () => {
+    // Test version of the component using the new HelpText component
+    const TestComponent = () => {
```
sgalsaleh, line 1515: why create an entire new test component just to test the HelpText component? this test doesn't seem relevant to the ConfigurationStep component.

diamonwiggins: yeah, you're right. i think this is left over from the prior implementation

web/src/components/common/HelpText.tsx
diamonwiggins: yeah, i'll make these consistent

web/src/components/wizard/config/ConfigurationStep.tsx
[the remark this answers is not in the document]
sgalsaleh: but those components don't pass the default value to the HelpText component so that should be covered there? that was the point of my initial suggestion

web/src/components/common/HelpText.tsx
```diff
@@ -0,0 +1,38 @@
...
+          code: ({ children }) => (
+            <code className="font-mono text-xs bg-gray-100 px-1 py-0.5 rounded">
+              {children}
+            </code>
+          ),
+          p: ({ children }) => <span>{children}</span>, // Render as span instead of paragraph
```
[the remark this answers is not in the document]
diamonwiggins: yeah, this is to prevent any vertical spacing. i am frontend challenged, so if there's a better way, lmk

sgalsaleh (approved):

PR #2482, opened 2025-07-15
try to resolve relative data dir paths

hedge-sparrow:

If you specify a relative path to the embedded cluster installer's `--data-dir` flag, the install will fail when starting the LAM service. we should resolve the path given to us to an absolute path to avoid ever dealing with relative paths. 

```release-note
Fixed an issue where specifying a relative path to the installer's --data-dir flag would cause the installation to fail when starting the LAM service. Paths provided to --data-dir are now resolved to absolute paths, avoiding problems related to relative directory references.
```

### DIFF
--- a/cmd/installer/cli/install.go
+++ b/cmd/installer/cli/install.go
@@ -503,7 +503,12 @@ func preRunInstallLinux(cmd *cobra.Command, flags *InstallCmdFlags, rc runtimeco
 	}
 
 	// TODO: validate that a single port isn't used for multiple services
-	rc.SetDataDir(flags.dataDir)
+	// resolve datadir to absolute path
+	AbsoluteDataDir, err := filepath.Abs(flags.dataDir)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/installer/cli/install.go
```diff
@@ -503,7 +503,12 @@ func preRunInstallLinux(cmd *cobra.Command, flags *InstallCmdFlags, rc runtimeco
 	}
 
 	// TODO: validate that a single port isn't used for multiple services
-	rc.SetDataDir(flags.dataDir)
+	// resolve datadir to absolute path
+	AbsoluteDataDir, err := filepath.Abs(flags.dataDir)
+	if err != nil {
+		return fmt.Errorf("unable to construct path for directory: %w", err)
+	}
+	rc.SetDataDir(AbsoluteDataDir)
```
diamonwiggins, line 511:
```suggestion
	rc.SetDataDir(absoluteDataDir)
```

diamonwiggins (approved):
lgtm

PR #2491, opened 2025-07-16
feat(app): add support for the file type item

JGAntunes:

This PR adds actual support for the config item file type, validating the its value and setting the correct `Filename` property based on pre-existing values or the parent item filename.

Closes https://app.shortcut.com/replicated/story/126680/add-file-config-item-type-backend-foundation

Yes

```release-note
NONE
```

NONE

### DIFF
--- a/api/internal/managers/app/config/config.go
+++ b/api/internal/managers/app/config/config.go
@@ -309,6 +309,7 @@ func isFileType(item kotsv1beta1.ConfigItem) bool {
 	return item.Type == "file"
 }
 
+// isValueBase64Encoded checks if the value of a ConfigValue is base64 encoded, this is used for file items
 func isValueBase64Encoded(configValue kotsv1beta1.ConfigValue) bool {
 	if configValue.Value == "" {
 		return true // empty values are considered valid
--- a/api/internal/managers/app/config/config.go
+++ b/api/internal/managers/app/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"maps"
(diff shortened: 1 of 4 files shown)


sgalsaleh (approved):

sgalsaleh:
my only comment is that it took my a while to grok what [all of these functions](https://github.com/replicatedhq/embedded-cluster/blob/25c7dc1d229ae0ccaa1d0bf6c068927b847aa245/api/internal/managers/app/config/config.go#L200-L250) do and how the are tied together. is there a way to reduce the number of these functions into maybe 1 or 2 and give them names that are indicative of what’s happening at a higher level? there was only getConfigValueFromChildItem previously i think which was easy to follow.

api/internal/managers/app/config/config.go
```diff
@@ -182,58 +189,66 @@ func (m *appConfigManager) GetKotsadmConfigValues() (kotsv1beta1.ConfigValues, e
...
 
-	// override values from the config values store
-	if v, ok := configValues[item.Name]; ok {
-		if item.Type == "password" {
-			configValue.ValuePlaintext = v.Value
-		} else {
-			configValue.Value = v.Value
-		}
+// applyStoredValueOverride applies stored values from configValues if they exist
+func applyStoredValueOverride(cv *kotsv1beta1.ConfigValue, itemType string, itemName string, configValues types.AppConfigValues) {
```
sgalsaleh, line 210: my only comment is that it took my a while to grok what [all of these functions](https://github.com/replicatedhq/embedded-cluster/blob/25c7dc1d229ae0ccaa1d0bf6c068927b847aa245/api/internal/managers/app/config/config.go#L200-L250) do and how the are tied together. is there a way to reduce the number of these functions into maybe 1 or 2 and give them names that are indicative of what’s happening at a higher level? there was only getConfigValueFromChildItem previously i think which was easy to follow.

JGAntunes: Will add it as follow up.

JGAntunes: Done via https://github.com/replicatedhq/embedded-cluster/pull/2500

PR #2515, opened 2025-07-18
Support repl templates - Manager Integration

sgalsaleh:

Support repl templates - Manager Integration

[SC-126941](https://app.shortcut.com/replicated/story/126941)

Yes

```release-note
NONE
```

NONE


api/internal/managers/app/config/manager.go
```diff
@@ -26,14 +27,18 @@ type AppConfigManager interface {
 	// GetKotsadmConfigValues merges the config values with the app config defaults and returns a
 	// kotsv1beta1.ConfigValues struct.
 	GetKotsadmConfigValues() (kotsv1beta1.ConfigValues, error)
+	// TemplateConfig templates the config with provided values and returns the templated config
+	TemplateConfig(configValues types.AppConfigValues) (types.AppConfig, error)
```
emosbaugh, line 31: This function and the GetConfig function seem out of place together. Why does one return an api type and the other the raw config? Why arent the names more similar? I had to look at the function bodies to understand them. Could you clarify in the doc string, return the same type (types.AppConfig), and maybe try naming the two differently? Perhaps GetConfig and GetConfigWithValues. I'm open to other suggestions.

emosbaugh (changes requested):

emosbaugh (approved):

PR #2518, opened 2025-07-18
chore(ci): dryrun test airgap prompt

emosbaugh:

Adds tests to validate that we prompt to use the airgap flag when this release was downloaded with an airgap bundle.

```release-note

```


screspod (approved):
Nice! Perhaps we should do the same for the prompts you get when you use --http-proxy without --https-proxy and installing an old version (in a different story).

.cursor/rules/code-quality-checks.mdc
```diff
@@ -48,6 +51,7 @@ Before committing changes, ensure you have:
 - [ ] Checked for proper data-testid attributes in new components
 
 ### For Go Changes
+- [ ] Ran `go fmt ./...` to ensure consistent formatting
```
sgalsaleh, line 54: ah thanks! i thought i had this

pkg/dryrun/dryrun.go
```diff
@@ -26,6 +28,7 @@ import (
 var (
 	dr     *types.DryRun
 	drFile string
+	logout *bytes.Buffer
```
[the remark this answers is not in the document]
sgalsaleh: ah nevermind, looks like you did.

sgalsaleh, line 31: why does this need to be here and also in types.DryRun?

sgalsaleh: and why is it called logout? :)

sgalsaleh (approved):

PR #2528, opened 2025-07-21
feat(file-input): add support for default values

JGAntunes:
This PR adds support for default values for file inputs.

If a default value is provided we:
- Highlight it in the help text
- Allow a user to download the default file
- The filename will use the config item name if a filename isn't available (i.e. a default value is present or a value is present that wasn’t provided by the user).
- A user cannot remove the default file.
- The default file will be replaced if a file is uploaded.

Loom:
- https://www.loom.com/share/1eb4ab0f526e4e65ae8dd5b7ea12f45f?sid=c6e08a5e-88f8-4df6-920e-1921e2dce716

https://app.shortcut.com/replicated/story/126684/add-file-config-item-type-default-value-support

yes

```release-note
NONE
```

NONE


web/src/components/common/HelpText.tsx
```diff
@@ -19,7 +20,7 @@ const HelpText: React.FC<HelpTextProps> = ({ helpText, defaultValue, error }) =>
   }
 
   return (
-    <div className="mt-1 text-sm text-gray-500 [&_p]:inline [&_p]:mb-0">
+    <div data-testid={dataTestId ? `${dataTestId}-help-text` : "help-text"} className="mt-1 text-sm text-gray-500 [&_p]:inline [&_p]:mb-0">
```
sgalsaleh, line 23: for consistency
```suggestion
    <div data-testid={dataTestId ? `help-text${dataTestId}` : "help-text"} className="mt-1 text-sm text-gray-500 [&_p]:inline [&_p]:mb-0">
```

PR #2543, opened 2025-07-21
Support repl templates - Optimize template performance for top-level items

sgalsaleh:

Optimizes template performance for top-level items

[SC-127009](https://app.shortcut.com/replicated/story/127009)

Yes

```release-note
NONE
```

NONE


api/pkg/template/config.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	// Template the default field if present
+	if configItem.Default.String() != "" {
+		val, err := e.processTemplate(configItem.Default.String())
+		if err != nil {
+			return nil, fmt.Errorf("template default for %s: %w", name, err)
+		}
+		templatedDefault = val
+	}
```
emosbaugh, line 156: This seems like it doesn't need to run if effectiveValueGetter is not nil

sgalsaleh: Why? The value and default are used when templating individual config items that don't need the effective value.

api/pkg/template/config.go
```diff
@@ -0,0 +1,260 @@
...
+
+// resolveConfigItem processes a config item and returns its resolved values. It determines:
+// 1. The effective value - the final value used in templates
+// 2. The templated value - the templated result of the item's "value" field
+// 3. The templated default - the templated result of the item's "default" field
+//
+// The effective value is determined either by:
+// - The provided effectiveValueGetter function, if specified
+// - Or following priority: user value > config value > config default
+func (e *Engine) resolveConfigItem(name string, effectiveValueGetter func(configItem *kotsv1beta1.ConfigItem) (string, error)) (*ResolvedConfigItem, error) {
```
emosbaugh, line 113: caching like this wont work. e.g. this will fail

```go
func TestEngine_ConfigOptionFilenameAndDataAndValue(t *testing.T) {
	content := "default content"
	contentEncoded := base64.StdEncoding.EncodeToString([]byte(content))

	userContent := "user content"
	userContentEncoded := base64.StdEncoding.EncodeToString([]byte(userContent))

	config := &kotsv1beta1.Config{
		Spec: kotsv1beta1.ConfigSpec{
			Groups: []kotsv1beta1.ConfigGroup{
				{
					Name: "a_file_group",
					Items: []kotsv1beta1.ConfigItem{
						{
							Type:     "file",
							Name:     "a_file",
							Value:    multitype.FromString(contentEncoded),
							Filename: "a_file.txt",
						},
					},
				},
			},
		},
	}

	engine := NewEngine(config)

	err := engine.Parse("{{repl ConfigOptionFilename \"a_file\" }} {{repl ConfigOptionData \"a_file\" }} {{repl ConfigOption \"a_file\" }}")
	require.NoError(t, err)

	// Test with user value - should be user value
	result, err := engine.Execute(types.AppConfigValues{
		"a_file": {Value: userContentEncoded, Filename: "user_file.txt"},
	})
	require.NoError(t, err)
	assert.Equal(t, fmt.Sprintf("user_file.txt %s %s", userContent, userContentEncoded), result)
}
```

PR #2546, opened 2025-07-22
feat(ui): highlight required items

diamonwiggins:

This PR adds visual indicators for required configuration fields that haven't been filled out when users submit the configuration form. Previously, users would only receive a generic API error without clear guidance on which specific fields were missing.

The implementation adds:
- Error messages displaying `"<Config Display Name> is required"` below help text for unfilled required fields
- Red border highlighting on text-based input components (text, password, textarea, select) when required and unfilled
- Automatic focus on the first required field with a validation error, with support for switching tabs if the field is in a non-active tab

[SC-126369](https://app.shortcut.com/replicated/story/126369/if-a-user-didn-t-complete-one-or-more-required-items-make-that-clear-to-the-user-with-an-error-message-for-each)

Yes 

```release-note
NONE
```

NONE


web/src/components/wizard/config/ConfigurationStep.tsx
```diff
@@ -74,9 +82,104 @@ const ConfigurationStep: React.FC<ConfigurationStepProps> = ({ onNext }) => {
...
+        }
+      }
+    }
+    
+    return null;
+  };
+
+  // Helper function to find which group/tab contains a specific field
+  const findGroupForField = (fieldName: string): AppConfigGroup | null => {
+    if (!appConfig?.groups) return null;
```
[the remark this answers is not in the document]
diamonwiggins: should we keep this if its technically possible for Vendors to not provide a `kind: Config` in their app release at all?

web/src/components/wizard/config/ConfigurationStep.tsx
```diff
@@ -74,9 +82,104 @@ const ConfigurationStep: React.FC<ConfigurationStepProps> = ({ onNext }) => {
...
+        fieldElement = document.getElementById(fieldItem.items[0].name);
+      }
+      
+      if (fieldElement) {
+        fieldElement.focus();
+      } else {
+        console.warn(`Could not find DOM element for field: ${fieldItem.name} (type: ${fieldItem.type})`);
+      }
+    }, focusDelay);
+  };
```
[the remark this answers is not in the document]
diamonwiggins: Can follow up with this one in a new story.

api/internal/managers/app/config/config.go
```diff
@@ -102,6 +106,24 @@ func (m *appConfigManager) ValidateConfigValues(configValues types.AppConfigValu
 		}
 	}
 
+	// Set appropriate message based on error types
+	if ve != nil && len(ve.Errors) > 0 {
+		// Check if ALL errors are about required fields
+		allRequired := true
+		for _, err := range ve.Errors {
+			if !strings.Contains(err.Message, "is required") {
```
sgalsaleh, line 114: why error message contains `is required`? i see that this is removed: `ErrConfigItemRequired = errors.New("item is required")`. i would also suggest using errors.Is instead of checking against a hardcoded string.

web/src/components/wizard/config/ConfigurationStep.tsx
```diff
@@ -77,6 +83,103 @@ const ConfigurationStep: React.FC<ConfigurationStepProps> = ({ onNext }) => {
...
+    // Focus the field after a brief delay to ensure DOM updates complete
+    // This is especially important when switching tabs
+    const focusDelay = targetGroup.name !== activeTab ? FOCUS_DELAY_CROSS_TAB_MS : FOCUS_DELAY_SAME_TAB_MS;
+    setTimeout(() => {
+      let fieldElement: HTMLElement | null = null;
+      
+      // Special handling for file inputs - focus the upload button instead of hidden input
+      if (fieldItem.type === 'file') {
+        // For file inputs, focus the upload button using data-testid (follows pattern: file-input-{name}-button)
+        fieldElement = document.querySelector(`[data-testid="file-input-${fieldItem.name}-button"]`) as HTMLElement;
```
sgalsaleh, line 136: we shouldn't use data test ids in the actual component. also, query by id or selector isn't how it's typically done in react, i'd suggest using refs like so:
```js
// Holds refs to each field by name
const itemRefs = useRef<Record<string, HTMLElement | null>>({});

// Assigns a ref to the element dynamically
const setRef = (name: string) => (el: HTMLElement | null) => {
  itemRefs.current[name] = el;
};

// Usage when rendering:
<input
  key={field.name}
  id={field.name}
  ref={setRef(field.name)}
  type={field.type}
/>

// To focus later:
const focusItem = (name: string) => {
  const el = itemRefs.current[name];
  if (el) {
    el.focus();
    el.scrollIntoView({ behavior: "smooth", block: "center" });
  }
};

```

web/src/components/wizard/config/ConfigurationStep.tsx
```diff
@@ -77,6 +83,103 @@ const ConfigurationStep: React.FC<ConfigurationStepProps> = ({ onNext }) => {
...
+        console.warn(`Could not find DOM element for field: ${fieldItem.name} (type: ${fieldItem.type})`);
+      }
+    }, focusDelay);
+  };
+
+  // Helper function to parse server validation errors from API response
+  const parseServerErrors = (error: unknown): Record<string, string> => {
+    const fieldErrors: Record<string, string> = {};
+    
+    // Check if error has structured field errors
```
sgalsaleh, line 169: Why not create a type for the server's validation errors? example:

- https://github.com/replicatedhq/embedded-cluster/blob/5163536446e16847e8bae0e46de8c271b7a4ec2f/web/src/components/wizard/setup/KubernetesSetupStep.tsx#L37-L39
- https://github.com/replicatedhq/embedded-cluster/blob/5163536446e16847e8bae0e46de8c271b7a4ec2f/web/src/components/wizard/setup/KubernetesSetupStep.tsx#L72

web/src/components/wizard/config/ConfigurationStep.tsx
```diff
@@ -127,11 +242,21 @@ const ConfigurationStep: React.FC<ConfigurationStepProps> = ({ onNext }) => {
     return changedValues?.[item.name]?.value ?? (item.value || item.default || '');
   };
 
+
```
sgalsaleh, line 245: extra new line

api/internal/managers/app/config/config.go
```diff
@@ -102,6 +106,24 @@ func (m *appConfigManager) ValidateConfigValues(configValues types.AppConfigValu
 		}
 	}
 
+	// Set appropriate message based on error types
```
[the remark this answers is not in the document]
diamonwiggins: i guess i was thinking that we'd want the error from the API to be more specific in the case it was required fields that were the issue vs. something else, but i guess since we'll show a specific error on each item just keeping the generic validation error makes sense.

api/internal/managers/app/config/config.go
sgalsaleh: Yeah, i think it should be in the FE. Any reason it needs to be in the BE?

diamonwiggins: the feedback i got was to keep all validation in the backend, and i'm not sure if still being able to present this message using the frontend only would count as me doing additional "frontend validation"

web/src/components/wizard/config/ConfigurationStep.tsx
diamonwiggins: ran into some strangeness getting file items to focus correctly. let me know what you think of the implementation

web/src/components/wizard/config/ConfigurationStep.tsx
```diff
@@ -95,16 +165,28 @@ const ConfigurationStep: React.FC<ConfigurationStepProps> = ({ onNext }) => {
           handleUnauthorized(errorData);
           throw new Error('Session expired. Please log in again.');
         }
-        throw new Error(errorData.message || 'Failed to save configuration');
+        // Re-throw with full error data for parsing in onError
```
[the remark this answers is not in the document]
diamonwiggins: I think its this way for the type safety? To make sure we throw a `ConfigError` and not just an error?

sgalsaleh (approved):

PR #2572, opened 2025-07-25
chore(api): add app controller

JGAntunes:

```release-note

```


api/internal/states/install/states.go
```diff
@@ -0,0 +1,42 @@
+package states
```
JGAntunes, line 1: This is what I came up as a way to share the state definitions between linux/kubernetes and app.

api/controllers/app/install/config.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"runtime/debug"
 
+	states "github.com/replicatedhq/embedded-cluster/api/internal/states/install"
```
JGAntunes, line 8: So App controller is doing state transitions but not defining the valid state transitions here (that's still in linux and kubernetes controllers). Personally, while not ideal (leaky interface and we're running under the assumption this definition comes from "somewhere") the alternative of breaking out the valid state transition definition would just add to the confusion of the state transition definitions (we define part of the state transitions in a place and then the rest in another). Curious to hear your thoughts?

- https://github.com/replicatedhq/embedded-cluster/blob/c5c2d873476789e44b6899523c4f51d51430d1ef/api/controllers/kubernetes/install/statemachine.go#L9-L21
- https://github.com/replicatedhq/embedded-cluster/blob/c5c2d873476789e44b6899523c4f51d51430d1ef/api/controllers/linux/install/statemachine.go#L10-L30

api/integration/app/install/config_test.go
```diff
@@ -0,0 +1,741 @@
...
+					linuxinstall.WithConfigValues(configValues),
+				)
+				require.NoError(t, err)
+				// Create the API with the install controller
+				return integration.NewAPIWithReleaseData(t,
+					api.WithLinuxInstallController(controller),
+					api.WithAuthController(auth.NewStaticAuthController("TOKEN")),
+					api.WithLogger(logger.NewDiscardLogger()),
+				)
+			},
```
JGAntunes, line 707: While looking for a way to create the API instance on a per test scenario I was thinking that it would be helpful to actually consolidate the Controller options on a single shared struct for all the controllers. This way we could be passing over generic Controller options on each test that could be applied here without much fuss. Opted not to go down that route just yet since this does the trick and we avoid yet another set of large changes around the code.

api/integration/kubernetes/install/app_test.go
```diff
@@ -14,387 +12,16 @@ import (
...
-		// Parse the response body
-		var apiError types.APIError
-		err = json.NewDecoder(rec.Body).Decode(&apiError)
-		require.NoError(t, err)
-		assert.Equal(t, http.StatusUnauthorized, apiError.StatusCode)
-	})
-}
-
 // TestInstallController_PatchAppConfigValuesWithAPIClient tests the PatchAppConfigValues endpoint using the API client
 func TestInstallController_PatchAppConfigValuesWithAPIClient(t *testing.T) {
```
JGAntunes, line 26: So, the test suite works well for our API tests, but things get more complicated with the API client ones and I hadn't considered that.

I don't think there's an easy/pretty way out this that doesn't involve passing around function parameters or instead creating an interface for the client (or changing the api for the client altogether). So, I'm curious for:
- Your thoughts?
- What do you think about doing these as follow ups?

api/controllers/app/install/controller.go
```diff
@@ -0,0 +1,79 @@
...
+	TemplateAppConfig(ctx context.Context, values types.AppConfigValues, maskPasswords bool) (types.AppConfig, error)
+	PatchAppConfigValues(ctx context.Context, values types.AppConfigValues) error
+	GetAppConfigValues(ctx context.Context) (types.AppConfigValues, error)
+}
+
+var _ Controller = (*InstallController)(nil)
+
+type InstallController struct {
+	appConfigManager appconfig.AppConfigManager
+	configValues     types.AppConfigValues
```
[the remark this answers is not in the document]
JGAntunes: Good catch, removed via - https://github.com/replicatedhq/embedded-cluster/pull/2572/commits/14b647d148b9a0389d62147a2d2237d34569dfe3

sgalsaleh (approved):

api/controllers/app/app_test.go
```diff
@@ -0,0 +1,44 @@
+package app
```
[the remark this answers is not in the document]
sgalsaleh: i think what you have is fine

api/controllers/app/install/config.go
sgalsaleh: i think defining valid state transitions in the linux/kubernetes controller is the right approach, as that's where the workflow lives

api/integration/kubernetes/install/app_test.go
sgalsaleh: yeah, what you have in this pr is a great improvement, we can revisit these if they become a burden

PR #2594, opened 2025-07-29
Do not fail image scan workflows until we fully utilize SecureBuild images

sgalsaleh:

Right now, our CI builds on main are all red, but we release in spite of that because there errors are "expected" and should go away when we utilize fully SecureBuild images.

Having a red pipeline all the time makes it hard to determine if it's okay to release or not, and messes up with other reporting metrics around our CI.

This PR temporarily makes it so that our CI doesn't fail for image scans.

NONE

```release-note
NONE
```

NONE


dydx (approved):
Seems reasonable

PR #2595, opened 2025-07-29
Add logic to run app preflight checks from a preflight spec

sgalsaleh:

- Refactors `pkg-new/preflights` package to allow running app preflight checks
- Adds the ability to run app preflight checks from a preflight spec

[SC-127240](https://app.shortcut.com/replicated/story/127240)

Yes

```release-note
NONE
```

NONE


api/types/preflight.go
```diff
@@ -6,30 +6,38 @@ type PostInstallRunHostPreflightsRequest struct {
...
 // HostPreflights represents the host preflight checks state
 type HostPreflights struct {
-	Titles                    []string              `json:"titles"`
-	Output                    *HostPreflightsOutput `json:"output"`
-	Status                    Status                `json:"status"`
-	AllowIgnoreHostPreflights bool                  `json:"allowIgnoreHostPreflights"`
+	Titles                    []string          `json:"titles"`
+	Output                    *PreflightsOutput `json:"output"`
+	Status                    Status            `json:"status"`
+	AllowIgnoreHostPreflights bool              `json:"allowIgnoreHostPreflights"`
```
JGAntunes, line 12: Am I reading correctly that this is the only property that distinguishes `HostPreflights` from `AppPreflights`? I wonder if we really need this here or if there's an alternative method, feels a bit off compared to the rest of the data.

pkg-new/preflights/run.go
```diff
@@ -15,30 +15,53 @@ import (
...
-	"github.com/replicatedhq/embedded-cluster/pkg/runtimeconfig"
 	troubleshootv1beta2 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
 	"sigs.k8s.io/yaml"
 )
 
-// Run runs the provided host preflight spec locally. This function is meant to be
-// used when upgrading a local node.
-func (p *PreflightsRunner) Run(ctx context.Context, spec *troubleshootv1beta2.HostPreflightSpec, rc runtimeconfig.RuntimeConfig) (*apitypes.HostPreflightsOutput, string, error) {
+// RunHostPreflights runs the provided host preflight spec locally.
+func (p *PreflightsRunner) RunHostPreflights(ctx context.Context, spec *troubleshootv1beta2.HostPreflightSpec, opts RunOptions) (*apitypes.PreflightsOutput, string, error) {
```
JGAntunes, line 23: I think the methods in this file are largely untested right?

We should consider setting up some `os/exec` `Cmd` interface - https://pkg.go.dev/os/exec#Cmd - we can mock in tests and test the remaining logic around the `run` methods.

JGAntunes (commented):
Some minor comments + the conflicts but overall lgtm

pkg-new/preflights/run.go
sgalsaleh: agreed. ok if it's a separate pr though? feels like something this package is lacking prior to this pr.

sgalsaleh:
bugbot run

sgalsaleh:
bugbot run

PR #2609, opened 2025-07-31
feat: skip KOTS app preflights when manager experience installer is enabled

screspod:

Skips running KOTS app preflights when the manager experience installer is enabled 

https://app.shortcut.com/replicated/story/127247/pass-the-skip-preflights-flag-to-kots-install

NONE

```release-note
NONE
```

NONE

### DIFF
--- a/api/internal/managers/linux/infra/install.go
+++ b/api/internal/managers/linux/infra/install.go
@@ -29,7 +29,12 @@ import (
 	kyaml "sigs.k8s.io/yaml"
 )
 
-const K0sComponentName = "Runtime"
+const (
+	// Skip running the KOTS app preflights in the Admin Console; we run them in the manager experience installer when ENABLE_V3 is enabled
+	SkipPreflights = true
+
+	K0sComponentName = "Runtime"
+)
 
 func AlreadyInstalledError() error {
 	//nolint:staticcheck // ST1005 TODO: use a constant here and print a better error message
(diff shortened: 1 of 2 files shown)


api/internal/managers/linux/infra/install.go
```diff
@@ -314,6 +319,7 @@ func (m *infraManager) getAddonInstallOpts(license *kotsv1beta1.License, rc runt
 			Namespace:             constants.KotsadmNamespace,
 			ClusterID:             m.clusterID,
 			AirgapBundle:          m.airgapBundle,
+			SkipPreflights:        SkipPreflights,
```
sgalsaleh, line 322: isn't v3 an env var? why not just
```suggestion
			SkipPreflights:        os.Getenv("ENABLE_V3") == "1",
```

api/internal/managers/linux/infra/install.go
```diff
@@ -29,7 +29,12 @@ import (
 	kyaml "sigs.k8s.io/yaml"
 )
 
-const K0sComponentName = "Runtime"
+const (
+	// Skip running the KOTS app preflights in the Admin Console; we run them in the manager experience installer when ENABLE_V3 is enabled
+	SkipPreflights = true
```
sgalsaleh, line 34: nit: no need to define a global var, just set it to true in the options and add a comment there

sgalsaleh (approved):

PR #2610, opened 2025-08-01
Move app installation logic out of infrastructure setup

sgalsaleh:

- Moves app installation logic out of infrastructure setup but does not decouple the actual app installation from the infra installation yet. That will be done in a later PR.
- Moves app specific managers initialization to the app install controller.

[SC-127428](https://app.shortcut.com/replicated/story/127428)

Yes

```release-note
NONE
```

NONE

### DIFF
--- a/api/controllers/app/install/apppreflight.go
+++ b/api/controllers/app/install/apppreflight.go
@@ -44,6 +44,9 @@ func (c *InstallController) RunAppPreflights(ctx context.Context, opts RunAppPre
 	if err != nil {
 		return fmt.Errorf("extract app preflight spec: %w", err)
 	}
+	if appPreflightSpec == nil {
+		return fmt.Errorf("no app preflight spec found")
+	}
 
 	err = c.stateMachine.Transition(lock, states.StateAppPreflightsRunning)
 	if err != nil {
--- a/api/controllers/app/install/controller.go
+++ b/api/controllers/app/install/controller.go
@@ -194,5 +194,11 @@ func (c *InstallController) validateInit() error {
 	if c.stateMachine == nil {
 		return errors.New("stateMachine is required for App Install Controller")
 	}
+	if c.store == nil {
+		return errors.New("store is required for App Install Controller")
+	}
+	if c.releaseData == nil {
+		return errors.New("releaseData is required for App Install Controller")
+	}
 	return nil
 }
--- a/api/controllers/app/install/controller.go
+++ b/api/controllers/app/install/controller.go
@@ -41,6 +41,8 @@ type InstallController struct {
 	releaseData         *release.ReleaseData
 	store               store.Store
 	configValues        types.AppConfigValues
+	clusterID           string
+	airgapBundle        string
 }
 
 type InstallControllerOption func(*InstallController)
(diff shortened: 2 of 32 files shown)


api/internal/states/install/states.go
```diff
@@ -37,6 +37,8 @@ const (
 	StateInfrastructureInstalling statemachine.State = "InfrastructureInstalling"
 	// StateInfrastructureInstallFailed is a final state of the install process when the infrastructure failed to isntall
 	StateInfrastructureInstallFailed statemachine.State = "InfrastructureInstallFailed"
+	// StateInfrastructureInstallSucceeded is the state of the install process when the infrastructure install has succeeded
+	StateInfrastructureInstallSucceeded statemachine.State = "InfrastructureInstallSucceeded"
```
emosbaugh, line 41: you need to add these to the valid state transitions object in https://github.com/replicatedhq/embedded-cluster/blob/main/api/controllers/kubernetes/install/statemachine.go#L9 and linux

emosbaugh: or perhaps this is a TODO

api/controllers/app/install/controller.go
```diff
@@ -78,21 +118,64 @@ func NewInstallController(opts ...InstallControllerOption) (*InstallController,
 		return nil, err
 	}
 
+	if controller.appConfigManager == nil {
```
emosbaugh, line 135: why initialize these if they are nil? how come we werent before? would it be better to return an error?

api/internal/managers/app/install/install.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	configValuesFile, err := os.CreateTemp("", "config-values*.yaml")
+	if err != nil {
+		return "", fmt.Errorf("create temp file: %w", err)
+	}
+	defer configValuesFile.Close()
+
+	if _, err := configValuesFile.Write(data); err != nil {
+		os.Remove(configValuesFile.Name())
```
emosbaugh, line 64: im a believer in explicitly ignoring the error if that is the intention so it doesnt go unhandled unless it is intentional

```suggestion
		_ = os.Remove(configValuesFile.Name())
```

api/controllers/app/install/controller.go
emosbaugh: ok

cmd/installer/goods/materializer.go
```diff
@@ -30,11 +30,15 @@ func NewMaterializer(rc runtimeconfig.RuntimeConfig) *Materializer {
 	}
 }
 
+func (m *Materializer) InternalBinary(name string) (string, error) {
```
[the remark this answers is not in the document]
sgalsaleh: it is still used elsewhere. i didn't wanna refactor the old code. also helps avoid running the legacy e2e tests :)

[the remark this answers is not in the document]
emosbaugh: i dont think it is. i was able to remove it and still build

api/controllers/app/install/apppreflight.go
```diff
@@ -44,6 +44,9 @@ func (c *InstallController) RunAppPreflights(ctx context.Context, opts RunAppPre
 	if err != nil {
 		return fmt.Errorf("extract app preflight spec: %w", err)
 	}
+	if appPreflightSpec == nil {
```
emosbaugh, line 47: Is a preflight spec required? I just saw this support case

https://github.com/replicated-collab/rockwell-automation-replicated/issues/46

emosbaugh (approved):

PR #2612, opened 2025-08-01
fix(ci): pipe grep exits with 141 no reader

emosbaugh:

Failing test is a bug as grep -q exits immediately after first match and kernel returns 141 error as the command is still writing to the pipe and there is no reader

https://stackoverflow.com/questions/19120263/why-exit-code-141-with-grep-q/19120674#19120674

https://github.com/replicatedhq/embedded-cluster/actions/runs/16650744879/job/47123777677#step:4:236

```release-note

```


diamonwiggins (approved):
lgtm

PR #2619, opened 2025-08-02
tests: add test that reports network events while deploying an EC app

mikhailswift:

Adds a test that uses CMX network reports to verify that all requests go to expected destinations during an app install.

The first iteration of this just prints out the IPs reached and any domains used in the connection if available. Follow ups will assert that the connections made are limited to the expected destinations for the EC install process.

https://app.shortcut.com/replicated/story/127479/setup-testing-for-network-reporting-in-ec

NONE

```release-note
NONE
```

NONE

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -904,6 +904,7 @@ jobs:
   e2e:
     name: E2E # this name is used by .github/workflows/automated-prs-manager.yaml
     runs-on: ${{ matrix.runner || 'ubuntu-22.04' }}
+    continue-on-error: ${{ matrix.continue-on-error || false }}
     needs:
       - build-current
       - build-previous-k0s
@@ -926,7 +927,6 @@ jobs:
           - TestMultiNodeAirgapHAInstallation
           - TestSingleNodeAirgapDisasterRecovery
           - TestMultiNodeAirgapHADisasterRecovery
-          - TestSingleNodeNetworkReport
         include:
           - test: TestVersion
             is-lxd: true
@@ -940,6 +940,8 @@ jobs:
             is-lxd: true
           - test: TestInstallWithMITMProxy
             is-lxd: true
+          - test: TestSingleNodeNetworkReport
+            continue-on-error: true
     steps:
       - name: Checkout
         uses: actions/checkout@v5
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -926,6 +926,7 @@ jobs:
           - TestMultiNodeAirgapHAInstallation
           - TestSingleNodeAirgapDisasterRecovery
           - TestMultiNodeAirgapHADisasterRecovery
+          - TestSingleNodeNetworkReport
         include:
           - test: TestVersion
             is-lxd: true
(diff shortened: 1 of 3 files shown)


xavpaice:
bugbot run

xavpaice:
This does output every line of the report, which isn't particularly helpful.  It's not too much of an issue, but fills the logs up a bunch.  Is there a consideration to just capture the unique destination IPs for processing later on?  I guess the process/pod name calling would be helpful too for later troubleshooting.

xavpaice:
The interesting part of the test output so far:

```
  install_test.go:2054: Logged outbound external network accesses:
        
    install_test.go:2061: IP: 62.80.139.68
    install_test.go:2061: IP: 162.159.138.43
    install_test.go:2061: IP: 104.21.18.220
    install_test.go:2063: 
        	- kots.io
    install_test.go:2061: IP: 185.199.109.133
    install_test.go:2063: 
        	- release-assets.githubusercontent.com
    install_test.go:2061: IP: 140.82.121.4
    install_test.go:2063: 
        	- release-assets.githubusercontent.com
    install_test.go:2061: IP: 162.159.134.41
    install_test.go:2061: IP: 162.159.137.43
    install_test.go:2061: IP: ff02::2
    install_test.go:2061: IP: 162.159.133.41
    install_test.go:2063: 
        	- ec-e2e-replicated-app.testcluster.net
        	- 10.0.0.10
    install_test.go:2061: IP: 185.125.190.56
    install_test.go:2061: IP: 37.27.215.85
    install_test.go:2061: IP: 95.217.220.251
    install_test.go:2061: IP: 91.189.91.157
    install_test.go:2061: IP: 185.125.190.57
    install_test.go:2061: IP: 37.27.202.175
    install_test.go:2061: IP: 185.125.190.58
```

This would be great for comparing to a known list, which is the next step for this work.  I'd like to get this PR merged into the repo so we can prove out that the additional test, plus reporting, isn't exploding the EC CI processes, and also so that we have some additional real-world usage of the feature to allow the CMX folks to determine what to about storing the reports.

.github/workflows/ci.yaml
```diff
@@ -926,6 +926,7 @@ jobs:
           - TestMultiNodeAirgapHAInstallation
           - TestSingleNodeAirgapDisasterRecovery
           - TestMultiNodeAirgapHADisasterRecovery
+          - TestSingleNodeNetworkReport
```
sgalsaleh, line 929: why not integrate to all existing tests instead of adding a new test for this? different tests can behave differently and would be valuable to have this for all of them

mikhailswift: Good point, no reason we couldn't do that instead. Happy to make that change!

sgalsaleh: Every test in the suite of tests we have will block merging the PR if it fails though. Would there be future work to integrate it into the other tests and replace this one?

sgalsaleh (approved):

.github/workflows/ci.yaml
mikhailswift: Yeah, the intent is that as network reporting gets closer to GA we would expand the usage of it in testing

PR #2672, opened 2025-08-14
feat(v3): add support for local registry templates

diamonwiggins:

```release-note

```


api/pkg/template/execute.go
```diff
@@ -187,3 +192,81 @@ func (e *Engine) recordDependency(dependency string) {
...
+	err := e.kubeClient.Get(e.ctx, client.ObjectKey{
+		Namespace: constants.RegistryNamespace,
+		Name:      "registry",
+	}, svc)
+
+	if err != nil || svc.Spec.ClusterIP == "" {
+		return ""
+	}
+
+	return fmt.Sprintf("%s:5000", svc.Spec.ClusterIP)
```
emosbaugh, line 229: I think all the registry values are deterministic and you should not need a kube client and only a service cidr. see the [GetRegistryClusterIP](https://github.com/replicatedhq/embedded-cluster/blob/df68a5584cd6fea844e4e3eacbc5080ff79f6d45/pkg/addons/registry/registry.go#L59) function

emosbaugh (changes requested):

api/controllers/app/install/controller.go
```diff
@@ -164,10 +180,21 @@ func NewInstallController(opts ...InstallControllerOption) (*InstallController,
 	}
 
 	if controller.appReleaseManager == nil {
-		appReleaseManager, err := appreleasemanager.NewAppReleaseManager(
-			*controller.releaseData.AppConfig,
+		var appReleaseManagerOpts []appreleasemanager.AppReleaseManagerOption
+		appReleaseManagerOpts = append(appReleaseManagerOpts,
 			appreleasemanager.WithLogger(controller.logger),
 			appreleasemanager.WithReleaseData(controller.releaseData),
+			appreleasemanager.WithRuntimeConfig(controller.runtimeConfig),
```
sgalsaleh, line 187: app controller should not be aware of runtimeconfig as that's only available in linux targets, not kubernetes targets, and the app controller should be agnostic of that

diamonwiggins: yeah, pretty sure that can be removed now, and its not used anymore. i'll do that real quick.

diamonwiggins: done

api/controllers/linux/install/controller.go
```diff
@@ -263,6 +266,20 @@ func NewInstallController(opts ...InstallControllerOption) (*InstallController,
...
+
+		// Detect registry settings and pass to app controller
+		var license *kotsv1beta1.License
+		if controller.license != nil {
+			// Parse license from bytes
+			var parsedLicense kotsv1beta1.License
+			if err := yaml.Unmarshal(controller.license, &parsedLicense); err == nil {
+				license = &parsedLicense
+			}
+		}
```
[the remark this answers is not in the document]
emosbaugh: this should return an error

api/controllers/linux/install/registry.go
```diff
@@ -0,0 +1,45 @@
...
+	// Check if this is an airgap install
+	isAirgap := c.airgapBundle != ""
+	if !isAirgap {
+		return settings // Return empty settings for online installs
+	}
+
+	settings.HasLocalRegistry = true
+
+	// Get registry host using deterministic function
+	if c.rc != nil {
```
emosbaugh, line 24: return an error here. rc should not be nil

api/controllers/app/install/controller.go
```diff
@@ -164,11 +171,21 @@ func NewInstallController(opts ...InstallControllerOption) (*InstallController,
...
-		appReleaseManager, err := appreleasemanager.NewAppReleaseManager(
-			*controller.releaseData.AppConfig,
+		var appReleaseManagerOpts []appreleasemanager.AppReleaseManagerOption
+		appReleaseManagerOpts = append(appReleaseManagerOpts,
 			appreleasemanager.WithLogger(controller.logger),
 			appreleasemanager.WithReleaseData(controller.releaseData),
 		)
+
+		// Add registry settings if available
+		if controller.registrySettings != nil {
```
emosbaugh, line 181: appReleaseManagerOpts and this nil check is probably unnecessary as WithRegistrySettings accepts a pointer

api/internal/managers/app/release/manager.go
```diff
@@ -72,11 +80,18 @@ func NewAppReleaseManager(config kotsv1beta1.Config, opts ...AppReleaseManagerOp
...
-		manager.templateEngine = template.NewEngine(
-			&manager.rawConfig,
+		var templateOpts []template.EngineOption
+		templateOpts = append(templateOpts, 
 			template.WithLicense(manager.license),
 			template.WithReleaseData(manager.releaseData),
 		)
+
+		// Add registry settings if available
+		if manager.registrySettings != nil {
```
emosbaugh, line 90: this nil check is unnecessary

api/types/registry.go
```diff
@@ -0,0 +1,10 @@
+package types
+
+// RegistrySettings encapsulates all registry-related configuration
+type RegistrySettings struct {
+	HasLocalRegistry    bool   // whether a local registry is available
```
emosbaugh, line 5: since this type is exported can you move these comments above the line so they are in the standard godoc format?

```suggestion
	// HasLocalRegistry is true when a local registry is available
	HasLocalRegistry    bool
```

api/controllers/linux/install/controller.go
```diff
@@ -265,6 +268,24 @@ func NewInstallController(opts ...InstallControllerOption) (*InstallController,
...
+		var license *kotsv1beta1.License
+		if controller.license != nil {
+			// Parse license from bytes
+			var parsedLicense kotsv1beta1.License
+			if err := yaml.Unmarshal(controller.license, &parsedLicense); err != nil {
+				return nil, fmt.Errorf("failed to parse license data: %w", err)
+			}
+			license = &parsedLicense
+		}
+		registrySettings, err := controller.detectRegistrySettings(license)
```
emosbaugh, line 282: license cannot be nil or this will panic. see the comment from cursor

emosbaugh (changes requested):

emosbaugh (approved):

api/controllers/linux/install/controller.go
sgalsaleh, line 282: this happens when the api starts, which is before the user goes through the setup page, so if the user changes the cidr then the registry settings would have the wrong values

api/controllers/linux/install/controller.go
```diff
@@ -265,6 +268,24 @@ func NewInstallController(opts ...InstallControllerOption) (*InstallController,
 			appcontroller.WithAirgapBundle(controller.airgapBundle),
 			appcontroller.WithPrivateCACertConfigMapName(adminconsole.PrivateCASConfigMapName), // Linux installations use the ConfigMap
 		)
+
+		// Detect registry settings and pass to app controller
+		var license *kotsv1beta1.License
+		if controller.license != nil {
+			// Parse license from bytes
```
sgalsaleh, line 275: nit: i think there's a helper method to parse the license which can make this a bit cleaner.

PR #2712, opened 2025-08-21
feat(v3): add support for local registry templates 

diamonwiggins:

```release-note

```

### DIFF
--- a/api/internal/handlers/linux/install.go
+++ b/api/internal/handlers/linux/install.go
@@ -1,12 +1,16 @@
 package linux
 
 import (
+	"context"
+	"encoding/base64"
+	"fmt"
 	"net/http"
 
 	appinstall "github.com/replicatedhq/embedded-cluster/api/controllers/app/install"
 	"github.com/replicatedhq/embedded-cluster/api/controllers/linux/install"
 	"github.com/replicatedhq/embedded-cluster/api/internal/handlers/utils"
 	"github.com/replicatedhq/embedded-cluster/api/types"
+	"github.com/replicatedhq/embedded-cluster/pkg/addons/registry"
 )
 
 // GetInstallationConfig handler to get the installation config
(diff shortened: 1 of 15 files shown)


api/internal/handlers/linux/install.go
```diff
@@ -403,3 +415,44 @@ func (h *Handler) GetAppInstallStatus(w http.ResponseWriter, r *http.Request) {
 
 	utils.JSON(w, r, http.StatusOK, appInstall, h.logger)
 }
+
+// calculateRegistrySettings calculates registry settings for airgap installations
+func (h *Handler) calculateRegistrySettings(ctx context.Context) (*types.RegistrySettings, error) {
```
sgalsaleh, line 420: the expectation is that the handlers package should be a light wrapper to the business logic. this logic has high potential of being re-used somewhere else as well. how about moving this to the linux install manager and expose it via the linux install controller and then use it here?

api/internal/managers/app/release/template.go
```diff
@@ -81,7 +81,13 @@ func (m *appReleaseManager) templateHelmChartCRs(configValues types.AppConfigVal
 		}
 
 		// Execute the template with config values
-		templatedYAML, err := m.templateEngine.Execute(configValues, template.WithProxySpec(proxySpec))
+		execOptions := []template.ExecOption{
+			template.WithProxySpec(proxySpec),
+		}
+		if registrySettings != nil {
```
sgalsaleh, line 87: nit: why does this need an if statement? i think we should be consistent with the proxySpec above. it would still work the same if it's nil

api/internal/managers/linux/installation/config.go
```diff
@@ -233,3 +236,62 @@ func (m *installationManager) ConfigureHost(ctx context.Context, rc runtimeconfi
...
+	// Get the current config to access service CIDR
+	config, err := m.GetConfig()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get installation config: %w", err)
+	}
+
+	serviceCIDR := config.ServiceCIDR
+	if serviceCIDR == "" {
+		// Fallback to runtime config if not set in installation config
+		rc, err := runtimeconfig.NewFromDisk()
```
sgalsaleh, line 256: runtimeconfig should be passed as an argument to the function similar to `SetConfigDefaults` and `ConfigureHost`

api/internal/managers/linux/installation/config.go
```diff
@@ -233,3 +236,62 @@ func (m *installationManager) ConfigureHost(ctx context.Context, rc runtimeconfi
 
 	return nil
 }
+
+// CalculateRegistrySettings calculates registry settings for airgap installations
+func (m *installationManager) CalculateRegistrySettings(ctx context.Context, releaseData *release.ReleaseData) (*types.RegistrySettings, error) {
```
sgalsaleh, line 241: release data should be passed as an option to the manager similar to all other managers since it's not mutable

api/internal/managers/linux/installation/config.go
```diff
@@ -233,3 +235,55 @@ func (m *installationManager) ConfigureHost(ctx context.Context, rc runtimeconfi
...
 }
+
+// CalculateRegistrySettings calculates registry settings for airgap installations
+func (m *installationManager) CalculateRegistrySettings(ctx context.Context, rc runtimeconfig.RuntimeConfig) (*types.RegistrySettings, error) {
+	// Only return registry settings for airgap installations
+	if m.airgapBundle == "" {
+		return nil, nil
+	}
+
+	// Get the current config to access service CIDR
```
sgalsaleh, line 246: looking at the codebase, i think you should only use the runtimeconfig. the installation config is only used for the setup form and accepting values from the user, which then gets used to calculate and update the runtimeconfig. the runtimeconfig is the source of truth

sgalsaleh (approved):

PR #2751, opened 2025-08-26
fix(api): use correct helm client k8s version

emosbaugh:

App release manager needs to use a helm client with the k8s version so that it can properly render templates using KubeVersion helm built-in.

Additionally, app install manager is using the incorrect k8s version when instantiating the helm client for target=kubernetes. It is using the version from the build when it should be using the version discovered from the cluster. This makes it so the version is always retrieved from the cluster.

```release-note

```

### DIFF
--- a/api/controllers/app/install/controller.go
+++ b/api/controllers/app/install/controller.go
@@ -48,6 +48,7 @@ type InstallController struct {
 	clusterID                  string
 	airgapBundle               string
 	privateCACertConfigMapName string
+	k8sVersion                 string
 	restClientGetter           genericclioptions.RESTClientGetter
 	kubeConfigPath             string
 }
@@ -132,6 +133,12 @@ func WithPrivateCACertConfigMapName(configMapName string) InstallControllerOptio
 	}
 }
 
+func WithK8sVersion(k8sVersion string) InstallControllerOption {
+	return func(c *InstallController) {
+		c.k8sVersion = k8sVersion
+	}
+}
+
 func WithRESTClientGetter(restClientGetter genericclioptions.RESTClientGetter) InstallControllerOption {
 	return func(c *InstallController) {
 		c.restClientGetter = restClientGetter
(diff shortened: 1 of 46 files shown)


api/internal/managers/app/install/util.go
```diff
@@ -32,10 +32,15 @@ func (m *appInstallManager) setupHelmClient() error {
 		return nil
 	}
 
+	k8sVersion, err := m.getK8sVersion()
```
sgalsaleh, line 35: edge case: this logic can break later on upgrades if we start to deploy the app first before upgrading infra. for now it's fine

api/internal/managers/app/release/template.go
```diff
@@ -215,14 +209,55 @@ func (m *appReleaseManager) dryRunHelmChart(ctx context.Context, templatedCR *ko
...
+
+	manifests, err := m.hcli.Render(ctx, installOpts)
 	if err != nil {
 		return nil, fmt.Errorf("render helm chart %s: %w", templatedCR.Name, err)
 	}
 
 	return manifests, nil
 }
 
+func (m *appReleaseManager) setupHelmClient() error {
```
sgalsaleh, line 224: i think we should break this template.go file down later on, it feels overloaded now

sgalsaleh (approved):

api/internal/managers/app/install/util.go
emosbaugh: why would we install the app before upgrading the infra? is this how it works today? i could see running the app preflights before installing the infra.

api/controllers/app/install/controller.go
```diff
@@ -205,6 +212,7 @@ func NewInstallController(opts ...InstallControllerOption) (*InstallController,
 			appreleasemanager.WithReleaseData(controller.releaseData),
 			appreleasemanager.WithLicense(license),
 			appreleasemanager.WithPrivateCACertConfigMapName(controller.privateCACertConfigMapName),
+			appreleasemanager.WithK8sVersion(controller.k8sVersion),
```
emosbaugh, line 215: this is only passing the version to the release manager, not the install manager. is that correct?

api/internal/managers/app/install/util.go
sgalsaleh: Ah preflights is a good point. No that's not how it works today, we deploy infra first today

api/controllers/app/install/controller.go
emosbaugh: updated so they both use it

sgalsaleh (approved):

PR #2770, opened 2025-08-28
Use Helm binary instead of the Go SDK to manage charts

sgalsaleh:

Uses the Helm binary instead of the Go SDK to manage Helm charts

[SC-128058](https://app.shortcut.com/replicated/story/128058)

Yes

```release-note
NONE
```

NONE


api/controllers/kubernetes/install/controller_test.go
```diff
@@ -90,6 +90,8 @@ func TestGetInstallationConfig(t *testing.T) {
 		},
 	}
 
+	t.Setenv("HELM_BINARY_PATH", "helm") // use the helm binary in PATH
```
emosbaugh, line 93: Does this mean i need helm install locally to run unit tests?

pkg/helm/client.go
```diff
@@ -76,36 +41,44 @@ func newClient(opts HelmOptions) (*HelmClient, error) {
...
+func getHelmPath(opts HelmOptions) (string, error) {
+	if opts.HelmPath != "" {
+		// used in operations that do not have the helm binary embedded (i.e. upgrade via operator, etc..)
+		return opts.HelmPath, nil
+	}
+	if hp := os.Getenv("HELM_BINARY_PATH"); hp != "" {
+		// used in tests so that we don't pass the helm binary path all over the place
+		return hp, nil
+	}
+	hp, err := goods.Binary("helm")
```
emosbaugh, line 65: would it be possible to pass around a HelmClient rather than an EnvSettings so you dont have to use the goods package outside of the cli? Looks like you would have to factor out logging into the call to the client functions.

pkg/helm/client.go
```diff
@@ -258,296 +208,439 @@ func (h *HelmClient) PullByRefWithRetries(ctx context.Context, ref string, versi
...
+// addKubernetesEnvArgs adds kubernetes environment arguments to the helm command
+func (h *HelmClient) addKubernetesEnvArgs(args []string) []string {
+	if h.kubernetesEnvSettings == nil {
+		return args
 	}
 
-	chartRequested, err := loader.Load(localPath)
-	if err != nil {
-		return nil, fmt.Errorf("load: %w", err)
+	// Add all helm CLI flags from kubernetesEnvSettings
```
emosbaugh, line 607: could you move addKubernetesCLIFlags to this package and export it or make this dynamic?

pkg/helm/client.go
```diff
@@ -258,296 +208,439 @@ func (h *HelmClient) PullByRefWithRetries(ctx context.Context, ref string, versi
...
 }
 
-func (h *HelmClient) Upgrade(ctx context.Context, opts UpgradeOptions) (*release.Release, error) {
-	cfg, err := h.getActionCfg(opts.Namespace)
+// resolveChartPath handles chart source resolution for install, upgrade, and render operations
+func (h *HelmClient) resolveChartPath(ctx context.Context, releaseName, chartPath, chartVersion string) (string, error) {
+	if h.airgapPath != "" {
+		// Use chart from airgap path
+		return filepath.Join(h.airgapPath, fmt.Sprintf("%s-%s.tgz", releaseName, chartVersion)), nil
+	}
```
emosbaugh, line 414: this never felt like it belonged in this package. perhaps something to address later

cmd/installer/cli/install.go
```diff
@@ -590,7 +588,7 @@ func preRunInstallKubernetes(_ *cobra.Command, flags *InstallCmdFlags, _ kuberne
 		}
 	}
 
-	restConfig, err := flags.kubernetesEnvSettings.RESTClientGetter().ToRESTConfig()
+	restConfig, err := clientcmd.BuildConfigFromFlags("", flags.kubernetesEnvSettings.KubeConfig)
```
emosbaugh, line 591: why change this? is it equivalent? what if i dont explicitly set the kubeconfig as a flag?

pkg/helm/client.go
```diff
@@ -76,36 +41,44 @@ func newClient(opts HelmOptions) (*HelmClient, error) {
...
 type HelmOptions struct {
-	KubeConfig       string
-	RESTClientGetter genericclioptions.RESTClientGetter
-	K8sVersion       string
-	AirgapPath       string
-	Writer           io.Writer
-	LogFn            action.DebugLog
+	HelmPath              string
+	KubernetesEnvSettings *helmcli.EnvSettings
+	K8sVersion            string
```
emosbaugh, line 75: is K8sVersion no longer used? Is this going to be an issue for things like buildtools or rendering before install k0s? If not can you remove it?

api/controllers/kubernetes/install/controller_test.go
sgalsaleh: yes, any other ideas? helm binary is currently a requirement for the dev env anyways.

cmd/installer/cli/install.go
sgalsaleh: good catch

pkg/helm/client.go
sgalsaleh: it is used

pkg/helm/client.go
```diff
@@ -76,36 +41,44 @@ func newClient(opts HelmOptions) (*HelmClient, error) {
...
+	if hp := os.Getenv("HELM_BINARY_PATH"); hp != "" {
+		// used in tests so that we don't pass the helm binary path all over the place
+		return hp, nil
+	}
+	hp, err := goods.Binary("helm")
+	if err != nil {
+		return "", fmt.Errorf("get embedded helm binary: %w", err)
+	}
+	return hp, nil
+}
```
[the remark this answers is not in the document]
sgalsaleh: it does get removed on Close

pkg/helm/client.go
sgalsaleh: hmm this will make it hard to override the log function for the helm client and other operation specific fields.

sgalsaleh: what if i pass the helm binary path explicitly from callers instead of using the goods package?

pkg/helm/client.go
sgalsaleh: can you elaborate? what do you mean by making it dynamic?

pkg/helm/client.go
emosbaugh: If you didnt want to add the log as an argument to all methods, could you add the log to the context? Similar to this https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/log

api/controllers/kubernetes/install/controller_test.go
emosbaugh: if you were to instantiate only one client as suggested you would not have to set the env var

emosbaugh (approved):

emosbaugh (approved):

PR #2798, opened 2025-09-03
feat(web): setup page should not pre-fill default values

JGAntunes:
This is the UI follow up to - https://github.com/replicatedhq/embedded-cluster/pull/2788

**NOTE:** This PR is based against - https://github.com/replicatedhq/embedded-cluster/pull/2788 - to make review easy while we wait for that to merged.

Loom:
- https://www.loom.com/share/56937cad67cb4f1284b5851fc937a1d2?sid=56962e79-5b01-41f3-9a63-b6808d019a4b

Currently we cannot go back from the host preflight execution so had to refresh and deliberately fail the preflights so that I could show the behaviour and UI changes.

https://app.shortcut.com/replicated/story/127692/setup-page-should-not-pre-fill-default-values-in-the-form

Yes

```release-note
Decouple default values from user provided values for the installation Setup step on V3
```

NONE


web/src/components/wizard/setup/LinuxSetupStep.tsx
```diff
@@ -122,6 +140,10 @@ const LinuxSetupStep: React.FC<LinuxSetupStepProps> = ({ onNext, onBack }) => {
       return response.json();
     },
     onSuccess: () => {
+      // Update the global (context) config we use accross the project
+      updateConfig(configValues);
+      // Clear any previous errors
+      setError(null);
```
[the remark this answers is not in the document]
JGAntunes: Not true, `updateConfig` supports partial updates which is what we're doing here:
- https://github.com/replicatedhq/embedded-cluster/blob/53aad7964f7cf8032f8d4b3a0fb2eea0f484581a/web/src/providers/KubernetesConfigProvider.tsx#L21-L23

web/src/contexts/LinuxConfigContext.tsx
```diff
@@ -1,16 +1,6 @@
...
-  adminConsolePort?: number;
-  localArtifactMirrorPort?: number;
-  dataDirectory: string;
-  useProxy: boolean;
-  httpProxy?: string;
-  httpsProxy?: string;
-  noProxy?: string;
-  networkInterface?: string;
-  globalCidr?: string;
-}
```
JGAntunes, line 13: Cleaning up types

web/src/contexts/WizardModeContext.tsx
```diff
@@ -1,8 +1,6 @@
 import { createContext, useContext } from "react";
 import { InstallationTarget } from "../types/installation-target";
-import { WizardText } from "../providers/WizardProvider";
-
-export type WizardMode = "install" | "upgrade";
```
JGAntunes, line 5: Cleaning up types

web/src/contexts/LinuxConfigContext.tsx
```diff
@@ -1,16 +1,6 @@
 import { createContext, useContext } from 'react';
```
JGAntunes, line 1: Both the `LinuxConfigContext` and the `KubernetesConfigContext` seem to exist for the sole need of having the admin console port later on (once the installation is finished). I think it would be worthwhile discussing cutting down this complexity and instead using React Query to get the config. My understanding from react query is that if we use consistent cache keys this wouldn't even result in a new request and instead re-use the initial data unless it's past the staleness ttl. Or worst case we get an extra cheap request with a fresh config instead of needing to use context for a couple of values.

sgalsaleh (approved):

PR #2800, opened 2025-09-03
feat(UI): Add back button to installation step

screspod:

Adds a back button to the installation step with the following behavior:
- Enabled when the installation target is linux (kubernetes target does not have a back button), the host preflights run is complete and the status is failed.
- Disabled when host preflights are running, the status is succeeded, and for all subsequent installation phases

https://app.shortcut.com/replicated/story/127700/host-preflights-appear-complete-before-they-actually-finish-allowing-invalid-transition-to-installation

Tests were added

```release-note
NONE
```

NONE


web/src/components/wizard/installation/InstallationStep.tsx
```diff
@@ -188,13 +195,25 @@ const InstallationStep: React.FC<InstallationStepProps> = ({ onNext }) => {
...
+              Back
+            </Button>
+          )}
           <Button
             ref={nextButtonRef}
             onClick={nextButtonConfig.onClick}
             disabled={nextButtonConfig.disabled}
             icon={<ChevronRight className="w-5 h-5" />}
             dataTestId="installation-next-button"
+            className={target === 'kubernetes' ? 'ml-auto' : ''}
```
JGAntunes, line 216: Not super fond of us starting to add a lot of branching around `if isLinux` and `if isKubernetes`. I understand we're just using this Back button for a single phase right now but I can see us falling down this pattern more and more as requirements change if we open the door to it.

I was thinking if we could use a similar logic to the one we do for the nextButton right now? Where we're passing a setNextButtonConfig state mutation method to each phase and we let each phase decide on how the button should behave? We could have a similar thing where we let each phase dictate if we show the Back button and if it should be enabled or disabled?

screspod: Thanks @JGAntunes, that's a good suggestion. I will implement it using a similar logic to the next button.

screspod:
converting this back to draft until I fix the UI flickering caused by the API

PR #2801, opened 2025-09-03
chore(deps): use new go.yaml.in/yaml/v3

emosbaugh:

Use "go.yaml.in/yaml/v3" in favor of deprecated "gopkg.in/yaml.v3"

```release-note

```


JGAntunes:
Seems like there's some functionality that is breaking as consequence of the change:
- https://github.com/replicatedhq/embedded-cluster/actions/runs/17448054807/job/49551060531?pr=2801#step:4:306

PR #2806, opened 2025-09-04
chore(web): run typechecks during test runs

JGAntunes:

This PR does 2 things:
- Changes vitest to run tsc (typescript compiler) when running tests, therefore making sure we run and validate our TS test code. This is relevant because as part of - https://github.com/replicatedhq/embedded-cluster/pull/2798 - I've found there were some buggy HTTP mock responses that would have been caught if we were running TS type validation for them.
- Adds import checks to our eslint config so that we error `npm run lint` when we're importing code not present in our `package.json` (extraneous deps) and we're exporting unused code.

Most of the adjacent changes fix the underlying problems these changes surfaced.

NA

NONE

```release-note
NONE
```

NONE

### DIFF
--- a/web/vitest.setup.ts
+++ b/web/vitest.setup.ts
@@ -8,16 +8,18 @@ expect.extend(matchers);
 // Mock URL for test environment
 const originalURL = globalThis.URL;
 class URLWithMocks extends originalURL {
-  static createObjectURL = vi.fn(() => 'blob:test-url');
+  static createObjectURL = vi.fn(() => "blob:test-url");
   static revokeObjectURL = vi.fn();
 }
 
 // Set up global URL mock
-vi.stubGlobal('URL', URLWithMocks);
+vi.stubGlobal("URL", URLWithMocks);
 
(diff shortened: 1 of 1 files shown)


web/vitest.setup.ts
```diff
@@ -51,20 +35,12 @@ beforeEach(() => {
...
-    ENVIRONMENT: import.meta.env.VITE_ENVIRONMENT,
-    REGISTRY_ENDPOINT: import.meta.env.VITE_REGISTRY_ENDPOINT,
-    PROXY_ENDPOINT: import.meta.env.VITE_PROXY_ENDPOINT,
-    PUSH_ENDPOINT: import.meta.env.VITE_PUSH_ENDPOINT,
-  };
 });
 
 afterEach(async () => {
+  // Clean up DOM
+  cleanup();
```
JGAntunes, line 42: We need to run DOM cleanup between test executions without the `global: true` for some reason 🤷.

sgalsaleh (approved):

PR #2918, opened 2025-09-17
Add support for password hash for API auth

screspod:

```release-note

```


cmd/installer/cli/install.go
```diff
@@ -971,6 +979,13 @@ func ensureAdminConsolePassword(flags *InstallCmdFlags) error {
 		return NewErrorNothingElseToAdd(errors.New("password is not valid"))
 	}
 
+	// Generate bcrypt hash and store it in the installConfig
+	hash, err := generatePasswordHash(flags.adminConsolePassword)
+	if err != nil {
+		return fmt.Errorf("failed to hash password: %w", err)
+	}
+	flags.passwordHash = hash
```
sgalsaleh, line 987: no need to store in flags, check my comment above

api/controllers/auth/controller.go
```diff
@@ -16,32 +23,48 @@ type Controller interface {
...
-	password string
+	passwordHash []byte
 }
 
-type AuthControllerOption func(*AuthController)
-
-func NewAuthController(password string, opts ...AuthControllerOption) (*AuthController, error) {
-	controller := &AuthController{
-		password: password,
+// NewAuthController creates a new auth controller from a plaintext password
```
sgalsaleh, line 29: i think there should only be a single auth controller method that accepts a hash, and update `runManagerExperienceInstall` in the cli code to generate the hash from the provided password and pass it to the api.

sgalsaleh (approved):

PR #3019, opened 2025-10-07
ignore .claude in ci

St0rmz1:

The .claude directory is centrally managed elsewhere and does not require another CI review. Ignore .claude directory.

NONE

```release-note
NONE
```

NONE

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,11 +1,15 @@
 name: CI
 
 on:
-  pull_request: {}
+  pull_request:
+    paths-ignore:
+      - '.claude/**'
 
   push:
     branches:
       - main
+    paths-ignore:
+      - '.claude/**'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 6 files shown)


St0rmz1:
The Cursor bugbot finding is incorrect. The negated pattern `!.claude/**` within the `paths` array is actually the **correct** approach according to [GitHub Actions documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore).

GitHub Actions explicitly states: "You cannot use both the `paths` and `paths-ignore` filters for the same event." When a workflow already has a `paths` filter (like `kinds/**`), the only way to exclude files is by using negation patterns with `!` prefix.

The implementation in this PR follows the documented pattern:
```yaml
paths:
  - 'kinds/**'
  - '!.claude/**'  # Correct: negation pattern within paths
```

Using `paths-ignore` as the bugbot suggests would cause a workflow validation error since these workflows already define `paths` filters.

Reference: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

sgalsaleh (approved):

PR #3676, opened 2026-04-29
fix: add upgrade preflight for stale containerd tmpmounts

kriscoleman:
## Summary

- Add a preflight check to the upgrade spec that detects stale containerd overlay mounts under `containerd/tmpmounts/` and fails with a clear remediation message before the upgrade proceeds
- These stale mounts (left by interrupted image pulls, often caused by security software holding files open) cause "device or resource busy" errors that cascade into broken CNI, unreachable services, and an unrecoverable cluster state

## Context

Investigation of [sc-136511](https://app.shortcut.com/replicated/story/136511) found that neither v2 nor v3 has a preflight check for this condition. The previously referenced ec#277 improved security tool detection in the *support bundle* (diagnostic), not preflights (preventive).

The collector greps `/proc/mounts` for entries under `containerd/tmpmounts`. On a healthy system at upgrade time, there should be zero such mounts — false positives are unlikely.

## Test plan

- [x] YAML validates
- [x] `go vet ./pkg-new/preflights/...` passes
- [x] `go test ./pkg-new/preflights/...` passes (updated test that previously asserted upgrade spec was empty)
- [ ] Manual verification on a test cluster with simulated stale mounts

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg-new/preflights/specs/host-preflight-upgrade.yaml
+++ b/pkg-new/preflights/specs/host-preflight-upgrade.yaml
@@ -4,5 +4,37 @@ metadata:
   name: embedded-cluster-upgrade
 spec:
   # upgrade only collectors and analyzers should be added here
-  collectors: []
-  analyzers: []
+  collectors:
+    - run:
+        collectorName: 'check-stale-containerd-tmpmounts'
+        command: 'sh'
+        args:
+          - -c
+          - |
+            # Detect stale containerd overlay mounts left by interrupted image pulls.
+            # These cause "device or resource busy" errors on subsequent pulls.
+            if grep -q 'containerd/tmpmounts' /proc/mounts 2>/dev/null; then
+              echo "stale_mounts_found=true"
+              echo "---"
+              grep 'containerd/tmpmounts' /proc/mounts
+            else
+              echo "stale_mounts_found=false"
+            fi
+  analyzers:
+    - textAnalyze:
+        checkName: Stale Containerd Temporary Mounts
+        fileName: host-collectors/run-host/check-stale-containerd-tmpmounts.txt
+        regex: 'stale_mounts_found=true'
+        outcomes:
+          - fail:
+              when: 'true'
+              message: >-
+                Stale containerd temporary mounts were detected. These leftover overlay
+                mounts from a previous interrupted operation will cause image pull failures
+                with "device or resource busy" errors during the upgrade. To resolve, run:
+                mount | grep containerd-mount | awk '{print $3}' | xargs -r umount -l
+                Then restart the k0s service with: systemctl restart k0scontroller.service
+                (or k0sworker.service on worker nodes).
+          - pass:
+              when: 'false'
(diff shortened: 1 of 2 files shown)


pkg-new/preflights/specs/host-preflight-upgrade.yaml
```diff
@@ -4,5 +4,37 @@ metadata:
...
+            if grep -q 'containerd/tmpmounts' /proc/mounts 2>/dev/null; then
+              echo "stale_mounts_found=true"
+              echo "---"
+              grep 'containerd/tmpmounts' /proc/mounts
+            else
+              echo "stale_mounts_found=false"
+            fi
+  analyzers:
+    - textAnalyze:
+        checkName: Stale Containerd Temporary Mounts
```
kriscoleman, line 25: "Stale" might not be the best choice of words here, open to suggestions

pkg-new/preflights/specs/host-preflight-upgrade.yaml
```diff
@@ -4,5 +4,37 @@ metadata:
   name: embedded-cluster-upgrade
 spec:
   # upgrade only collectors and analyzers should be added here
-  collectors: []
```
kriscoleman, line 7: looks like we intentionally left these empty before

I'm wondering if it makes sense to add this here?

pkg-new/preflights/specs/host-preflight-upgrade.yaml
```diff
@@ -4,5 +4,37 @@ metadata:
...
+        fileName: host-collectors/run-host/check-stale-containerd-tmpmounts.txt
+        regex: 'stale_mounts_found=true'
+        outcomes:
+          - fail:
+              when: 'true'
+              message: >-
+                Stale containerd temporary mounts were detected. These leftover overlay
+                mounts from a previous interrupted operation will cause image pull failures
+                with "device or resource busy" errors during the upgrade. To resolve, run:
+                mount | grep containerd-mount | awk '{print $3}' | xargs -r umount -l
```
emosbaugh, line 35: im concerned that this isnt the right advice to give. typically when we have seen stale mounts the root cause has been security tools that need to be configured differently.

kriscoleman: yea that is what led us here. I focused this on just cleaning up the mounts, but made no mention of the possible security tools problem. 

If the security tools are the root cause, and they need to be reconfigured, would the customer need to follow these steps still to recover and try again, regardless?

emosbaugh: what you have here just seems like a temporary fix that is likely to break the next time they try to upgrade.

kriscoleman:
Closing — v2 does not have an upgrade preflight path for this. Will retarget the check to `replicatedhq/ec` per @emosbaugh's suggestion.

PR #3771, opened 2026-06-18
feat: add k0s 1.36 support and refresh 1.34/1.35 patches

banjoh:
### What this PR does / why we need it:

Adds k0s 1.36 support and bumps the 1.34/1.35 patches. The supported window keeps rolling via the relative CI matrix (1.36/1.35/1.34, plus 1.33 for upgrade tests). The load-bearing change is containerd v1→v2 (k0s 1.36 ships containerd 2.x), which rejects the legacy `io.containerd.grpc.v1.cri` drop-in schema **and** can't resolve a sandbox image pinned by digest — both surface only on the 1.35→1.36 upgrade.

#### Core version/image changes
- **Version rotation** — `versions.mk` default → 36 (`v1.36.1+k0s.0`), bump 1.34→`v1.34.8`/1.35→`v1.35.5`, add `metadata-1_36.yaml`; regenerated `go.mod`/`go.sum` (+`kinds`) k8s.io replace directives → v0.36.1 and CRDs. Dev Dockerfiles + dryrun bumped to Go 1.26.4.
- **Image metadata** — `metadata-1_3{4,5,6}.yaml` pinned to the mirrored `proxy.replicated.com` tags that match each k0s release (e.g. kube-proxy tracks the cluster's Kubernetes version) rather than the greatest available patch: 1.36 = calico v3.32.0, coredns 1.14.3, envoy v1.37.2, kube-proxy v1.36.1, pause 3.10.2, metrics-server v0.8.1.
- **`airgap.GetImageURIs` API change (1.35+)** — `ListK0sImages` calls the new `airgap.GetImageURIs(TargetEnv{...})` signature via a build-tagged `allK0sImageURIs` (`images_targetenv.go` / `images_legacy.go`); the installer always compiles against the committed k0s module, so the split just keeps both signatures compiling. Tagged `TODO(k0s-1.36-oldest)`.
- **Installer build tags** — `make embedded-cluster` builds with `$(GO_INSTALLER_BUILD_TAGS)` (was bare `osusergo,netgo`).

#### containerd 2.x can't pull a digest-pinned sandbox image
k0s feeds `spec.images.pause` to containerd as the CRI sandbox image. containerd 2.x (k0s 1.36) rejects a digest-pinned ref (`pause:<tag>-<arch>@sha256:…`) — the arch-suffixed tag doesn't exist in the registry, so resolution fails with `FailedCreatePodSandBox` and **no pod can start**.
- `metadata-1_36.yaml` pins `pause` to a **plain tag** (`3.10.2`), not an arch+digest ref like the other images. This is generated, not hand-edited: buildtools gains a `usePlainTag` flag (set on the `pause` component) that emits the tag as-is instead of the usual `-<arch>@<digest>` form, so it survives metadata regeneration.
- On upgrade, the existing cluster's `spec.images.pause` still holds the old digest ref. k0s bakes the sandbox image into the containerd config at worker start and does **not** hot-reload it (confirmed in the k0s `worker/containerd` component), so the ClusterConfig is now updated **before** the autopilot k0s upgrade (`pkg-new/upgrade`) — k0s 1.36 then boots with a pullable, tag-only pause image. Updating it early is safe: the running k0s/containerd ignores the change until it restarts into the new version.

#### Traefik image filtering
k0s 1.36 adds a Traefik image for its NLLB. EC uses Envoy for NLLB, so `ListK0sImages` skips any `/traefik` image (matched by string, since the typed accessor only exists in the 1.36 module) — keeping it out of the airgap bundle and image lists.

#### fix: don't ship `embedded-registry.toml` on online installs; migrate it on airgap
The insecure-registry drop-in (`/etc/k0s/containerd.d/embedded-registry.toml`) is only needed for the airgap in-cluster registry, but was written on every install in the legacy v1 schema — which k0s 1.36 rejects on restart.
- **Online installs no longer create it** (`AddInsecureRegistry` is now airgap-gated in `install.go`; `join`/`restore` were already airgap-only).
- **`AddInsecureRegistry` emits the v3 schema** (`config_path` drop-in + `hosts.toml` `skip_verify`) for k0s 1.36+, legacy below.
- **Upgrades reconcile the stale drop-in via `migrate-containerd-config`** (run per node by the copy-artifacts job before k0s 1.36 starts): airgap rewrites v2→v3, online removes the file. It runs the freshly-pulled target-release binary (`/embedded-cluster/bin/local-artifact-mirror`) so the k0s version it gates on is correct. It's a subcommand of the existing `local-artifact-mirror` binary rather than a new standalone binary — a separate binary would pull in `hostutils` (and its embedded goods), bloating what we ship/materialize.
- The bundled `local-artifact-mirror` is now built with `versions.K0sVersion` baked in (`local-artifact-mirror/Makefile`) — the migration gates on the k0s minor, and without it the binary defaulted to `0.0.0` and silently no-op'd.
- v2-only paths tagged `TODO(k0s-1.36-oldest)`.

#### Host preflights
- Kernel floor → 4.5 and strict cgroup v2 for k8s 1.35+; legacy 3.10 / "v1 or v2" checks gated to older minors.

#### Cleanup
- **update-prober** — the controller now unconditionally disables it (`--disable-components=konnectivity-server,update-prober`), dropping the `canDisableUpdateProber()` runtime probe that guarded it. It's disableable on every supported minor (>= 1.33), k0sproject/k0s#6326.
- Removed `K0sClusterConfigTo129Compat`, reverting to a plain typed `cli.Update` / `k8syaml.Marshal` of the cluster config. Safe because the `ClusterConfig` schema is the same across all three supported minors (1.34/1.35/1.36) — verified against the k0s source: `ImageSpec` is byte-identical, and the helm chart `timeout` has been a `BackwardCompatibleDuration` (accepts a string or an int) since 1.33. The shim only existed to coerce that `timeout` into an int for k0s 1.29, which is moot now that the floor is 1.34.

#### CI / test fixes
- **Pin helm in the kind integration tests** to the version EC ships (`HELM_VERSION`, v4.1.4) instead of `azure/setup-helm@latest` — helm 4.2.1 regressed `--wait` (addon installs hung), turning CI red across all branches.
- `find-previous-stable` falls back to the oldest available previous-minor release when there are fewer than `previous_release_gap` releases.
- `TestCollectSupportBundle` now sets `DISABLE_FILESYSTEM_PERFORMANCE_CHECK=1` (the etcd-disk preflight is flaky on slow CI runners).

**Known limitation (autopilot timing).** Autopilot gives us no hook to rewrite containerd config while k0s is stopped, so the airgap v3 registry drop-in is written *before* autopilot swaps the k0s binary. This leaves a small window where the still-running old containerd (1.7.x) could restart and read the v3 config: it silently ignores it — including the airgap registry pull credentials — so airgap image pulls can fail during that window. It is brief and self-resolving: as soon as autopilot swaps in k0s 1.36 / containerd 2.x, the v3 config is honored.

### Which issue(s) this PR fixes:

[sc-138518](https://app.shortcut.com/replicated/story/138518/add-k0s-1-36-support-to-ec-v2)

### Does this PR require a test?

Yes — covered:
- Unit tests for the containerd schema gating + v2→v3 migration.
- `TestSingleNodeAirgapUpgradeSelinux` asserts the registry drop-in migrated to v3 across the 1.35→1.36 airgap upgrade.
- `TestSingleNodeInstallation` / `TestSingleNodeUpgradePreviousStable` assert the registry drop-in is **absent** on online install and after the online upgrade.

### Does this PR require a release note?

```release-note
Add support for Kubernetes 1.36 (k0s 1.36). Hosts now require cgroup v2 and a kernel >= 4.5 for Kubernetes 1.35 and later.
```

NONE


.github/workflows/ci.yaml
```diff
@@ -25,6 +25,7 @@ jobs:
     outputs:
       git_sha: ${{ steps.git_sha.outputs.git_sha }}
       ec_version: ${{ steps.output_vars.outputs.ec_version }}
+      helm_version: ${{ steps.output_vars.outputs.helm_version }}
```
banjoh, line 28: Version CI is testing with has a regression cause test failures. We should use the same version the installer is using anyway, to ensure the behaviour from helm is the same

kriscoleman (dismissed):
Nice, I think pinning to the same helm version as ec ships makes a ton of sense. 

I need these changes for the release of max redactors and troubleshoot. Thanks! 🙏

kriscoleman (approved):
:shipit:

operator/pkg/artifacts/upgrade.go
```diff
@@ -63,6 +63,20 @@ var copyArtifactsJob = &batchv1.Job{
 							},
 						},
 					},
+					{
+						// Gives the migrate-containerd-config step write access to the host
+						// containerd config so it can reconcile the registry drop-in for the
+						// containerd v1 -> v2 (k0s 1.36+) schema change before k0s restarts:
+						// migrate it on airgap, delete it on online.
+						// TODO(k0s-1.36-oldest): drop this mount and the migration.
```
sgalsaleh, line 71: careful with this todo comment, i think we would drop this when 1.37 is oldest not 1.36, no? because if we still support 1.36 it means we support upgrading from 1.35 to 1.36, which still requires this migration. but if we only support 1.37+, it means you have to be on 1.36 to upgrade and that means the migration already took place.

sgalsaleh: same for other todo comments

sgalsaleh (approved):

operator/pkg/artifacts/upgrade.go
banjoh: Good point

PR #3909, opened 2026-09-01
Don't include EC images in backup

divolgin:

Exclude registry data from snapshots. Snapshot restore requires the same application bundle that is being restored. In case of air gap installation, the bundle also contains all the necessary images, which can be pushed into registry during the restore process.

New features:
 ```release-notes-features
 
 ```
 Bug fixes:
 ```release-notes-fixes
 
 ```
 Improvements:
 ```release-notes-improvements
Air-gapped disaster recovery backups are now faster and smaller by excluding registry data that can be reconstructed from the air gap bundle during restore.
 ```

### DIFF
--- a/cmd/installer/cli/restore.go
+++ b/cmd/installer/cli/restore.go
@@ -17,6 +17,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
+	s3types "github.com/aws/aws-sdk-go-v2/service/s3/types"
 	k0sv1beta1 "github.com/k0sproject/k0s/pkg/apis/k0s/v1beta1"
 	apitypes "github.com/replicatedhq/embedded-cluster/api/types"
 	"github.com/replicatedhq/embedded-cluster/cmd/installer/kotscli"
@@ -26,6 +27,7 @@ import (
 	"github.com/replicatedhq/embedded-cluster/pkg-new/kubernetesinstallation"
 	"github.com/replicatedhq/embedded-cluster/pkg-new/preflights"
 	"github.com/replicatedhq/embedded-cluster/pkg/addons"
+	"github.com/replicatedhq/embedded-cluster/pkg/addons/seaweedfs"
 	addontypes "github.com/replicatedhq/embedded-cluster/pkg/addons/types"
 	"github.com/replicatedhq/embedded-cluster/pkg/disasterrecovery"
 	"github.com/replicatedhq/embedded-cluster/pkg/helm"
@@ -62,6 +64,7 @@ const (
 	ecRestoreStateWaitForNodes         ecRestoreState = "wait-for-nodes"
 	ecRestoreStateRestoreSeaweedFS     ecRestoreState = "restore-seaweedfs"
 	ecRestoreStateRestoreRegistry      ecRestoreState = "restore-registry"
+	ecRestoreStatePopulateRegistry     ecRestoreState = "populate-registry-from-airgap-bundle"
 	ecRestoreStateAdminConsoleEnableHA ecRestoreState = "admin-console-enable-ha"
 	ecRestoreStateRestoreECO           ecRestoreState = "restore-embedded-cluster-operator"
 	ecRestoreStateRestoreExtensions    ecRestoreState = "restore-extensions"
@@ -76,6 +79,7 @@ var ecRestoreStates = []ecRestoreState{
 	ecRestoreStateWaitForNodes,
 	ecRestoreStateRestoreSeaweedFS,
 	ecRestoreStateRestoreRegistry,
+	ecRestoreStatePopulateRegistry,
 	ecRestoreStateAdminConsoleEnableHA,
 	ecRestoreStateRestoreECO,
 	ecRestoreStateRestoreExtensions,
@@ -279,6 +283,20 @@ func runRestore(ctx context.Context, appSlug, appTitle string, flags installFlag
 		}
 
 		err = runRestoreRegistry(ctx, installCfg, backupToRestore)
+		if err != nil {
+			return err
+		}
+
+		fallthrough
+
+	case ecRestoreStatePopulateRegistry:
+		logrus.Debugf("setting restore state to %q", ecRestoreStatePopulateRegistry)
+		err := setECRestoreState(ctx, ecRestoreStatePopulateRegistry, backupToRestore.GetName())
+		if err != nil {
+			return fmt.Errorf("unable to set restore state: %w", err)
+		}
+
+		err = runPopulateRegistry(ctx, appSlug, flags.airgapBundle, backupToRestore)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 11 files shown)


divolgin:
# Reconstructing the Airgap Registry During Restore for Embedded Cluster v2

The restore currently requires the matching installation assets and validates that the supplied airgap bundle matches the executable. The documentation also requires assets for the exact application version contained in the backup. [Replicated restore documentation](https://docs.replicated.com/embedded-cluster/v2/embedded-disaster-recovery#restore-from-a-backup)

## What happens today

During an airgap restore:

1. `restore` reads and validates the supplied airgap bundle.
2. It extracts only:
   - `embedded-cluster/images-amd64.tar` for k0s/containerd
   - `embedded-cluster/charts.tar.gz`
3. It installs k0s, OpenEBS, and Velero.
4. It restores the Admin Console.
5. For HA airgap installations, it restores SeaweedFS.
6. It restores the registry Kubernetes resources and registry volume using Velero.
7. It continues with HA, operator, extensions, and application restoration.

The relevant current registry step is [restore.go](/Users/divolgin/git/go/src/github.com/replicatedhq/embedded-cluster/cmd/installer/cli/restore.go:689). It only performs the Velero restore and configures containerd; it does not push anything from the bundle.

## Existing push functionality

EC already contains a wrapper around:

```text
kubectl-kots admin-console push-images
```

KOTS’s implementation does everything needed:

- Pushes application images
- Pushes EC images from the bundle’s embedded registry
- Pushes the embedded artifacts:
  - `embedded-cluster-amd64`
  - `images-amd64.tar`
  - `charts.tar.gz`
  - `kots.tar.gz`
  - `operator.tar.gz`
  - `version-metadata.json`
  - local artifact mirror content

This matches the repositories we found in the backed-up registry. So EC would not need a new bundle parser or image copier; it could call the existing [KOTS wrapper](/Users/divolgin/git/go/src/github.com/replicatedhq/embedded-cluster/cmd/installer/kotscli/kotscli.go:382).

## Proposed restore sequence

For a new backup format:

```text
Restore EC Installation
Restore Admin Console
Wait for additional HA nodes
Restore/create SeaweedFS without its old artifact data
Restore/create registry with an empty backing store
Wait for registry readiness
Push images and EC artifacts from the supplied airgap bundle
Enable Admin Console HA
Restore/reconcile EC operator
Reinstall extensions
Restore application resources and mutable application data
```

The image push should occur immediately after `runRestoreRegistry`, before the operator, extensions, or application depend on registry artifacts.

## Backup changes

For single-node airgap installations:

- Remove `backup.velero.io/backup-volumes: data` from the registry pod.
- Continue backing up the registry’s Kubernetes metadata, Secrets, Service, and configuration.
- Do not back up `/var/lib/registry`.

For HA airgap installations, removing only the registry annotation is insufficient. Registry data has been migrated to SeaweedFS, which is dedicated to registry storage and whose volumes are also explicitly backed up. Its immutable blob data should therefore also be excluded and rebuilt from the bundle. The SeaweedFS filer data must be excluded together with the volume-server data: restoring filer metadata without its referenced chunks leaves dangling registry objects that prevent an idempotent push from fully reconstructing the backing store.

That distinction matters for `TestMultiNodeAirgapHADisasterRecovery`: the inspected HA backup contained approximately 5.38 GB in the SeaweedFS volume-server data, while the filer metadata, filer logs, and KOTS database together were only around 2 MB. The application definition caused the same SeaweedFS data to be included in both component backups, producing a worst-case logical backup of approximately 10.76 GB that took 13 minutes to create. Regardless of that duplication, the HA equivalent of the single-node registry volume is the SeaweedFS data and it contains the same reconstructable airgap content.

## Backup-format marker

Do not infer the new behavior merely from a missing `PodVolumeBackup`. Add an explicit annotation, for example:

```yaml
kots.io/embedded-registry-data-source: airgap-bundle
```

Then restoration can behave safely:

- Legacy backup without the annotation:
  - Restore registry/SeaweedFS volume data as today.
- New backup with the annotation:
  - Restore empty storage.
  - Repopulate it from the supplied bundle.

This provides a clean compatibility boundary. Restore requires assets matching the EC version in the backup, so backup creation and restore behavior move together; a separate transitional format is unnecessary.

## Credentials

The restored backup already contains the registry authentication material, including:

- Registry authentication Secret
- KOTS private-registry Secret
- Registry TLS Secret
- Original registry Service IP annotation

The restore implementation should resolve the authoritative credentials from the restored registry-related Secrets rather than assuming a Secret name. EC should avoid using `registry.GetRegistryPassword()` at this point because that value is randomly generated by the new restore process and might not match the restored registry Secret.

## Resume behavior

Add a distinct restore state:

```text
restore-registry
populate-registry-from-airgap-bundle
admin-console-enable-ha
```

The population step must be resumable and idempotent. KOTS already supports skipping images whose destination manifests exist, which is useful if the command is interrupted halfway through.

## Historical registry content

The registry can retain images from application releases imported through earlier airgap bundles. Those historical images are not directly usable in Embedded Cluster v2:

- Application rollback is not supported.
- Pending application versions do not exist.

The documentation requires restoration with assets matching the application version in the backup, so the supplied bundle contains the usable application release that must be reconstructed. Preserving inaccessible images from older releases is not part of the restore contract and does not justify backing up the registry data.

## Delivery

This is a self-contained change. Excluding registry and SeaweedFS data from backups and reconstructing that data during restore will ship together in the next EC release. The existing full `TestMultiNodeAirgapHADisasterRecovery` test will cover the complete backup and restore path before the change is released.

No staged compatibility rollout is required. Customers can install or skip any EC release, and restore already requires installation assets matching the EC version recorded in the backup. Each release therefore uses the backup and restore behavior implemented by that matching release.

## Expected impact

For the backups we inspected:

- Single-node registry volume: approximately 2.69 GB
- Three-node HA SeaweedFS data: approximately 5.38 GB per component backup
- Worst-case HA logical backup with the application-defined duplication: approximately 10.76 GB
- Worst-case measured snapshot creation time: 13 minutes
- Mutable KOTS data, SeaweedFS metadata, and resource metadata: only a few MB
- Reconstructable registry content: virtually all remaining data

Backup creation and S3 storage would improve dramatically. Restore would still need to push the registry content from the local bundle, so it does not make that work disappear—it moves it from S3/Kopia restoration to deterministic local reconstruction. We would need to measure the existing KOTS push path; it may still consume several minutes.

One documentation caveat: the current v2 page says airgap DR is not officially supported because AWS S3 requires internet access, although the code and E2E use MinIO/S3-compatible endpoints. That product-support mismatch should be resolved separately. [Documented v2 limitation](https://docs.replicated.com/embedded-cluster/v2/embedded-disaster-recovery#limitations-and-known-issues)

cmd/installer/cli/restore.go
```diff
@@ -709,6 +727,142 @@ func runRestoreRegistry(ctx context.Context, installCfg *installConfig, backupTo
...
+	if err != nil {
+		return fmt.Errorf("get kotsadm namespace: %w", err)
+	}
+	username, password, err := registryCredentialsFromSecrets(ctx, kcli, kotsadmNamespace, registryAddress)
+	if err != nil {
+		return err
+	}
+	loading := spinner.Start()
+	defer loading.Close()
+	loading.Infof("Populating the embedded registry from the airgap bundle")
```
sgalsaleh: How about:
```suggestion
	loading.Infof("Restoring registry data")
```

sgalsaleh (approved):

PR #3942, opened 2026-09-10
Use SecureBuild release images and remove legacy build templates

mikhailswift:

SecureBuild now builds and publishes the operator and local artifact mirror images. Remove the duplicate GitHub Actions Docker Hub publishers and consume the SecureBuild images in releases instead.

- Make SecureBuild build and external registry publishing failures block stable releases, resolve published image digests directly, and preserve the installer's existing digest references. The operator chart continues to reference its image by tag only.
- Use the existing Docker Hub repositories with SecureBuild tags such as `2.19.9-k8s1.36`, including in nightly scans.
- Delete the four legacy APKO/Melange templates and unused template and addon image-building helpers.
- Preserve Dagger builders used by CI and local development while reading the Git-managed SecureBuild specs without rewriting YAML. As in KOTS, Melange uses a local `git-checkout` pipeline override, `--source-dir`, and `--env-file`; APKO uses repository/key arguments to install locally signed head packages through their aliases, including the LAM compat package. The operator recipes honor a supplied `VERSION` and default to the release package version.

Prerelease limitation: the SecureBuild workflow currently accepts only stable versions. Prerelease releases require images to have been published separately under the same naming convention; there is no GitHub image-building fallback.

NONE

Validated locally:

- `go test -tags containers_image_openpgp,exclude_graphdriver_btrfs,exclude_graphdriver_devicemapper,exclude_graphdriver_overlay ./cmd/buildtools`.
- Built and exported both Kubernetes 1.36 ARM64 images through Dagger using the unchanged specs and the prerelease version `v2.19.10+k8s-1.36-rc1`, without publishing. Verified local source selection, installation of the local head packages (including LAM compat), the legacy LAM symlink, and the requested operator binary version.
- Checked the image-resolution shell syntax and exercised successful digest lookups and immediate failure for an unavailable image using mocked registry responses.
- Actionlint: SecureBuild publishing and nightly scan workflows pass; release workflow introduces no new findings compared with the base revision (10 existing findings remain).
- `git diff --check`.

Full GitHub release publishing was not run locally.

New features:
```release-notes-features
NONE
```
Bug fixes:
```release-notes-fixes
NONE
```
Improvements:
```release-notes-improvements
NONE
```

Updated `dagger/README.md` to describe direct use of the shared recipes, release image naming, and the prerelease limitation. No external documentation changes required.


.github/workflows/release-prod.yaml
```diff
@@ -83,36 +82,62 @@ jobs:
...
         run: |
-          make -C operator build-deps build-and-push-operator-image \
-            PACKAGE_VERSION=${{ needs.get-tag.outputs.tag-name }}
-          echo "image=$(cat operator/build/image)" >> $GITHUB_OUTPUT
+          printf '%s' "$REGISTRY_PASSWORD" | crane auth login "$REGISTRY_SERVER" --username "$REGISTRY_USERNAME" --password-stdin
+          tag="${VERSION#v}"
+          tag="${tag/+k8s-/-k8s}"
+          operator="$OPERATOR_IMAGE_NAME:$tag"
+          mirror="$LOCAL_ARTIFACT_MIRROR_IMAGE_NAME:$tag"
+          # External registry publishing can finish after the SecureBuild build.
```
divolgin: is this really true? seems like a bug

divolgin (dismissed):

.github/workflows/release-prod.yaml
mikhailswift: i don't think it is true

divolgin (approved):

